PR #1, opened 2023-01-14
Initial content

gkellogg:
Updates basic content for RDF Concepts. Note some ReSpec errors that have been reported at https://github.com/w3c/respec/issues/4362.


gkellogg:
Presumably, PR Preview updates will kick in after this is merged.

.github/workflows/auto-publish.yml
```diff
@@ -0,0 +1,26 @@
...
+          TOOLCHAIN: respec
+          SOURCE: spec/index.html
+          W3C_ECHIDNA_TOKEN: ${{ secrets.ECHIDNA_TOKEN }}
+          VALIDATE_LINKS: false
+          VALIDATE_MARKUP: true
+          ACTIONS_STEP_DEBUG: true
+          # W3C_WG_DECISION_URL: https://www.w3.org/2022/11/09-rch-minutes#r03
+          W3C_BUILD_OVERRIDE: |
+            specStatus: WD
+            shortName: rdf12-concepts
```
pchampin, line 26: I would rather comment that out until we have a vote and a correct setup
```suggestion
#name: Echidna Auto-publish
#on:
#  pull_request:
#    paths: ["spec/**"]
#  push:
#    branches: [main]
#    paths: ["spec/**"]
#jobs:
#  main:
#    name: Echidna Auto-publish WD
#    runs-on: ubuntu-20.04
#    steps:
#      - uses: actions/checkout@v3
#      - uses: w3c/spec-prod@v2
#        with:
#          TOOLCHAIN: respec
#          SOURCE: spec/index.html
#          W3C_ECHIDNA_TOKEN: ${{ secrets.ECHIDNA_TOKEN }}
#          VALIDATE_LINKS: false
#          VALIDATE_MARKUP: true
#          ACTIONS_STEP_DEBUG: true
#          # W3C_WG_DECISION_URL: https://www.w3.org/2022/11/09-rch-minutes#r03
#          W3C_BUILD_OVERRIDE: |
#            specStatus: WD
#            shortName: rdf12-concepts
```

pchampin (changes requested):
LGTM, except for the auto-publish.yml that I prefer to comment out for the moment

.github/workflows/auto-publish.yml
gkellogg: Sure, I figured commenting out the Decision URL would be enough, but may as well have the whole block commented out.

gkellogg:
The only problem with commenting everything out, is that we don't get the build issues noted. Maybe it would be better to leave it in, an comment out the pieces that might cause it to auto-publish: W3C_ECHIDNA_TOKEN and W3C_WG_DECISION_URL.

Of course, if there is no ECHIDNA_TOKEN configured, it couldn't do that anyway, but it's nice to see the ReSpec errors (even if currently broken).

gkellogg:
I'm going a head and merging, so that it will at least be visible to the group. Based on the reaction to this, I can upload other RDF notes/specs. But, if there are some form issues, better get them addressed now so they can be applied to other specs before too much time is invested.

afs:
LGTM though I'm looking to learn what the current good practice is for document publishing.

I'm following the `spec/index.html` style.

gkellogg:
We're getting build errors, as the GH Action is commented out, so looks like a problem. Either it should be removed, or (my preference) restored to where it was, with the Echidna fields commented out so it doesn't actually try to publish to /TR.

Generally, the group decides on FPWD and documents are manually prepared. The group can create a decision to automatically publish WDs after that, usually when pushing to "main".

We can put more into the GH Action (either by adding to the workflow, or creating new actions) do do other things, such as check examples for syntax or correctness, and potentially auto-build some resources.

Otherwise, the use of ReSpec comes down to style, and it would be good to adopt a fairly consistent style across the different specs (unlike previous WGs), so having some consistent editorial rules (and use of ReSpec features), if no fully-shared editor roles, is something we should strive for. Probably requires a separate communication channel for editorial matters, if not coordination meetings.

PR #17, opened 2023-02-19
IRI Terminology

gkellogg:
For #15.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/17.html" title="Last updated on Feb 20, 2023, 6:59 PM UTC (8912f1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/17/e93ecf0...8912f1d.html" title="Last updated on Feb 20, 2023, 6:59 PM UTC (8912f1d)">Diff</a>

Issue #15, opened 2023-02-17
Revise terminology to align with current RFCs.


afs:
RFC 3987 is a "proposed standard" in the RFC process.

RDF relies on the syntax extensions that [RFC 3987](https://www.rfc-editor.org/rfc/rfc3987) makes to [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986).

RFC 3986 is updated by [RFC 6874](https://www.rfc-editor.org/rfc/rfc6874) and [RFC 8820](https://www.rfc-editor.org/rfc/rfc8820).

This should include [RDF1.1 Erratum 29](https://www.w3.org/2001/sw/wiki/RDF1.1_Errata#erratum_29) which refers to the use of certain URI/IRI terminolgy. (["absolute IRI", "relative IRI reference"](https://lists.w3.org/Archives/Public/public-rdf-comments/2021Mar/0002.html))

afs:
Yes - I nearly put HTML URLs in. I couldn't find a grammar to link to. They do come from a different design space and have various features related to that (e.g. display, normalization of output). They do discuss "c:"!

gkellogg:
There have been issues raised before on using the URL spec rather than IRI. The main advantage is that it is properly citable, but the lack of a BNF grammar.

My thought is that we continue to use the term "IRI", defined in RDF Concepts for others specs to cite. That is, in turn, defined in an appendix based on the [URL living standard](https://url.spec.whatwg.org/) with a historical reference to [RFC 3987](https://www.rfc-editor.org/rfc/rfc3987) (also in changes). The appendix can include the [ABNF from RFC397](https://www.rfc-editor.org/rfc/rfc3987#section-2.2), but probably should be non-normative. Validation comes from the URL spec, with informative reference to the historical ABNF and note that there are valid IRIs not handled by the ABNF.

We also update our terminology here, and elsewhere as noted in [RDF1.1 Erratum 29](https://www.w3.org/2001/sw/wiki/RDF1.1_Errata#erratum_29).

gkellogg:
[...]

PR #21, opened 2023-02-20
ABNF grammar for RFC3987.

### LINKED ISSUE #20: Add  Appendix on complete grammar for IRI based on RFC3987.
See https://github.com/w3c/rdf-concepts/issues/15#issuecomment-1436962088 for the discussion leading to this issue.

gkellogg:
This is the merged grammar for an IRI from RFC3987 and RFC3986.

The HTML is generated using [ebnf](https://github.com/dryruby/ebnf) with a fair amount of hand editing. It's questionable if the HTML is better than the raw ABNF which has manual formatting, but no interlinkage of productions

It likely needs further updates from referenced RFCS (6874, 7320, 8820).

Text in the appendix and in the referencing note is pretty basic.

Fixes #20.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/21.html" title="Last updated on Mar 28, 2023, 8:45 PM UTC (59df579)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/21/1f16ce9...59df579.html" title="Last updated on Mar 28, 2023, 8:45 PM UTC (59df579)">Diff</a>


domel (commented):
Why is the grammar in ABNF (not in EBNF)? I know that ABNF is common in RFCs, but let's stay consistent nad use in all specs EBNF.

afs:
@domel - in this case, we aren't providing a new grammar, we're helping developers by combining existing material. Being able to compare with the original version is useful for that.

domel:
@afs yes I know, but I do not want to create a new grammar. Just use different metasyntax.

afs:
Changed the appear suggests we have changed something (which we might by accident!).

This is most important for the text form.

gkellogg:
EBNF is also. Ore limited than ABNF in expressing fixed ranges, such as `[ *5( h16 ":" ) h16 ] "::" h16`, which would need to be expanded in EBNF.

gkellogg:
Note, the HTML table at the end of https://github.com/w3c/rdf-concepts/issues/20#issuecomment-1437506718 is more or less what the grammar looks like when changed to EBNF.

spec/index.html
```diff
@@ -1362,6 +1367,16 @@ <h2>Internationalization Considerations</h2>
   <p class="issue" data-number="9"></p>
 </section>
 
+<section id="iri-abnf" class="appendix informative">
+  <h2>IRI Grammar</h2>
+
+  <p>The following [[ABNF]] grammar is from [[RFC3987]] and includes the referenced
+    productions from [[RFC3986]].</p>
```
afs, line 1374:
```suggestion
  <p>
    The following [[ABNF]] grammar applies the changes from [[RFC3987]] to the section
    <a href="https://www.rfc-editor.org/rfc/rfc3986#appendix-A">Collected ABNF for URI</a> of
    [[RFC3986]] to give a consolidated grammar for IRIs.
   </p>
   <p>
    This is provided for convenience only.
    If it differs from definitions in [[RFC3986]], [[RFC3987]] or any subsequent updates,
    then those definitions should be used.
    </p>
  ```

Comment (afs) [reply spec/index.html:1374]: Words to say that the RFCs are definitive/normative.

File: spec/index.html [hunk 2]
```diff
@@ -1362,6 +1367,16 @@ <h2>Internationalization Considerations</h2>
...
+
+  <div data-include="iri-grammar.html"></div>
```
afs, line 1376:
```suggestion
  <div data-include="iri-grammar.abnf" data-include-format='text'></div>
```
This may need to be include markdown and put `\`` ... \``` around the ABNF inclusion.

Ideally, the inclusion would be just the text and there would be a URL text/plain that can be read directly.

afs (changes requested):

afs:
> It's questionable if the HTML is better than the raw ABNF

Yes. Putting something with a different appearance will do more harm than good.

I offered to produce a combination of RFC 3986 together with the changes in RFC 3987 applied. By doing the minimal integration, readers can easily check for themselves that the grammar is indeed a mechanical process of and nothing more than described.

If HTML links in the ABNF is easily doable, while retaining the appearance and layout of the original RFCs, that _might_ worth considering; it still raises the question "where did the HTML come from?" (I don't think it is worth the effort, especially not at the moment.)

Anything that changes the appearance carries risk with little gain. We do not want to become another IRI grammar.

> It likely needs further updates from referenced RFCS (6874, 7320, 8820).

Only 6874 changes the grammar in any way. It only touches IPv6 addresses.

8820 is "best practice"
(7320 is obsolete)

spec/index.html
gkellogg: Updated to use `data-cite`.

spec/index.html
gkellogg: I'll work on this.

spec/index.html
```diff
@@ -1362,6 +1367,26 @@ <h2>Internationalization Considerations</h2>
   <p class="issue" data-number="9"></p>
 </section>
 
+<section id="iri-abnf" class="appendix informative">
+  <h2>IRI Grammar</h2>
+
+  <p>
+    The following [[ABNF]] grammar applies the changes from [[RFC3987]] to the section
```
gkellogg, line 1374:
```suggestion
    The following [[ABNF]] grammar applies the changes from [[RFC3987]] and [[RFC 8820]] to the section
```

afs: 8820 is a best practice document. I don't see any grammar changes. 6874 makes a change - suggestion updated.

afs: In the RFCs, the grammar lines are indented by 3 spaces. Can we please keep that?

gkellogg: That’s respec likely normalizing the leading space. It may take a hook to add leading white space programmatically.

gkellogg:
Actually, I’m sure we can do it with some CSS.

spec/index.html
afs: They are not in `iri-grammar.abnf`.

spec/index.html
```diff
@@ -1362,6 +1368,28 @@ <h2>Internationalization Considerations</h2>
...
+  <h2>IRI Grammar</h2>
+
+  <p>
+    The following [[ABNF]] grammar applies the changes from [[RFC3987]] and [[RFC6874]] to the section
+    <a data-cite="RFC3986#appendix-A">Collected ABNF for URI</a> of
+    [[RFC3986]] to give a consolidated grammar for IRIs.
+   </p>
+   <p>
+    This is provided for convenience only.
+    If it differs from definitions in [[RFC3986]], [[RFC3987]] or any subsequent updates,
```
TallTed, line 1381:
```suggestion
    If it differs from definitions in [[RFC3986]], [[RFC3987]], or any subsequent updates,
```

domel (approved):

TallTed (approved):

afs (approved):
The note "RFC3987 never made it past the Proposed Standard phase and should not be normatively cited." can be removed.

gkellogg:
> (not to hold-up this PR ...)
> 
> Is there a simple way to not have "IRI" in the section title "IRI Grammar" be an _abbr_?
> 
> It looks odd when one section has an under-dotted item in it when all the other don't. It looks like a rendering error.

Looks like something ReSpec does as it sees that "IRI" was defined as an abbreviation. I'll investigate, but as "IRI"  _is_ an abbreviation, it's not wrong for it to be identified as such in a heading.

TallTed (approved):

afs:
> Looks like something ReSpec does as it sees that "IRI" was defined as an abbreviation. I'll investigate, but as "IRI" _is_ an abbreviation, it's not wrong for it to be identified as such in a heading.

So is RDF 😄 and if that becomes under dotted, we'll get inconsistent appearance (not in "RDF Graph" which is terminology in its own right)
And it doesn't touch "IRIs".

Maybe the trick is to define IRI.

afs:
> Looks like something ReSpec does as it sees that "IRI" was defined as an abbreviation. I'll investigate, but as "IRI" _is_ an abbreviation, it's not wrong for it to be identified as such in a heading.

So is RDF 😄 and if that becomes under-dotted, we'll get inconsistent appearance (not in "RDF Graph" which is terminology in its own right)

May be the trick is, for RDF/SPARQL usage, to define IRI.

gkellogg:
> May be the trick is, for RDF/SPARQL usage, to define IRI.

We actually do define IRI as a term, so maybe we just don't need to make it an abbreviation, as we don't abbreviate RDF, as you point out. That's probably best done as part of #17.

afs:
Added to issue #15 because PR #17 is closed.

afs:
These latest commits (march 28th) don't seem to be relate to the grammar the PR is providing.

gkellogg:
> These latest commits (march 28th) don't seem to be relate to the grammar the PR is providing.

True, but there are a lot of overlapping changes, and we were caught up in the problems introduced by using "abbr" for the IRI term. 9aa08f1faa62aa3fa626a4ea8fda90e91a9d0c77 could have probably waited until we address #15, but it represents a broken identifier which is causing link errors. Since this PR is on IRI it made sense to lump that change in here, but it could instead go in as part of #27. I don't expect to create a PR addressing #15 until after the current batch of PRs is merged, otherwise.

59df5790e942a8c75c1e15beffd893df82878eb6 Should probably be part of #27, as changes are not directly related to IRI; likely a case of committing to the wrong branch. If you like, I can reverse 59df5790e942a8c75c1e15beffd893df82878eb6 and add it to #27. I think we should just leave 9aa08f1faa62aa3fa626a4ea8fda90e91a9d0c77 as part of this PR. My thought was that we'll likely merge these changes after the upcoming meeting.

afs:
No problem - just checking they were going to get lost in a jungle of merges.

gkellogg:
With this many open PRs there is always the danger that something is lost when rebasing.

Issue #22, opened 2023-03-16
mention `xsd:string` as default datatype


VladimirAlexiev:
`"foo"` is the same as `"foo"^^xsd:string`, eg in SPARQL:
- `select ("foo"="foo"^^xsd:string as ?x) {}` is true
- `select * {?x ?p "foo"^^xsd:string}` fetches the same triples as  `select * {?x ?p "foo"}`

But I don't see any mention of `xsd:string` as default datatype:
- https://w3c.github.io/rdf-concepts/spec/#section-Datatypes
- https://w3c.github.io/rdf-concepts/spec/#xsd-datatypes

pfps:
I don't see that the notion of a default datatype has a place in this document or, indeed, elsewhere in RDF or SPARQL.   Every literal in RDF has at least a lexical form and a datatype.

That "foo" or 7 are syntax for literals in some surface syntaxes is a separate matter.

gkellogg:
Indeed, all literals have explicit datatypes in the abstract syntax: a Unicode string, a datatype IRI, and a language tag is the datatype IRI is `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString`. No defaults or options. There is a note about [simple literals](https://w3c.github.io/rdf-concepts/spec/#dfn-simple-literal), which is closest to what you describe as a default datatype, but this is a description of what concrete syntaxes may support. Most (all?) concrete syntaxes do support this concept, including SPARQL. (Note  that there's still a sentence "A plain literal is lower than an RDF literal with type xsd:string of the same lexical form." in the Editor's Draft which is probably obsolete).

afs:
> (Note that there's still a sentence "A plain literal is lower than an RDF literal with type xsd:string of the same lexical form." in the Editor's Draft which is probably obsolete).

https://github.com/w3c/sparql-query/issues/26

It is now obsolete.

VladimirAlexiev:
I'm happy with the note about [simple literals](https://w3c.github.io/rdf-concepts/spec/#dfn-simple-literal) and I wonder how I missed it. Maybe because it mentions http://www.w3.org/2001/XMLSchema#string but not `xsd:string`.
Would it be ok to mention the prefixed forms after the full URLs, i.e. add these two in parentheses:

[...]

PR #28, opened 2023-04-01
mention the prefixed forms after the full URI

hartig:
... as suggested by @VladimirAlexiev in https://github.com/w3c/rdf-concepts/issues/22#issuecomment-1492868520

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/28.html" title="Last updated on Apr 4, 2023, 9:42 PM UTC (492cda2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/28/7d01308...492cda2.html" title="Last updated on Apr 4, 2023, 9:42 PM UTC (492cda2)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -574,7 +574,7 @@
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>, and</li>
       <li>if and only if the <a>datatype IRI</a> is
-        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>, a
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>), a
         non-empty <dfn>language tag</dfn> as defined by [[!BCP47]]. The
         language tag MUST be well-formed according to
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
@@ -592,11 +592,11 @@
       Simple literals are syntactic sugar for abstract syntax
       <a>literals</a>
       with the <a>datatype IRI</a>
-      <code>http://www.w3.org/2001/XMLSchema#string</code>. Similarly, most
-      concrete syntaxes represent
+      <code>http://www.w3.org/2001/XMLSchema#string</code> (<code>xsd:string</code>).
+      Similarly, most concrete syntaxes represent
       <a>language-tagged strings</a> without
       the <a>datatype IRI</a> because it always equals
-      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>.</p>
+      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>).</p>
 
     <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is:</p>
 
@@ -875,7 +875,7 @@
     from the lexical space to the value space.</p>
 
   <p class="note"><a>Language-tagged strings</a> have the <a>datatype IRI</a>
-    <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>.
+    <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>).
     No datatype is formally defined for this IRI because the definition
     of <a>datatypes</a> does not accommodate
     <a>language tags</a> in the <a>lexical space</a>.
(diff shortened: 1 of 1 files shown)


afs:
It's not true in RDF/XML, nor N-Triples/N-Quads.
For Turtle/JSON-LD, it assumes the appropriate declarations have been made.

So this at least needs qualification that it applies to certain syntaxes and only when the appropriate declarations have been made.

Otherwise a reasonable assumption is that certain prefixes are part of the abstract data model.

pfps:
There are already prefixed names in the document, such as rdf:HTML, and a description of how prefixed names are used in the construction of IRIs.  I don't think that much qualification is needed here, if any.

spec/index.html
```diff
@@ -574,7 +574,7 @@ <h2>Literals</h2>
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>, and</li>
       <li>if and only if the <a>datatype IRI</a> is
-        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>, a
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>), a
```
afs, line 577: I suggest that we do not use the prefix form at this point.

afs:
And [section 1.4](https://w3c.github.io/rdf-concepts/spec/#vocabularies) mentions prefixes -- it is non-normative.

My concern is that the proposed text (language tag case) is the normative definition the definition is for the abstract data model. Using later seems fine because it has said "syntax".

The IRI section just above does not mention prefix names.

gkellogg (commented):
I'm undecided on if this is a good idea. @afs's point about the difference between normative and non-normative text is useful, but if we were to change, say, `rdf:HTML` to be normative, would we want to comprehensively change to use the full IRI version? Not clear.

That said, a parenthetical using the prefixed name after the full IRI does not make the spec less clear, and could help people scanning the text, and the cost of reinforcing that some prefixes are special.

afs:
My suggestion on the PR is that the abbreviated for is not used where a definition happens (literal bullet 3) and can be used when the syntax is discussed at "Please note that concrete syntaxes ...".

This is the same as `rdf:HTML`

`rdf:HTML` is used as the (human) name of the datatype and at the key point the full IRI is given:

> The IRI denoting this datatype
> is http://www.w3.org/1999/02/22-rdf-syntax-ns#HTML.

spec/index.html
hartig:
```suggestion
        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>, a
```

hartig:
Okay @afs, as per your suggestion, I have removed `rdf:langString` again from the definition of _literal_ (but kept the prefixed forms in the other places where added by this PR).

gkellogg (approved):

afs (approved):

spec/index.html
```diff
@@ -592,11 +592,11 @@ <h2>Literals</h2>
       Simple literals are syntactic sugar for abstract syntax
       <a>literals</a>
       with the <a>datatype IRI</a>
-      <code>http://www.w3.org/2001/XMLSchema#string</code>. Similarly, most
-      concrete syntaxes represent
+      <code>http://www.w3.org/2001/XMLSchema#string</code> (<code>xsd:string</code>).
```
hartig, line 595: Change along the lines of @TallTed's suggestion in https://github.com/w3c/rdf-concepts/issues/22#issuecomment-1494927596

```suggestion
      <code>http://www.w3.org/2001/XMLSchema#string</code>
      (which is commonly abbreviated as <code>xsd:string</code>).
```

spec/index.html
```diff
@@ -875,7 +875,7 @@ <h2>Datatypes</h2>
     from the lexical space to the value space.</p>
 
   <p class="note"><a>Language-tagged strings</a> have the <a>datatype IRI</a>
-    <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>.
+    <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>).
```
hartig, line 878:
```suggestion
    <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>
    (commonly abbreviated as <code>rdf:langString</code>).
```

PR #32, opened 2023-04-10
Adds quoted triples and asserted triples from RDF-star CG report

### LINKED ISSUE #23: Quoted Triples
The [RDF-star CG Report](https://w3c.github.io/rdf-star/cg-spec/2021-12-17.html#concepts) describes the core concepts `RDF-star graph`, `RDF-star dataset`, and `RDF-star triple`. Additionally, it defines `quoted triple`, `asserted triple`, and `constituent terms`.

This issue is for either using these definitions in RDF 1.2 Concepts, or updating the definitions of core components in RDF 1.2 to include the notion of a triple as an RDF term so that a triple is inherently recursively defined.

gkellogg:
* Add quoted and asserted triple definitions.
* Add "classic" and "full" conformance levels.

Some license taken to make terminology more native to RDF, rather than a modification of RDF.

Fixes #23.
References #34 for potentially adding additional terminology.
Relates to w3c/rdf-star-wg#33 and w3c/rdf-star-wg#23

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/32.html" title="Last updated on Apr 22, 2023, 7:40 PM UTC (f24e320)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/32/9a3cf20...f24e320.html" title="Last updated on Apr 22, 2023, 7:40 PM UTC (f24e320)">Diff</a>


spec/index.html
```diff
@@ -167,6 +169,68 @@ <h3>Resources and Statements</h3>
...
+      which uses double angle brackets to enclose a <a>triple</a> serving as a <a>subject</a> or <a>object</a> resource:</p>
+
+    <pre id="ex-quoted-triple" class="example"
+         data-transform="updateExample"
+         data-content-type="text/turtle">
+      <!--
+      PREFIX :    <http://www.example.org/>
+
+      :employee38 :familyName "Smith" .
+      << :employee38 :jobTitle "Assistant Designer" >> :accordingTo :employee22 .
```
TallTed, line 205: Using optional whitespace to align conceptual columns helps enormously with comprehension. I'm not going to apply this throughout this PR, but I am going to suggest it here:
```suggestion
      :employee38                                       :familyName   "Smith" .
      << :employee38 :jobTitle "Assistant Designer" >>  :accordingTo  :employee22 .
```

spec/index.html
```diff
@@ -167,6 +169,68 @@ <h3>Resources and Statements</h3>
...
+
+    <pre id="ex-quoted-asserted-triple" class="example"
+         data-transform="updateExample"
+         data-content-type="text/turtle">
+      <!--
+      PREFIX :    <http://www.example.org/>
+
+      :employee38 :familyName "Smith" .
+      << :employee38 :jobTitle "Assistant Designer" >> :accordingTo :employee22 .
+      :employee38 :jobTitle "Assistant Designer" .
```
TallTed, line 229:
```suggestion
      :employee38                                       :familyName   "Smith" .
      << :employee38 :jobTitle "Assistant Designer" >>  :accordingTo  :employee22 .
      :employee38                                       :jobTitle     "Assistant Designer" .
```

spec/index.html
```diff
@@ -443,32 +519,68 @@ <h3>Triples</h3>
...
+    <p>A <dfn data-local-lt="quoted">quoted triple</dfn> is an <a>RDF triple</a>
+      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+
+    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
+      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
+
+    <p>A <dfn>complex triple</dfn> is an <a>RDF triple</a>
+      having a <a>quoted triple</a> as a component.</p>
+
+    <p>An <dfn>simple triple</dfn> is an <a>RDF triple</a>
```
TallTed, line 542:
```suggestion
    <p>A <dfn>simple triple</dfn> is an <a>RDF triple</a>
```

spec/index.html
```diff
@@ -443,32 +519,68 @@ <h3>Triples</h3>
...
+      that does not have a <a>quoted triple</a> as a component.</p>
+
+    <p>A <dfn class="export">complex graph</dfn> is
+      an <a>RDF graph</a> containing one or more <a>complex triples</a>.</p>
+
+    <p>A <dfn class="export">simple graph</dfn> is
+      an <a>RDF graph</a> composed entirely of <a>simple triples</a>.</p>
+
+    <p>The <dfn data-lt="constituent">constituent terms</dfn> (or simply constituents)
```
TallTed, line 551:
```suggestion
    <p>The <dfn data-lt="constituent">constituent terms</dfn> (or simply <a>constituents</a>)
```

spec/index.html
```diff
@@ -443,32 +519,68 @@ <h3>Triples</h3>
...
+      of a <a>triple</a>
+      is the set containing its <a>subject</a>, its <a>predicate</a>, its <a>object</a>,
+      plus all the <a>constituent terms</a> of its <a>subject</a> and/or its <a>object</a>
+      if they are themselves <a>quoted triples</a>.
+      By extension, we define the <a>constituent terms</a> of a <a>complex graph</a>
+      to be the union set of the <a>constituent terms</a> of all its triples.</p>
+
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
-      is the set of subjects and objects of triples in the graph.
+      is the set of subjects and objects of triples in the graph
```
TallTed, line 560:
```suggestion
      is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph
```

spec/index.html
```diff
@@ -443,32 +519,68 @@ <h3>Triples</h3>
...
       nor to a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
+
+    <p class="note">In a given <a>RDF graph</a>,
+      the same <a>triple</a> MAY be both <a>quoted</a> and <a>asserted</a>.</p>
+
+    <p class="note">The definition of <a>quoted triple</a> is recursive.
+      That is, a quoted triple can itself have a
+      subject or object component which is another quoted triple.
+      By their nature, cycles of quoted triples can not be created.</p>
```
TallTed, line 583:
```suggestion
      That is, a <a>quoted triple</a> can itself have a
      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
      However, by their nature, cycles of <a>quoted triples</a> cannot be created.</p>
```

TallTed (changes requested):

spec/index.html
```diff
@@ -167,6 +169,68 @@ <h3>Resources and Statements</h3>
...
+         data-content-type="text/turtle">
+      <!--
+      PREFIX :    <http://www.example.org/>
+
+      :employee38 :familyName "Smith" .
+      << :employee38 :jobTitle "Assistant Designer" >> :accordingTo :employee22 .
+      -->
+    </pre>
+
+    <p>After declaring a prefix so that IRIs can be abbreviated,
```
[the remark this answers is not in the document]
gkellogg: I'd rather avoid citing too much specifically in Turtle, because of layering concerns. My comment at the top of the PR is that we might want to use a diagram to avoid tying too closely to any specific notation, including Turtle.

spec/index.html
```diff
@@ -167,6 +169,68 @@ <h3>Resources and Statements</h3>
...
+      This use of a triple as the subject or object resource of another triple
+      so that we can say things about that triple is known as
+      <a data-cite='RDF12-SEMANTICS#reification'>reification</a>.
+      The concept of reification has always been part of RDF,
+      but expressing it in most RDF concrete syntaxes such as Turtle and N-Triples,
+      as well as processing or querying it with SPARQL,
+      has been verbose and cumbersome [[RDF-STAR-FOUNDATION]].</p>
+
+    <p>This specification extends RDF's conceptual data model,
+      which provides a compact alternative to <a data-cite='RDF12-SEMANTICS#reification'>standard RDF reification</a>.
```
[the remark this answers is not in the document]
gkellogg: Not really the same meaning of "classic", which describes conformance levels, and not reification.

spec/index.html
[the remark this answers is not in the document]
TallTed: Ah, I overlooked the suggestion of a diagram. That may help.

spec/index.html
```diff
@@ -167,6 +169,69 @@ <h3>Resources and Statements</h3>
...
+
+    <pre id="ex-quoted-triple" class="example"
+         data-transform="updateExample"
+         data-content-type="text/turtle">
+      <!--
+      PREFIX :    <http://www.example.org/>
+
+      :employee38                                      :familyName "Smith" .
+      << :employee38 :jobTitle "Assistant Designer" >> :accordingTo :employee22 .
+      -->
```
TallTed, line 206: If retaining Turtle, with or without diagram, a few other tweaks to help new people comprehend the example —
* it's easy to overlook or misunderstand the bare `:`; hence, `ex:`
* removing the extra line indents within the `<pre>`
* putting two space chars (instead of one) between columns
```suggestion
PREFIX  ex:  <http://www.example.org/>

ex:employee38                                           ex:familyName   "Smith" .
<< ex:employee38  ex:jobTitle  "Assistant Designer" >>  ex:accordingTo  ex:employee22 .
      -->
```

spec/index.html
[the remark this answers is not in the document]
gkellogg: I'll wait for some more opinions and may do a diagram in a subsequent PR.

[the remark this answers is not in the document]
gkellogg: I think @pchampin may have figured out how to create a diagram illustrating the use of quoted triples in complex statements, and also where the quoted triple is also asserted, perhaps with some SVG that can be re-purposed.

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
     <p>An RDF triple is conventionally written in the order subject,
       predicate, object.</p>
 
+    <p>A <dfn data-local-lt="quoted">quoted triple</dfn> is an <a>RDF triple</a>
+      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+
+    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
+      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
+
+    <p>A <dfn>complex triple</dfn> is an <a>RDF triple</a>
```
afs, line 540: Remove.

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
+
+    <p>A <dfn class="export">simple graph</dfn> is
+      an <a>RDF graph</a> composed entirely of <a>simple triples</a>.</p>
+
+    <p>The <dfn data-lt="constituent">constituent terms</dfn> (or simply <a>constituents</a>)
+      of a <a>triple</a>
+      is the set containing its <a>subject</a>, its <a>predicate</a>, its <a>object</a>,
+      plus all the <a>constituent terms</a> of its <a>subject</a> and/or its <a>object</a>
+      if they are themselves <a>quoted triples</a>.
+      By extension, we define the <a>constituent terms</a> of a <a>complex graph</a>
```
afs, line 557: Do we need this extension in RDF Concept in the primary section about triple?

If used in the semantics, put the referable terminology there.

afs (changes requested):

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
 
       <li>the <dfn>predicate</dfn>, which is an <a>IRI</a></li>
 
       <li>the <dfn>object</dfn>, which is an <a>IRI</a>,
-      a <a>literal</a> or a <a>blank node</a></li>
+      a <a>literal</a>, a <a>blank node</a>, or a <a>quoted triple</a></li>
     </ul>
 
     <p>An RDF triple is conventionally written in the order subject,
       predicate, object.</p>
```
afs, line 532: Add 
```
<p>RDF triple is usually written <i>triple</i>.<p>
```

spec/index.html
```diff
@@ -167,6 +169,69 @@ <h3>Resources and Statements</h3>
...
+      namely the triple "`employee38` has a `jobTitle` of 'Assistant Designer'".
+      This use of a triple as the subject or object resource of another triple,
+      so that we can say things about the triple so used, is known as
+      <a data-cite='RDF12-SEMANTICS#dfn-reification'>reification</a>.
+      The concept of reification has always been part of RDF,
+      but expressing it in most RDF concrete syntaxes such as Turtle and N-Triples,
+      as well as processing or querying it with SPARQL,
+      has been verbose and cumbersome [[RDF-STAR-FOUNDATION]].</p>
+
+    <p>This specification extends RDF's conceptual data model,
```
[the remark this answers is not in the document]
gkellogg: Maybe some of this narrative can be moved to What's New in RDF 1.2.

spec/index.html
```diff
@@ -427,14 +492,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+
+  <ul>
+    <li><dfn class="no-export lint-ignore">Classic</dfn> conformance
+      only supports <a>simple graphs</a> or <a>simple datasets</a>,
+      and does not allow the use of <a>quoted triples</a>.</li>
+    <li><dfn class="no-export lint-ignore">Full</dfn> conformance
+      supports <a>complex graphs</a> or <a>complex datasets</a>,
+      and allows the use of <a>quoted triples</a>.
+      Such input syntaxes include
+      [[RDF12-N-TRIPLES]], [[RDF12-N-QUADS]], [[RDF12-TURTLE]], and [[RDF12-TRIG]].</li>
```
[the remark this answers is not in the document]
gkellogg: I'm gong to add a note in SOTD that explains that these terms are tentative placeholders, some of which may not continue going forward.

spec/index.html
gkellogg:
```suggestion 
    <p>An RDF triple is conventionally written in the order subject,
      predicate, object.</p>
      
    <p>RDF triple is usually written "<a>triple</a>".<p>
```

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
 
       <li>the <dfn>object</dfn>, which is an <a>IRI</a>,
-      a <a>literal</a> or a <a>blank node</a></li>
+      a <a>literal</a>, a <a>blank node</a>, or a <a>quoted triple</a></li>
     </ul>
 
     <p>An RDF triple is conventionally written in the order subject,
       predicate, object.</p>
 
+    <p>A <dfn data-local-lt="quoted">quoted triple</dfn> is an <a>RDF triple</a>
```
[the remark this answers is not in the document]
gkellogg:
```suggestion
    <p>A <dfn data-local-lt="quoted">quoted triple</dfn> is an <a>RDF term</a> representing a <a>triple</a>
```

spec/index.html
gkellogg:
```suggestion
      predicate, object.</p>

    <p>RDF triple is usually written "<a>triple</a>".<p>
```

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
+      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+
+    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
+      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
+
+    <p>A <dfn>complex triple</dfn> is an <a>RDF triple</a>
+      having a <a>quoted triple</a> as a component.</p>
+
+    <p>A <dfn>simple triple</dfn> is an <a>RDF triple</a>
```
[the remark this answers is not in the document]
gkellogg: Maybe change from "complex" and "plain" to "red" and "blue" to make clear that they are just placeholders, and describe as such in an SOTD note.

spec/index.html
gkellogg: I think we should recognize "constituent terms" and "nodes". We may not need "constituent terms" either of a triple or a graph, at all, but it was in the CG report. But, the term "nodes" would by that same logic to better be in semantics than concepts.

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
+      plus all the <a>nodes</a> of those <a>subjects</a> and/or its <a>objects</a>,
+      recursively, for <a>complex triples</a>.
       It is possible for a predicate IRI to also occur as a node in
       the same graph.</p>
 
-    <p><a>IRIs</a>, <a>literals</a> and
-      <a>blank nodes</a> are collectively known as
+    <p><a>IRIs</a>, <a>literals</a>,
+      <a>blank nodes</a>, and <a>quoted triples</a> are collectively known as
       <span id="dfn-rdf-terms"><!-- obsolete term--></span><dfn data-lt="rdf term">RDF terms</dfn>.</p>
```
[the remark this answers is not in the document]
gkellogg: For consistency, we define singular terms, for which plural versions are automatically deterred via ReSpec, so the term identifier is "dfn-rdf-term", the span was added so that any references to "dfn-rdf-terms" will still go to the right place.

When using ReSpec to find specs defining terms, we want the singular version to appear.

spec/index.html
```diff
@@ -443,32 +520,68 @@ <h3>Triples</h3>
...
-      and <a>blank nodes</a> are distinct and distinguishable.
+    <p><a>IRIs</a>, <a>literals</a>,
+      <a>blank nodes</a>, and <a>quoted triples</a> are distinct and distinguishable.
       For example, <code>http://example.org/</code> as a string literal
       is neither equal to <code>http://example.org/</code> as an IRI,
       nor to a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
+
+    <p class="note">In a given <a>RDF graph</a>,
+      the same <a>triple</a> MAY be both <a>quoted</a> and <a>asserted</a>.</p>
```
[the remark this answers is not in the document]
gkellogg: Given that we're in an informative note, there is no normative meaning to "MAY", so it can be written "may".

```suggestion
      the same <a>triple</a> may be both <a>quoted</a> and <a>asserted</a>.</p>
```

gkellogg:
I added something to the SOTD; at this point I'd rather not remove terminology until there is more discussion. We can create some new sections, possibly informative, to move around definitions as a next step. It also needs to be rebased on recent commits to main, which may make some existing comments obsolete requiring them to be re-considered, but we can wait on that until we're closer to concensus.

spec/index.html
gkellogg: Same could be said of "graph" and "dataset". Note that the exported terms remain `dfn-rdf-triple`, `dfn-rdf-graph`, and `dfn-rdf-dataset`, but we can use various aliasing mechanism internally, such as `data-local-lt`.

spec/index.html
```diff
@@ -427,14 +581,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
     but can conform to such other specifications that normatively
     reference terms defined here.</p>
+
+  <p>This specification establishes two conformance levels:</p>
+
+  <ul>
+    <li><dfn class="no-export lint-ignore">Full</dfn> conformance
+      supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
+      containing <a>quoted triples</a>.
+      Such concrete syntaxes include
```
hartig, line 591:
```suggestion
      Concrete syntaxes in which such graphs and datasets can be expressed include
```

spec/index.html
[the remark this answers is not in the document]
gkellogg: I updated this section and replaced the Turtle examples with diagrams. Although they won't render through PR Preview, the diagrams should be viewable in the files tab, or through the URL comments within the figures.

spec/index.html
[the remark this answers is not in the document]
gkellogg: The mention of reification has been removed as a general reorganization of this section.

spec/index.html
```diff
@@ -65,22 +82,98 @@
...
+      <dd>Originally defined as <a data-cite="?RDF-STAR-CG#dfn-quoted">quoted triple</a>
+        in [[?RDF-STAR-CG]] as an <a data-cite="?RDF-STAR-CG#dfn-triple">RDF-star triple</a>
+        which is the subject or object of another
+        <a data-cite="?RDF-STAR-CG#dfn-triple">RDF-star triple</a>.
+        Alternate names include <em>triple term</em>.</dd>
+      <dt><a>asserted triple</a></dt>
+      <dd>Originally defined as <a data-cite="?RDF-STAR-CG#dfn-asserted">asserted triple</a>
+        in [[?RDF-STAR-CG]] as an <a data-cite="?RDF-STAR-CG#dfn-triple">RDF-star triple</a>
+        which is an element of an <a data-cite="?RDF-STAR-CG#dfn-graph">RDF-star graph</a>.</dd>
+      <dt><a>complex triple</a></dt>
```
afs, line 123: "complex*" "plain*" are not in the CG report.

Nor is compliance.
Profile is, referring to [Content Negotiation by Profile](https://www.w3.org/TR/dx-prof-conneg/#introduction).

afs:
> at this point I'd rather not remove terminology until there is more discussion.

(referring to plain/complex)

The introduced of this terminology, which is not in the CG report, is of significant size, does not reflect the discussion and concerns of the telecon (2023-04-13).  The terminology is described as "adapted from the CG report" - that's misleading.

I'd rather not put terminology into the drafts that is highlighting one position over another. In this case, another possibility is not having top-level terminology or having terminology to describe one situation.

(referring to w3c/rdf-star-wg#19, w3c/rdf-star-wg#23)

Put in issue markers without opinion or direction.

gkellogg:
I think an issue marker for everything other than “quoted triple” is a reasonable interim solution.

spec/index.html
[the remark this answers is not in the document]
gkellogg: Changed to "containing _quoted triples_".

spec/index.html
gkellogg: Whether it's "profile" of "conformance" is still being considered in https://github.com/w3c/rdf-star-wg/issues/19. Content Negotiation would be outside the scope of RDF Concepts. I think the point is that some implementations may not support quoted triples, and there should probably be a way to discuss this. We can either re-visit this when the conversation is more complete, or remove the language altogether until there is something more substantial.

spec/index.html
```diff
@@ -427,14 +498,26 @@ <h2>RDF Documents and Syntaxes</h2>
     <em>RDF 1.2 Concepts and Abstract Syntax</em>,
     but can conform to such other specifications that normatively
     reference terms defined here.</p>
+
+  <p>This specification establishes two conformance levels:</p>
```
afs, line 502: The WG has not agreed that nor the words used. It's WIP.

"Profiles" has also been mentioned.

Also - "This specification" would be the semantics.

At a minimum this text needs a note that it is tentative.
As concepts does not need this language, and the dfn's are unstable, removing it for now would be better.

spec/quoted-triple.svg
afs, line 1: I'm note sure this diagram works. The two lines are different things.

afs (changes requested):

spec/index.html
```diff
@@ -167,6 +186,58 @@ <h3>Resources and Statements</h3>
...
+      asserted within an <a>RDF graph</a>.
+      This allows statements to be made about relationships that may be contradictory.
+      For a <a>quoted triple</a> to be asserted,
+      it must also appear in a graph as an <a>asserted triple</a>.</p>
+
+    <p>The following diagram represents a statement with an unasserted <a>quoted triple</a>
+      as the subject.</p>
+
+    <figure id="fig-quoted-triple">
+      <a href="quoted-triple.svg">
```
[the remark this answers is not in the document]
gkellogg: Yes, a shortcoming of PR Preview. But, you can either see it in the file view, or using the Google drawing referenced in the comment: https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8.

spec/index.html
```diff
@@ -167,6 +186,58 @@ <h3>Resources and Statements</h3>
...
+             alt="An RDF graph containing an triple referencing an unasserted quoted triple as the subject"
+             style="width:70%"
+             aria-describedby="fig-quoted-triple-alt"/>
+      </a>
+      <figcaption id="fig-quoted-triple-alt">
+        An RDF graph containing an triple referencing an unasserted quoted triple (with grey background) as the subject.
+      </figcaption>
+    </figure>
+
+    <p>The same graph can be described where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
```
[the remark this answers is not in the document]
gkellogg: It is hard to convey the subtleties in a diagram, and it requires being able to perceive and understand the difference between the quoted/unasserted triple being shared and the quoted/asserted triple being unshaded. This was taken from presentations @pchampin and @hartig have done, and maybe there's a better way to do it.


```suggestion
    <p>A variation on the graph shown in <a href="#fig-quoted-triple"></a> can be described where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
```

spec/index.html
```diff
@@ -427,14 +498,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+  <ul>
+    <li><dfn class="no-export lint-ignore">Full</dfn> conformance
+      supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
+      containing <a>quoted triples</a>.
+      Concrete syntaxes in which such graphs and datasets can be expressed include
+      [[RDF12-N-TRIPLES]], [[RDF12-N-QUADS]], [[RDF12-TURTLE]], and [[RDF12-TRIG]].</li>
+    <li><dfn class="no-export lint-ignore">Classic</dfn> conformance
+      only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>,
+      that do not allow the use of use <a>quoted triples</a>.</li>
+  </ul>
```
gkellogg, line 513:
```suggestion
  </ul>
  <p class="ednote">The conformance levels described above are tentative,
    and still the subject of group discussion.
    As an alternative, "profiles" rather than conformance levels,
    may be abandoned, or described in another specification.</p>
```

spec/index.html
gkellogg: See note added below. My opinion is that if it relates to the data model/abstract syntax, it should be described in Concepts rather than Semantics. I suggested elsewhere that the text could just be removed from this PR and taken up another time. But, the notion that there may be some variation allowed or supported is probably worth signaling sooner rather than later.

spec/index.html
```diff
@@ -427,14 +497,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+
+  <ul>
+    <li><dfn class="no-export lint-ignore">Full</dfn> conformance
+      supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
+      containing <a>quoted triples</a>.
+      Concrete syntaxes in which such graphs and datasets can be expressed include
+      [[RDF12-N-TRIPLES]], [[RDF12-N-QUADS]], [[RDF12-TURTLE]], and [[RDF12-TRIG]].</li>
+    <li><dfn class="no-export lint-ignore">Classic</dfn> conformance
+      only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>,
+      that do not allow the use of use <a>quoted triples</a>.</li>
```
TallTed, line 511:
```suggestion
      that do not include <a>quoted triples</a>.</li>
```

spec/index.html
TallTed:
```suggestion
  </ul>
  <p class="ednote">The conformance levels described above are tentative,
    and still the subject of group discussion. An alternative to conformance
    levels, "profiles", may be adopted instead, abandoned, or described in 
    another specification.</p>
```

spec/index.html
```diff
@@ -443,32 +525,50 @@ <h3>Triples</h3>
...
 
       <li>the <dfn>object</dfn>, which is an <a>IRI</a>,
-      a <a>literal</a> or a <a>blank node</a></li>
+      a <a>literal</a>, a <a>blank node</a>, or a <a>quoted triple</a></li>
     </ul>
 
     <p>An RDF triple is conventionally written in the order subject,
       predicate, object.</p>
 
+    <p>RDF triple is usually written "<a>triple</a>".<p>
```
TallTed, line 539:
```suggestion
    <p>"RDF triple" is usually written "<a>triple</a>".<p>
```

spec/index.html
```diff
@@ -443,32 +525,50 @@ <h3>Triples</h3>
...
+      <a>blank nodes</a>, and <a>quoted triples</a> are collectively known as
       <span id="dfn-rdf-terms"><!-- obsolete term--></span><dfn data-lt="rdf term">RDF terms</dfn>.</p>
 
-    <p><a>IRIs</a>, <a>literals</a>
-      and <a>blank nodes</a> are distinct and distinguishable.
+    <p><a>IRIs</a>, <a>literals</a>,
+      <a>blank nodes</a>, and <a>quoted triples</a> are distinct and distinguishable.
       For example, <code>http://example.org/</code> as a string literal
       is neither equal to <code>http://example.org/</code> as an IRI,
       nor to a blank node with the <a>blank node identifier</a>
```
TallTed, line 562:
```suggestion
      is equal to neither <code>http://example.org/</code> as an IRI,
      nor a blank node with the <a>blank node identifier</a>
```

spec/index.html
```diff
@@ -443,32 +525,50 @@ <h3>Triples</h3>
...
       <code>http://example.org/</code>.</p>
+
+    <p class="note">In an <a>RDF graph</a>,
+      a <a>triple</a> may occur as both a <a>quoted</a> triple term and as an <a>asserted</a> triple.</p>
```
TallTed, line 566:
```suggestion
      a <a>triple</a> may occur as either, neither, or both a <a>quoted</a>
      triple term and an <a>asserted</a> triple.</p>
```

TallTed (changes requested):

spec/index.html
gkellogg: Hard to see how "neither" factors in here, as a triple which is neither quoted nor asserted is not in the graph at all, so isn't really a triple of discourse.

TallTed: Oh, you're right. I guess "either or both" covers my concern (which is the possible implication of "all or nothing"; i.e., "either" disallowed). I've revised my suggestion.

TallTed: I've edited mine to match your latest, @gkellogg. Either may be committed to the PR to make me happy.

afs (approved):

spec/index.html
```diff
@@ -443,32 +529,50 @@ <h3>Triples</h3>
...
+      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+
+    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
+      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
+
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
-      is the set of subjects and objects of triples in the graph.
+      is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph
+      plus all the <a>nodes</a> of those <a>subjects</a> and/or its <a>objects</a>,
+      recursively, for <a>quoted triples</a>.
```
hartig, line 554: Two things here:

1. This new way of defining the notion of "nodes of an RDF graph" depends on a notion of "nodes of a quoted triple" which is undefined at the moment. This issue can be fixed by defining this notion (see how the notion of constituent terms is defined in the CG report). However, I don't think that's even necessary because ...
3. ... I don't think we should change the definition of the notion of "nodes of an RDF graph" in this way. That's quite a substantial change. Are we sure that we don't break anything by this change? I mean, if this notion is used in other definitions somewhere in our documents, this change may affect these other definitions.

spec/index.html
```diff
@@ -443,32 +529,50 @@ <h3>Triples</h3>
...
+      nor a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
+
+    <p class="note">In an <a>RDF graph</a>,
+      a <a>triple</a> may occur as both a <a>quoted</a> triple term and as an <a>asserted</a> triple.</p>
+
+    <p class="note">The definition of <a>quoted triple</a> is recursive.
+      That is, a <a>quoted triple</a> can itself have a
+      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
+      However, by their nature, cycles of <a>quoted triples</a> cannot be created.</p>
```
hartig, line 575:
```suggestion
      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
```

hartig (changes requested):

gkellogg:
Squashed and rebased to address conflicts and simplify history.

spec/index.html
gkellogg: Can you suggest a better alternative. Alternatively, we could add an issue marker that the consideration of nodes contained within quoted triples being considered as "nodes of an RDF graph" is open and subject to further deliberation by the group.

pfps:
Here are the relevant definitions for RDF graphs and triples from the PR.

> An RDF graph is a set of RDF triples.
> An RDF triple consists of three components:
> - the subject, which is an IRI, a blank node, or a quoted triple
> - the predicate, which is an IRI
> - the object, which is an IRI, a literal, a blank node, or a quoted triple
> An asserted triple is an RDF triple that is an element of an RDF graph.
> A quoted triple is an RDF triple that is used as the subject or object of another triple.
> In an RDF graph, a triple may occur as either a quoted triple, an asserted triple, or both.

The above definitions do not forbid triples that contain themselves. 

Neither cycles of quoted triples nor creation is defined so the prohibition below is meaningless.

> However, by this definition, cycles of quoted triples cannot be created.

spec/index.html
afs: I agree with @hartig  - we shouldn't change "nodes in an RDF graph". Section 1.1 is putting "graph" in general graph terminology context. We need to be careful because this is "node" in the sense of a graph-theory graph. nodes-arcs / vertex-edge.

I also would like to know if we need the terminology? "constituent" is used in the CG report in the mapping section, which is unlikely to be in RDF 1.2 and even if it is, is limited to semantics; and it is used in the "merge" text which is also not in concepts.

SPARQL uses general-graph nodes for property paths (specifically, "node set", which it defines).

Until proven to need terminology, it seems safer to not have i, allowing semantics to have in its scope.

afs:
> The above definitions do not forbid triples that contain themselves.

We ought to clarify to make sure this isn't a possible reading. 

@pfps - do you have a suggestion?

A temporary measure, for FPWG,  is to say (informally) "Cycles are not permitted".
Or (temporary) note that all RDF syntaxes can not describe RDF data with cycles.

spec/index.html
pchampin: I also think that we should not change the definition of "node" at this point.

pchampin:
I agree that the current definition of 'RDF triple' leaves the door open for "self-containing" triples, and that the notion of "cycles of quoted triples" is confusing...

It seems to me that the [definition in the CG report](https://www.w3.org/2021/12/rdf-star.html#dfn-triple) is a better way to avoid "self-containing" triples (@pfps do you agree?), but it relies on the distinction between "RDF triple" and "RDF-star triple"... so in order to reuse this phrasing, we would need a distinction between "simple/complex" triples... :-/

pfps:
@pchampin Yes the CG definition rules out triples that contain themselves, so it is different than the definition in this PR.  Whether this is better depends on whether one wants to allow such triples.

spec/index.html
```diff
@@ -441,32 +527,51 @@ <h3>Triples</h3>
...
+      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+
+    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
+      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
+
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
-      is the set of subjects and objects of triples in the graph.
+      is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph
+      plus all the <a>nodes</a> of those <a>subjects</a> and/or its <a>objects</a>,
+      recursively, for <a>quoted triples</a>.
```
gkellogg, line 552:
```suggestion
      is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph.
```

hartig:
It is possible to adapt the definition of the CG report as follows. 

An RDF triple is a 3-tuple defined recursively as follows:
* if s is an IRI or a blank node, p is an IRI, o is an IRI, a blank node or a literal, then (s, p, o) is an RDF triple;
* if t and t' are RDF triples, s is an IRI or a blank node, p is an IRI, o is an IRI, a blank node or a literal, then (t, p, o), (s, p, t) and (t, p, t') are RDF triples.

(Since I am just writing this on my phone, I am posting this just as text rather than as a directly commit-able suggestion. However, I can create such a suggestion comment later (tomorrow) if you agree with this direction for the definition.)

hartig:
@gkellogg I have [my proposed change](https://github.com/w3c/rdf-concepts/pull/32#issuecomment-1518324303) properly written in my local copy of the branch now. However, I cannot put it here as a directly commit-able suggestion because it also changes line 440 ("&lt;p>An &lt;dfn data-local-lt="triple">RDF t...") which is too many lines away from the changes in this PR and, thus, I cannot include it when selecting lines in the "Files changed" tab of this PR. As an alternative, I could simply push a commit with the change to the branch of this PR. Would that be okay with you? Another option is to merge this PR now (in which case I would approve it), and then I create another PR with the change. Which one do you prefer?

gkellogg:
Yes, I would just commit to the branch. You can always clarify with a comment after the fact, if it seems necessary.

hartig:
Thanks @gkellogg! I have pushed the commit (https://github.com/w3c/rdf-concepts/pull/32/commits/f24e320ccae1a4902eb7637835775d001c69eb74).

hartig (approved):

PR #33, opened 2023-04-16
add note about rdf:PlainLiteral

domel:
`rdf:PlainLiteral` is no longer used in RDF. It is still available at http://www.w3.org/1999/02/22-rdf-syntax-ns#
see https://github.com/w3c/rdf-star-wg/issues/52

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/33.html" title="Last updated on Apr 16, 2023, 1:32 PM UTC (7e81b58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/33/9a3cf20...7e81b58.html" title="Last updated on Apr 16, 2023, 1:32 PM UTC (7e81b58)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -852,6 +852,9 @@
     The list of datatypes supported by an implementation is determined
     by its <a>recognized datatype IRIs</a>.</p>
 
+  <p class="note"><code>rdf:PlainLiteral</code> [[RDF-plain-literal]] is no longer used in RDF 
+    itself, but that it continues to be part of RIF [[RIF-RDF-OWL]] and OWL2 [[OWL2-mapping-to-RDF]].</p>
+
   <p>A <dfn>datatype</dfn> consists of a <a>lexical space</a>,
     a <a>value space</a> and a <a>lexical-to-value mapping</a>, and
     is denoted by one or more <a>IRIs</a>.</p>


pfps:
I believe that rdf:PlainLiteral was never used in RDF.   It is in neither RDF 1.1 Concepts nor RDF Concepts nor RDF Schema 1.1 nor RDF Schema.   So putting a note here seems to me to be counterproductive.

Given the history of rdf:PlainLiteral it is probably not necessary to change the RDF namespace file either.

afs:
(I hope this "needs discussion" can be resolved in this PR discussion.)

>Given the history of rdf:PlainLiteral it is probably not necessary to change the RDF namespace file either.

@pfps -
https://github.com/w3c/rdf-schema/issues/10#issuecomment-1507370363

I read that as suggesting adding a comment, at least, to the namespace file.

domel:
In my opinion, the namespace file should reflect the description in the specification. If `rdf:PlainLiteral` is in the RDF namespace, it would make sense to mention it in the spec as well. If we leave it as is, it is a misconception that `rdf:PlainLiteral` has the same status as e.g. `rdf:HTML`.

gkellogg (commented):
`rdf:PlainLiteral` never was part of RDF, at least not part of core RDF Concepts or Schema AFAICT. I think the preface of the note "`rdf:PlainLiteral` is no longer used in RDF itself" is, therefore, inaccurate. What's true is that it is not useful in RDF 1.2 (nor in RDF 1.1), but is still used in some current notes and recommendations (RIF and OWL2). As it's in the namespace, it I think it's a good policy to say something about it. It is probably worth informatively referencing the [rdf:PlainLiteral Recommendation](https://www.w3.org/TR/rdf-plain-literal/).

afs:
Unfortunately, http://www.w3.org/1999/02/22-rdf-syntax-ns has accumulated several concepts that are not in the RDF Data Model. Some aren't even normative. They do at least have `rdfs:isDefinedBy`.

In other words, the namespace document is not restricted to RDF Concepts.

As it is not mentioned in concepts, adding text about it to say "don't use it" seems to be counter productive.

We could add something to rdf-new, the RDF 1.2 specific overview.

Can we deprecate the rdf:PlainLiteral spec? i.e. put in a W3C spec overlay that alerts the reader they are looking at out-of-date information.

The only use of RDF-PLAIN-LITERAL is in rdf-semantics and the mention is only because rdf:PlainLiteral exists.

pfps:
As far as I can tell nothing about rdf:PlainLiteral has changed since RDF 1.1 came out.  I view this as strong evidence that no changes related to rdf:PlainLiteral need to be made to Concepts or Schema.

hartig:
I agree. Since `rdf:PlainLiteral` is not mentioned anywhere else in this document, there is no need for adding such a note. Instead, a note about the history of `rdf:PlainLiteral` may be added to rdf-new and the rdf:PlainLiteral spec should be deprecated if that is possible for our group to do.

pfps:
It appears that everyone is in agreement that no change is needed to RDF Concepts.   Who should close this PR without merging it?

PR #40, opened 2023-05-07
improve display on mobile phones

domel:
Improvement display on mobile phones and other small devices + whitespaces in styles

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/40.html" title="Last updated on May 18, 2023, 4:00 PM UTC (94f9d8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/40/5913afa...94f9d8a.html" title="Last updated on May 18, 2023, 4:00 PM UTC (94f9d8a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -60,10 +60,11 @@
     figure { text-align: center; }
     table.simple td, table th { border: 1px solid #ddd; padding: 0.2em 0.5em; }
     ol ol { list-style-type: lower-latin; }
-    .grammar td { font-family: monospace;}
-    .grammar-literal { color: gray;}
-    code {color: #ff4500;}  /* Old W3C Style */
-    div.abnf {margin-left: 1em;}
+    .grammar td { font-family: monospace; }
+    .grammar-literal { color: gray; }
+    code { color: #ff4500; }  /* Old W3C Style */
+    div.abnf { margin-left: 1em; }
+    table { word-break: break-all; max-width: 80vw; }
   </style>
 </head>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1062,7 +1062,7 @@
       information.</p>
 
     <table class="simple">
-      <caption>A list of the RDF-compatible XSD types, with short descriptions"</caption>
+      <caption>A list of the RDF-compatible XSD types, with short descriptions</caption>
       <tr><th></th><th>Datatype</th><th>Value space (informative)</th></tr>
 
       <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode character strings)</td></tr>
--- a/spec/index.html
+++ b/spec/index.html
@@ -64,8 +64,12 @@
     .grammar-literal { color: gray; }
     code { color: #ff4500; }  /* Old W3C Style */
     div.abnf { margin-left: 1em; }
+    @media (max-width: 850px) {
+        table { overflow-wrap: anywhere !important; }
+        .simple th, .simple td { font-size: 14px; }
+    }    
     @media (max-width: 767px) {
-        table { word-break: break-all; max-width: 80vw; }
+        table { word-break: normal; max-width: 80vw;}
     }
   </style>
 </head>
--- a/spec/index.html
+++ b/spec/index.html
@@ -69,7 +69,7 @@
         .simple th, .simple td { font-size: 14px; }
     }    
     @media (max-width: 767px) {
-        table { word-break: normal; max-width: 80vw;}
+        table { word-break: normal; /*max-width: 80vw;*/}
     }
   </style>
 </head>
--- a/spec/index.html
+++ b/spec/index.html
@@ -64,7 +64,9 @@
     .grammar-literal { color: gray; }
     code { color: #ff4500; }  /* Old W3C Style */
     div.abnf { margin-left: 1em; }
-    table { word-break: break-all; max-width: 80vw; }
+    @media (max-width: 767px) {
+        table { word-break: break-all; max-width: 80vw; }
+    }
   </style>
 </head>
 


gkellogg (approved):

hartig:
@domel Do you have a concrete example where this change makes a difference and what this difference looks like?

domel:
@hartig yes, actually on any mobile device this page (and others display badly). The most common issue is with tables that extend beyond the right margin. The result is a scroll on the x-axis, which is very annoying and unreadable. You can check this on your smartphone or in a desktop browser by pressing F12.

hartig:
What do you mean by "this page"? This one here on github? That one looks fine in my mobile browser. Here's a screenshot:

Screenshot_20230508-112641

Can you please point out concrete tables in the actual spec draft that you have a problem with, ideally using screenshots (and, ideally, screenshots before and after the changes in this PR).

domel:
I mean spec document. All tables has problems. If uou want I can take a screenshot later.

domel:
Screenshot_2023-05-08-12-40-27-563_org mozilla firefox

domel:
And after this PR

domel:
Screenshot_2023-05-08-12-43-21-447_org mozilla firefox
After this PR

domel:
I'm a bit surprised that you don't see the differences between these two websides of spec.  I specifically tested it on 43 different devices and it displays the same bad everywhere.  It is worth remembering that nowadays more traffic is generated by mobile browsers than by desktop browsers.  IMO it's worth taking care of these users.  Maybe I'm weird or maybe it's because I used to be a front-end developer.

pfps:
Just because there is more traffic from mobile browsers in general doesn't mean that there is any significant traffic to W3C recommendations from mobile browsers.  If this change makes viewing worse from larger-screen devices then it needs to be rethought.

pfps:
Just because there is more traffic from mobile browsers in general doesn't mean that there is any significant traffic to W3C recommendations from mobile browsers.  If this change makes viewing worse from larger-screen devices then it needs to be rethought.

domel:
@pfps If you think that the current desktop view is not good enough, I can improve it so that different styles are for desktop and different for mobile devices.

domel:
you can check again

afs:
There are three or four cases here:
* PC Full screen (wide)
* PC - narrow window
* Tablet (landscape/portrait?)
* Phone

Concepts isn't a good example - it has two tables, both of which are additionally style with ".simple" but I do see a worsening:

PC - narrow window: Chrome : different tabs, same window.

`main`
Screenshot from 2023-05-08 17-34-57

`rwd`
image

SPARQL Query:
A little narrowed than the concepts screenshots.

`main`
Screenshot from 2023-05-08 17-56-23

`rwd`
Screenshot from 2023-05-08 17-56-47

Would `break-word` would be better? It then should not spill the right-hand-side and it does not break the rule numbering at least for PC-narrow.

I've uploaded SPARQL query from "main", from "rwd" and with work-break to make it easier to try out at different widths and different devices:

https://afs.github.io/sparql-respec/Screen/sparql-query/spec/index.html#sparqlGrammar

https://afs.github.io/sparql-respec/Screen/sparql-query-rwd/spec/index.html#sparqlGrammar

https://afs.github.io/sparql-respec/Screen/sparql-query-word-break/spec/index.html#sparqlGrammar

afs:
@domel - the latest changes make PC/narrow worse. "word-break: break-word"; used to work but inside media it does not for PC/narrow - at the width where your PR starts breaking the rule numbering in the SPARQ grammar.

gkellogg:
I think `word-break` probably works better. There's only so far we can go to make these tables look reasonable on a phone, but some provisions for different breaks probably makes sense on sufficiently small windows, and the media query is the right way to go.

afs:
The "media" seems to stop zooming out to see the table, whereas the current set up does (Chrome on a OnePlus6).

TallTed:
The [Web-Developer Extension](https://chrispederick.com/work/web-developer/) can be helpful for previewing such things in various screensizes (click to activate the extension, then the **Resize** tab, then **View Responsive Layouts**). If the preset sizes are not sufficient, you can **Edit Resize Dimensions...**.

I would not expect many people with the tiniest phone screens to try to read docs like these thereon. Tablets, large phone screens, desktop, and printout should be our primary concerns.

afs:
@Tallted  - no longer relevant, due to later changes, and [the setting is deprecated](https://github.com/w3c/sparql-query/pull/71#discussion_r1187798919).

https://afs.github.io/sparql-respec/Screen/

domel:
> I would not expect many people with the tiniest phone screens to try to read docs like these thereon. Tablets, large phone screens, desktop, and printout should be our primary concerns.

I may not be representative but I read W3C specs on a smartphone 90% of the time.

TallTed:
@domel — What's your smartphone's screen size & resolution? 

My current iPhone 12 Pro Max is 6.7″, and 1284 x 2778 pixels at ~458 ppi. I *could* read most W3 TRs on it, but I find the touch environment ill-suited to my web surfing.

The original iPhone was ~3.5", 320 x 480 pixels at 163 ppi, and I think very few — even fewer than those who would use their phone to read W3 TRs — are still using such. These would be the people "with the tiniest phone screens" I referred to.

I also think that those who are most likely to read W3 TRs on phone screens are also the most likely to have the largest phone screens with the most pixels available.

spec/index.html
```diff
@@ -60,10 +60,17 @@
...
-    .grammar td { font-family: monospace;}
-    .grammar-literal { color: gray;}
-    code {color: #ff4500;}  /* Old W3C Style */
-    div.abnf {margin-left: 1em;}
+    .grammar td { font-family: monospace; }
+    .grammar-literal { color: gray; }
+    code { color: #ff4500; }  /* Old W3C Style */
+    div.abnf { margin-left: 1em; }
+    @media (max-width: 850px) {
+        table { overflow-wrap: anywhere !important; }
```
TallTed, line 68: What effect does `!important` have here?

TallTed:
The most complex table in this doc is labeled **A list of the RDF-compatible XSD types, with short descriptions"** (and that trailing quotation mark should be deleted), in [§5.1 The XML Schema Built-in Datatypes](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/40.html#xsd-datatypes). This table currently exhibits mid-word breakage enabling the table to fit within the smaller windows, instead of keeping words unbroken and overflowing window width.

I'm wondering, why are the CSS changes not the same for all documents? Possibly because they were created over time, and hence got different CSS to start? It seems to me it would be better to make the same CSS apply to all the documents, though I'm sure this will entail a fairly large piece of work.

gkellogg (approved):

Issue #43, opened 2023-05-22
Add Note about xsd:float and xsd:double Limitations


jmkeil:
What about adding a remark in section [The XML Schema Built-in Datatypes](https://w3c.github.io/rdf-concepts/spec/#xsd-datatypes) to increase the awareness of the limitations of the values spaces of `xsd:float` and` xsd:double` (see [EasierRDF#82](https://github.com/w3c/EasierRDF/issues/82#issuecomment-1436738861) and https://doi.org/10.1007/978-3-031-06981-9_10)? Something like:

* add in the last column of the table
  >(but excluding many decimal numbers like `0.1`)

  or
  >(but not all decimal numbers)
* somewhere after the table add a NOTE box like:
  > The Lexical-to-value mapping of `xsd:float` and` xsd:double` performs a rounding of the lexical value to the nearest value representable in [IEEE 754-2008](http://ieeexplore.ieee.org/servlet/opac?punumber=4610933) binary floating point representation of the according precission. Therefore, many decimal numbers are not exactly represented by `xsd:float` and` xsd:double`. For example,
  > `ex:a ex:p "0.1"^^xsd:float.`
  > actually represents / [D-entails](https://w3c.github.io/rdf-semantics/spec/#datatype_entailment_patterns)
  > `ex:a ex:p "0.100000001490116119384765625"^^xsd:float.`
  > Only `xsd:decimal` can exactly represent arbitrarry decimal numbers.

hartig:
@jmkeil thanks for the suggestion! I have created PR #44 with this change (using slightly different wording).

PR #44, opened 2023-05-23
Adds note about limitations of xsd:float and xsd:double

hartig:
... as suggested by @jmkeil in #43

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/44.html" title="Last updated on Jun 8, 2023, 4:18 PM UTC (9734f13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/44/d310888...9734f13.html" title="Last updated on Jun 8, 2023, 4:18 PM UTC (9734f13)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1156,15 +1156,15 @@
       of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
       <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
       include all decimal numbers. For every literal of
-      any of these two datatypes, the value
+      either of these two datatypes, the value
       of the literal is a value that can be represented as an
       <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
       binary floating point representation of the corresponding precision.
       For instance, the literal with
       lexical form `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
-      <a>denotes</a> the number 0.100000001490116119384765625.
-      As an alternative to <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
-      and <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
+      <a>denotes</a> the number `0.100000001490116119384765625`.
+      Rather than <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
+      or <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
       datatype <a data-cite="xmlschema11-2#decimal"><code>xsd:decimal</code></a>
       can be used to accurately capture arbitrary decimal numbers.</p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1161,9 +1161,9 @@
       <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
       binary floating point representation of the corresponding precision.
       As a consequence, not all decimal numbers are accurately captured by
-      using these two datatypes. For instance, the value of the literal with
-      lexical value `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
-      is rounded to the floating point number 0.100000001490116119384765625.
+      using these two datatypes. For instance, the literal with
+      lexical form `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
+      <a>denotes</a> the number 0.100000001490116119384765625.
       As an alternative to <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
       and <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
       datatype <a data-cite="xmlschema11-2#decimal"><code>xsd:decimal</code></a>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1084,8 +1084,8 @@
       <tr><td><a data-cite="xmlschema11-2#integer"><code>xsd:integer</code></a></td><td>Arbitrary-size integer numbers</td></tr>
 
       <tr><th rowspan="2">IEEE floating-point<br />numbers</th>
-          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
-      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
+          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
+      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
 
       <tr><th rowspan="4">Time and date</th>
           <td><a data-cite="xmlschema11-2#date"><code>xsd:date</code></a></td><td>Dates (yyyy-mm-dd) with or without timezone</td></tr>
@@ -1151,6 +1151,23 @@
         are sequence-valued datatypes which do not fit the RDF <a>datatype</a>
         model.</li>
     </ul>
+
+    <p class="note">As mentioned in the table above, the <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. The reason is that, for every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
+      of the literal is rounded to a value that can be represented as an
+      <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
+      binary floating point representation of the corresponding precision.
+      As a consequence, many decimal numbers are not accurately captured by
+      using these two datatypes. For instance, the value of the literal with
+      lexical value `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
+      is rounded to the floating point number 0.100000001490116119384765625.
+      As an alternative to <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
+      and <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
+      datatype <a data-cite="xmlschema11-2#decimal"><code>xsd:decimal</code></a>
+      can be used to accurately capture arbitrary decimal numbers.</p>
 
   </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1160,8 +1160,7 @@
       of the literal is a value that can be represented as an
       <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
       binary floating point representation of the corresponding precision.
-      As a consequence, not all decimal numbers are accurately captured by
-      using these two datatypes. For instance, the literal with
+      For instance, the literal with
       lexical form `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
       <a>denotes</a> the number 0.100000001490116119384765625.
       As an alternative to <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1084,8 +1084,8 @@
       <tr><td><a data-cite="xmlschema11-2#integer"><code>xsd:integer</code></a></td><td>Arbitrary-size integer numbers</td></tr>
 
       <tr><th rowspan="2">IEEE floating-point<br />numbers</th>
-          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
-      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
+          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
+      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
 
       <tr><th rowspan="4">Time and date</th>
           <td><a data-cite="xmlschema11-2#date"><code>xsd:date</code></a></td><td>Dates (yyyy-mm-dd) with or without timezone</td></tr>
@@ -1152,7 +1152,7 @@
         model.</li>
     </ul>
 
-    <p class="note">As mentioned in the table above, the <a>value spaces</a>
+    <p class="note">The <a>value spaces</a>
       of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
       <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
       include all decimal numbers. The reason is that, for every literal of
@@ -1160,7 +1160,7 @@
       of the literal is rounded to a value that can be represented as an
       <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
       binary floating point representation of the corresponding precision.
-      As a consequence, many decimal numbers are not accurately captured by
+      As a consequence, not all decimal numbers are accurately captured by
       using these two datatypes. For instance, the value of the literal with
       lexical value `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
       is rounded to the floating point number 0.100000001490116119384765625.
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">As mentioned in the table above, the <a>value spaces</a>
```
afs, line 1155: The definition of value space is given by https://www.w3.org/TR/xmlschema11-2/#sec-vs-double and https://www.w3.org/TR/xmlschema11-2/#sec-vs-float. 

It says "The double datatype is patterned after the IEEE double-precision 64-bit floating point datatype" so it is not claiming to be exactly IEEE 754-2008.

spec/index.html
```diff
@@ -1084,8 +1084,8 @@ <h3>The XML Schema Built-in Datatypes</h3>
       <tr><td><a data-cite="xmlschema11-2#integer"><code>xsd:integer</code></a></td><td>Arbitrary-size integer numbers</td></tr>
 
       <tr><th rowspan="2">IEEE floating-point<br />numbers</th>
-          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
-      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
+          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
```
afs, line 1087: I don't think text here is necessary because it already says "64-bit floating point numbers". Having the note below is enough.

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
 
+    <p class="note">As mentioned in the table above, the <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. The reason is that, for every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
+      of the literal is rounded to a value that can be represented as an
+      <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
+      binary floating point representation of the corresponding precision.
+      As a consequence, many decimal numbers are not accurately captured by
```
afs, line 1163:
```suggestion
      As a consequence, not all decimal numbers are accurately captured by
```

spec/index.html
afs, line 1155:
```suggestion
    <p class="note">The <a>value spaces</a>
```

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">As mentioned in the table above, the <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. The reason is that, for every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
+      of the literal is rounded to a value that can be represented as an
```
afs, line 1160:
```suggestion
      include all decimal numbers. The <a>lexical form</a> is mapped to a value that
      is rounded to a value represented by
```

spec/index.html
hartig: > I don't think text here is necessary because it already says "64-bit floating point numbers". Having the note below is enough.

Okay, I will remove these parentheses again.

spec/index.html
```diff
@@ -1084,8 +1084,8 @@ <h3>The XML Schema Built-in Datatypes</h3>
       <tr><td><a data-cite="xmlschema11-2#integer"><code>xsd:integer</code></a></td><td>Arbitrary-size integer numbers</td></tr>
 
       <tr><th rowspan="2">IEEE floating-point<br />numbers</th>
-          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
-      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
+          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
+      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN (note that this does <em>not</em> include all decimal numbers)</td></tr>
```
hartig, line 1088:
```suggestion
          <td><a data-cite="xmlschema11-2#double"><code>xsd:double</code></a></td><td>64-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
      <tr><td><a data-cite="xmlschema11-2#float"><code>xsd:float</code></a></td><td>32-bit floating point numbers incl. ±Inf, ±0, NaN</td></tr>
```

spec/index.html
hartig: Shortening the text as you suggest may introduce confusion. Readers may ask themselves: ' "The lexical form" of what? '

afs: Then how about just removing "The reason is that, "

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">The <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. The reason is that, for every literal of
```
hartig, line 1158: As suggested by @afs in https://github.com/w3c/rdf-concepts/pull/44#discussion_r1201874881
```suggestion
      include all decimal numbers. For every literal of
```

spec/index.html
hartig: > Then how about just removing "The reason is that, "

Done.

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. For every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
+      of the literal is rounded to a value that can be represented as an
+      <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
+      binary floating point representation of the corresponding precision.
+      As a consequence, not all decimal numbers are accurately captured by
+      using these two datatypes. For instance, the value of the literal with
+      lexical value `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
+      is rounded to the floating point number 0.100000001490116119384765625.
```
jmkeil, line 1166: This might be read as it is just a issue of the representation in memory after parsing (as one would type `0.1f` in Java). I would put more emphasis on the deviating semantic of the literale itself.
```suggestion
      using these two datatypes. For instance, the literal with lexical value
      `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
      denotes the number 0.100000001490116119384765625.
```

hartig:
> Is this just duplicating a discussion from XSD datatypes? The value space is defined by XSD, not RDF Concepts.

You are right. However, I agree with @jmkeil that the potential pitfalls of using `xsd:float`/`xsd:double` may be worth mentioning for RDF users.

afs:
> Is this just duplicating a discussion from XSD datatypes? The value space is defined by XSD, not RDF Concepts.

My concern as well. We have to find the balance between explanation as definition. RDF Concept is first and foremost definition. e.g. XMLSchema says "Floating point numbers are often used to approximate arbitrary real numbers." The proposal is different text for the same point.

The wording in XML Schema 1.1 is quite careful including to and from lexical representation.

Maybe something should go in the primer?

spec/index.html
gkellogg: The `<a href="#dfn-denote">` is actually incorrect; if it were external, you would use `data-cite` instead of `href`, however, as it's defined in RDF Concepts, it's sufficient to use `<a>denotes</a>`.

gkellogg (approved):
Approved, with a small comment on markup in an unmerged suggestion.

jmkeil:
> > Is this just duplicating a discussion from XSD datatypes? The value space is defined by XSD, not RDF Concepts.
> 
> My concern as well. We have to find the balance between explanation as definition. RDF Concept is first and foremost definition. e.g. XMLSchema says "Floating point numbers are often used to approximate arbitrary real numbers." The proposal is different text for the same point.
> 
> The wording in XML Schema 1.1 is quite careful including to and from lexical representation.
> 
> Maybe something should go in the primer?

The limitations of the value spaces defined in XSD have specific implications in context of RDF (denote, D-entailment). As RDF-Concepts recommends the use of `xsd:float` / `xsd:double`, it should also point to the implications of their use. In RDF-Primer, `xsd:float` / `xsd:double` are not even mentioned. It basically refers to RDF-Concepts. So the note would not be very helpful there.

Generally, there is less awareness of the value space limitations. `xsd:float` / `xsd:double` are often used without any specific requirements (e.g. need for `+INF`, `-INF`, `NaN`) that justify the limitations and the resulting distortion of represented values. Adding a note on that in RDF-Concepts might help to avoid these value distortions.

spec/index.html
afs: That value has more digits than an xsd:float represents - putting in "floating point" is confusing as whether it is xsd:float or abstract point on the number line or is the xsd:double for the xsd:float.

Suggestion remove "floating point" and leaving as "the number".

pfps:
I suppose that this is all mostly harmless but is there anyone who does not understand that floating point numbers do not include all rationals and that even some rationals that have short representations in decimal notation have to be approximated in any floating point number scheme that uses binary exponents.

I think it would be better to just point to some longer discussion of floating point numbers, such as https://www.w3.org/TR/xmlschema-2/#float or [Clinger 1990].

jmkeil:
> is there anyone who does not understand that floating point numbers do not include all rationals and that even some rationals that have short representations in decimal notation have to be approximated in any floating point number scheme that uses binary exponents.

Understanding is not the problem here, but awareness. People are used to use float/double for real numbers from programming. And using them in programming is reasonable in most cases. The lexical value in program code is the denoted value, and it is approximated by the binary floating point value in memory. Due to mechanisms in arithmetic and output (rounding), the actual meaning is preserved in most cases. In RDF however, the binary floating point value is by definition of XSD the denoted value and lexical representations, like `"0.1"`, are only an approximation of them. The implications of using float/double are different and people are not aware of that.

> I think it would be better to just point to some longer discussion of floating point numbers, such as https://www.w3.org/TR/xmlschema-2/#float or [Clinger 1990].

Sure, RDF-Concepts is not the right place for a long discussion. But pointing to https://www.w3.org/TR/xmlschema11-2/#float without any indication of the pitfall would not help much.

afs:
> The lexical representation in program code is the denoted value, and it is approximated by the binary floating point value in memory. ... In RDF however,

The lexical representation in program code has a datatype. The program text together with the datatype determines the value. XSD, RDF are no different. 

RDF Concepts says "Any IRI or literal denotes" (section 1.2).

The lexical representation in program code may have an expectation - that's down to the reader, not the definitions.

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">The <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. For every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
+      of the literal is rounded to a value that can be represented as an
```
afs, line 1160: If we're going to be precise, it's not "rounded". That implies the lexical form has a value in and of itself. The mapping is defined to yield the IEEE 754-2008 value within the limits of IEEE 754-2008.

The point is simply that 

"0.1"^^xsd:float != "0.1"^^xsd:decimal != "0.1"^^xsd:double

Application beware.

```suggestion
      of the literal is a value that can be represented as an
```

pfps:
I agree with what Andy said---program code, XML, and RDF are in the same boat.   One may write a floating point number as 0.1, but one has to remember in some cases that the actual floating point number is different than 0.1.  In other cases it doesn't matter - in program code, in XML, and in RDF.

For example:
```
:a :b "0.1"^^xsd:int .
```
matches
```
?x ?y "0.1"^^xsd:int.
```
and entails
```
_:x _:y "0.1"^^xsd:int .
```
even though the actual value is not exactly 1/10.

I seem to remember that there was a programming language that refused to do strict equality testing on floating point because of the pitfalls therein.

RDF itself is actually in a better shape than most programming languages because it does not have arithmetic operators.

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
+      <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
+      binary floating point representation of the corresponding precision.
+      As a consequence, not all decimal numbers are accurately captured by
+      using these two datatypes. For instance, the value of the literal with
```
jmkeil, line 1164: The sentence is redundant to the first sentence and the next sentence. To keep the note short, it could be skipped.

```suggestion
      
      For instance, the value of the literal with
```

spec/index.html
jmkeil: Putting the above suggestions together:

```suggestion
      using these two datatypes. For instance, the literal with
      lexical form `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
      <a>denotes</a> the number 0.100000001490116119384765625.
```

jmkeil:
> The lexical representation in program code has a datatype. The program text together with the datatype determines the value.

Fair point.

Nevertheless, xsd:double seems to be the first choice and xsd:decimal more the exception, despite the limitations of xsd:double. (See e.g. the [rdfs:range datatypes statistics out of >800 vocabularies](https://lov.linkeddata.es/dataset/lov/sparql?query=PREFIX+rdfs%3A%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0APREFIX+xsd%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23%3E%0A+%0ASELECT+%3Frange+(COUNT(*)+AS+%3Fcount)+%7B%0A++%5B%5D+rdfs%3Arange+%3Frange+.%0A++VALUES+%3Frange+%7Bxsd%3Adecimal+xsd%3Adouble+xsd%3Afloat%7D%0A%7D%0AGROUP+BY+%3Frange) [[not aggregated data](https://lov.linkeddata.es/dataset/lov/sparql?query=PREFIX+rdfs%3A%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0APREFIX+xsd%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23%3E%0A%0ASELECT+*+%7B%0A++%3Fproperty+rdfs%3Arange+%3Frange+.%0A++VALUES+%3Frange+%7Bxsd%3Adecimal+xsd%3Adouble+xsd%3Afloat%7D%0A%7D)]). I do not see a better explanaition for that, than limited awareness of the limitations.

spec/index.html
```diff
@@ -1152,6 +1152,23 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">The <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. For every literal of
+      any of these two datatypes, the value given as the <a>lexical form</a>
```
hartig, line 1159:
```suggestion
      any of these two datatypes, the value
```

afs:
Apologies not needed!

The auto-publish fails for other docs at the moment, may be all, as we move to Echidna publishing.

It does not have to stop us merging PRs. Things will sort themselves out. 

It'll help to merge so that new PRs will be relative to latest agreed state.

hartig:
Thanks for the clarification regarding the auto-publish failure.

I will wait with merging this PR until after today's telco.

spec/index.html
```diff
@@ -1152,6 +1152,22 @@ <h3>The XML Schema Built-in Datatypes</h3>
         model.</li>
     </ul>
 
+    <p class="note">The <a>value spaces</a>
+      of <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a> and
+      <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a> do not
+      include all decimal numbers. For every literal of
+      any of these two datatypes, the value
```
TallTed, line 1159:
```suggestion
      either of these two datatypes, the value
```

spec/index.html
```diff
@@ -1152,6 +1152,22 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
+      of the literal is a value that can be represented as an
+      <a href="http://ieeexplore.ieee.org/servlet/opac?punumber=4610933">IEEE 754-2008</a>
+      binary floating point representation of the corresponding precision.
+      For instance, the literal with
+      lexical form `"0.1"` and datatype <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>
+      <a>denotes</a> the number 0.100000001490116119384765625.
+      As an alternative to <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
+      and <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
```
TallTed, line 1167:
```suggestion
      Rather than <a data-cite="xmlschema11-2#double"><code>xsd:double</code></a>
      or <a data-cite="xmlschema11-2#float"><code>xsd:float</code></a>, the
```

PR #48, opened 2023-06-19
Add base direction as a fourth element of literals.

### LINKED ISSUE #9: base direction
A possible issue for RDF 1.2 is to standardize on a solution for the base direction of strings.

This would possibly include updating to the Abstract Syntax and associated changes to the various Concrete Syntax specifications.

See [RDF Literals and Base Directions](https://w3c.github.io/rdf-dir-literal/) for possible options.

JSON-LD introduced features for specifying the text direction. These included experimental features compatible with RDF 1.1:
[i18n namespace](https://www.w3.org/TR/json-ld11/#the-i18n-namespace), and [rdf:CompoundLiteral](https://www.w3.org/TR/json-ld11/#the-rdf-compoundliteral-class-and-the-rdf-language-and-rdf-direction-properties).

See the issue for the discussion of further options, and the [Working Group page](https://github.com/w3c/rdf-star-wg/blob/main/docs/text-direction.md) for further discussion.

gkellogg:
This addresses the conceptual representation of text direction in RDF language-tagged literals, based on discussion in:

* https://github.com/w3c/rdf-star-wg/blob/main/docs/text-direction.md, 
* [Internationalization Best Practices for Spec Developers](https://www.w3.org/TR/international-specs/#abstract),
* [Strings on the Web: Language and Direction Metadata](https://www.w3.org/TR/string-meta/), and
* [RDF Literals and Base Directions](https://w3c.github.io/rdf-dir-literal/).

Fixes #9.

For discussion:

* Separate `directional language-tagged string` with `rdf:dirLangString` datatype, or extend the notion of `language-tagged string` and continue to use `rdf:langString`.
* Bikeshedding terminology for literal element:
* [ ] `text direction` – From previous proposals
* [x] `base direction` – From [Unicode bidi basics](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics)
* [ ] `initial text direction` – From [Wiki](https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-considerations)
* [ ] `text directionality` – From HTML

(As the infrastructure seems to be having issues, you can also view the document via [GitHack](https://raw.githack.com/w3c/rdf-concepts/text-direction/spec/index.html).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/48.html" title="Last updated on Oct 13, 2023, 8:08 PM UTC (529b6dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/48/2d855bb...529b6dc.html" title="Last updated on Oct 13, 2023, 8:08 PM UTC (529b6dc)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -74,7 +74,7 @@
 
   <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
-    RDF 1.2 also introduces a <a>text direction</a> element to <a>language-tagged strings</a>,
+    RDF 1.2 also introduces a <a>base text direction</a> element to <a>directional language-tagged strings</a>,
     allowing the initial text direction to be specified when
     presented in a user agent.</p>
 
@@ -698,7 +698,7 @@
       case.
       The fourth element, <a>text direction</a> MUST NOT be
       present unless the <a>language tag</a> element is present.
-      The value space of <a>text direction</a> is either empty, or one of `ltr` or `rtl`.</p>
+      The value space of <a>base text direction</a> is either empty, or one of `ltr` or `rtl`.</p>
 
     <p>Please note that concrete syntaxes MAY support
       <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
@@ -717,8 +717,10 @@
 
     <ol>
       <li>If the literal is a <a>language-tagged string</a>,
-        then the literal value is a tuple consisting of its <a>lexical form</a>,
-        its <a>language tag</a>, and its <a>text direction</a>, in that order.</li>
+        then the literal value is a pair consisting of its <a>lexical form</a>
+        and its <a>language tag</a>, in that order.</li>
+      <li>if the literal is a <a>directional language string</a>, then the literal value is
+      a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base text direction</a>, likewise in that order.</li>
 
       <li>If the literal's <a>datatype IRI</a> is in the set of
         <a>recognized datatype IRIs</a>, let <var>d</var> be the
@@ -764,20 +766,20 @@
     <section id="section-text-direction" class="informative">
       <h3>Text Direction</h3>
 
-      <p>The <a>text direction</a> of a <a>language-tagged string</a>
-        provides means of establishing the initial direction of text,
+      <p>The <a>base text direction</a> of a <a>directional language string</a>
+        provides a means of establishing the initial direction of text,
         including text which is mixture of right-to-left and left-to-right scripts.
         The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
         a sequence of characters in logical order,
         so that they are visually ordered as expected.
         However, this is not sufficient to correctly render bidirectional text.</p>
 
-      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" may be displayed
-        incorrectly unless the <a>text direction</a> of `rtl` (for right-to-left)
+      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" (for Hebrew) may be displayed
+        incorrectly unless the <a>base text direction</a> of `rtl` (for right-to-left)
         is also provided.
         When this is provided to a user agent,
         for example using the HTML `dir` attribute,
-        this would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
+        it would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
     </section>
   </section>
 
@@ -1667,7 +1669,7 @@
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
       and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
-    <li>Added the <a>text direction</a> element as part of 
+    <li>Added the <a>base text direction</a> element as part of 
       a <a>literal</a>,
       and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
   </ul>
--- a/spec/index.html
+++ b/spec/index.html
@@ -757,9 +757,10 @@
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
       Two literals are term-equal (the same <a>RDF literal</a>)
       if and only if the two <a>lexical forms</a>,
-      the two <a>datatype IRIs</a> compare equal, character by character,
-      the two <a>language tags</a> (if any) compare equal, as if normalized to lower case, and
-      the two <a>base directions</a> (if any) compare equal, character by character.
+      the two <a>datatype IRIs</a>,
+      the two <a>language tags</a> (if any) normalized to lower case, and
+      the two <a>base directions</a> (if any),
+      all compare equal, character by character.
       Thus, two literals can have the same value
       without being the same <a>RDF term</a>.
       For example:</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -33,7 +33,7 @@
         { name: "Brian McBride" }
       ],
 
-      xref: ["i18n-glossary", "infra"],
+      xref: ["I18N-GLOSSARY", "INFRA"],
       github: "https://github.com/w3c/rdf-concepts/",
       group:           "rdf-star",
       doJsonLd:     true,
@@ -510,10 +510,10 @@
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
       or <a data-lt="string">RDF string</a>,
       is used to describe an ordered sequence of zero or more
-      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>
-      which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
+      <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a>
+      which are <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
       Unicode scalar values do not include the
-      <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
+      <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
       Note that most <a>concrete RDF syntaxes</a> require the use
       of the UTF-8 character encoding [[!RFC3629]], 
       and use the `\u0000` or `\U00000000` forms to express certain non-character values.
@@ -521,10 +521,10 @@
 
     <p>A string is identical to another string if it consists of the same sequence of code points.
       An implementation MAY determine string equality by comparing the
-      <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
-      that use the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a>
+      <a data-cite="I18N-GLOSSARY#dfn-code-unit">code units</a> of two strings
+      that use the same <a data-cite="I18N-GLOSSARY#dfn-character-encoding">Unicode character encoding</a>
       (UTF-8 or UTF-16) without decoding the string into a
-      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code point</a> sequence.</p>
+      <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code point</a> sequence.</p>
   </section>
 </section>
 
@@ -619,7 +619,7 @@
 
     <p><dfn>IRI equality</dfn>:
       Two IRIs are the same if and only if they consist of the same sequence of
-      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>,
+      <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a>,
       as in Simple String Comparison in
       <a data-cite="rfc3987#section-5.3.1">section 5.3.1</a> of [[!RFC3987]].
       (This is done in the abstract syntax, so the IRIs are resolved
@@ -705,10 +705,10 @@
 
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
-        <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>,
+        <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
+        which are <a data-cite="I18N-GLOSSARY#dfn-scalar-value">Unicode scalar values</a>,
         and therefore do not contain
-        <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">Unicode surrogate code points</a>.</li>
+        <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">Unicode surrogate code points</a>.</li>
       <li>a <dfn>datatype IRI</dfn>, being an <a>IRI</a>
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>, and</li>
@@ -788,7 +788,7 @@
       the two <a>datatype IRIs</a>,
       the two <a>language tags</a> (if any), and
       the two <a>base directions</a> (if any) compare equal,
-      using <a data-cite="i18n-glossary#dfn-case-sensitive">case sensitive matching</a>
+      using <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
       (see description of string comparison in <a href="#rdf-strings" class="sectionRef"></a>).
       Thus, two literals can have the same value
       without being the same <a>RDF term</a>.
@@ -1805,7 +1805,7 @@
       from <a data-cite="?JSON-LD11#the-rdf-json-datatype">Section&nbsp;10.2 The `rdf:JSON` Datatype</a>
       in [[?JSON-LD11]].</li>
     <li>Clarify Unicode terminology,
-      using <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>,
+      using <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a>,
       and restriction to the XML <a data-cite="XML11#charsets">Char</a> production.
       Also removes obsolete recommendations for the use of Normalization Form C in literals.
       Adds a definition of <a>string</a> that can be used in other RDF documents.</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -675,7 +675,7 @@
     <p>Literals are used for values such as strings, numbers, and dates.</p>
 
     <p>A <dfn data-local-lt="RDF literal">literal</dfn> in an <a>RDF graph</a> consists of
-      two to four elements:</p>
+      two, three, or four elements:</p>
 
     <ul>
       <li>a <dfn>lexical form</dfn>, being a Unicode [[!UNICODE]] string,
--- a/spec/index.html
+++ b/spec/index.html
@@ -74,7 +74,7 @@
 
   <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
-    RDF 1.2 also introduces a <a>base text direction</a> element to <a>language-tagged strings</a>,
+    RDF 1.2 also introduces a <a>base direction</a> element to <a>language-tagged strings</a>,
     allowing the initial text direction to be specified when
     presented in a user agent.</p>
 
@@ -689,16 +689,16 @@
         of [[!BCP47]].</li>
       <li>if and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>,
-        <dfn>base text direction</dfn> MUST be empty, `ltr` or `rtl`.</li>
+        <dfn>base direction</dfn> MUST be empty, `ltr` or `rtl`.</li>
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
       is present. Lexical representations of language tags MAY be converted
       to lower case. The value space of language tags is always in lower
       case.
-      The fourth element, <a>base text direction</a> MUST NOT be
+      The fourth element, <a>base direction</a> MUST NOT be
       present unless the <a>language tag</a> element is present.
-      The value space of <a>base text direction</a> is either empty, or one of `ltr` or `rtl`.</p>
+      The value space of <a>base direction</a> is either empty, or one of `ltr` or `rtl`.</p>
 
     <p>Please note that concrete syntaxes MAY support
       <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
@@ -720,7 +720,7 @@
         then the literal value is a pair consisting of its <a>lexical form</a>
         and its <a>language tag</a>, in that order.</li>
       <li>if the literal is a <a>language-tagged string</a>, then the literal value is
-      a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base text direction</a>, likewise in that order.</li>
+      a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>, likewise in that order.</li>
 
       <li>If the literal's <a>datatype IRI</a> is in the set of
         <a>recognized datatype IRIs</a>, let <var>d</var> be the
@@ -748,7 +748,7 @@
       if and only if the two <a>lexical forms</a>,
       the two <a>datatype IRIs</a>,
       the two <a>language tags</a> (if any), and
-      the two <a>base text directions</a> (if any)
+      the two <a>base directions</a> (if any)
       compare equal, character by character.
       Thus, two literals can have the same value
       without being the same <a>RDF term</a>. For example:</p>
@@ -764,9 +764,9 @@
       <a>lexical form</a> differs.</p>
 
     <section id="section-text-direction" class="informative">
-      <h3>Text Direction</h3>
+      <h3>Initial Text Direction</h3>
 
-      <p>The <a>base text direction</a> of a <a>language-tagged string</a>
+      <p>The <a>base direction</a> of a <a>language-tagged string</a>
         provides a means of establishing the initial direction of text,
         including text which is mixture of right-to-left and left-to-right scripts.
         The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
@@ -775,11 +775,17 @@
         However, this is not sufficient to correctly render bidirectional text.</p>
 
       <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" (for Hebrew) may be displayed
-        incorrectly unless the <a>base text direction</a> of `rtl` (for right-to-left)
+        incorrectly unless the <a>base direction</a> of `rtl` (for right-to-left)
         is also provided.
         When this is provided to a user agent,
         for example using the HTML `dir` attribute,
         it would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
+
+      <p class="note">The absence of a <a>base direction</a> does not necessarily imply that
+        the text has no initial text direct;
+        as described in [[[?UAX9]]],
+        strings may be embedded within structures which establish an <em>embedding direction</em>,
+        which determines the default bidirectional orientation of text.</p>
     </section>
   </section>
 
@@ -1516,17 +1522,14 @@
 
 <section id="internationalization" class="appendix informative">
   <h2>Internationalization Considerations</h2>
-  <p>RDF is restricted to representing string values with left-to-right or right-to-left direction indicators.
-    RDF provides a mechanism for specifying the language associated with
-    a string (<a>language-tagged string</a>),
-    but does not provide a means of indicating the base direction of the string.</p>
-
   <p>Unicode provides a mechanism for signaling direction within a string
-    (see [[[UAX9]]] [[UAX9]]),
-    however, when a string has an overall base direction which cannot be determined by the
+    (see [[[UAX9]]] [[UAX9]]).
+    RDF provides a machnism for specifying the <a>base direction</a>
+    of a <a>language-tagged string</a>
+    to signal the initial text direction of a string.
+    When a string has an overall base direction which cannot be determined by the
     beginning of the string, an external indicator is required,
-    such as the [[HTML]] <a data-cite="HTML/dom.html#the-dir-attribute">dir attribute</a>,
-    which currently has no counterpart for <a>RDF literals</a>.</p>
+    such as the [[HTML]] <a data-cite="HTML/dom.html#the-dir-attribute">dir attribute</a>.</p>
 
   <p>[[[JSON-LD11]]] [[JSON-LD11]] introduced the
     <a data-cite="JSON-LD11#the-i18n-namespace">i18n namespace</a> to use
@@ -1669,7 +1672,7 @@
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
       and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
-    <li>Added the <a>base text direction</a> element as part of 
+    <li>Added the <a>base direction</a> element as part of 
       a <a>literal</a>,
       and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
   </ul>
--- a/spec/index.html
+++ b/spec/index.html
@@ -692,9 +692,9 @@
       <li>if and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>,
         a non-empty <a>language tag</a>
-        MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
+        that MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]].
-        and <dfn>base direction</dfn> MUST be either `ltr` or `rtl`.</li>
+        and a <dfn>base direction</dfn> that MUST be either `ltr` or `rtl`.</li>
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
@@ -707,7 +707,7 @@
     <p>A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
       if both the third element and fourth elements are present.
       Language tags are treated identically to <a>language-tagged string</a>,
-      and the fourth element , <a>base direction</a> MUST BE either `ltr` or `rtl`.</p>
+      and the fourth element , <a>base direction</a> MUST BE either `ltr` or `rtl` in lower case.</p>
 
     <p>Please note that concrete syntaxes MAY support
       <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
@@ -721,7 +721,7 @@
       <a>language-tagged strings</a> and <a>directional language-tagged strings</a> without
       the <a>datatype IRI</a> because it always equals either
       <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>)
-      or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>).</p>
+      or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>), respectively.</p>
 
     <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is:</p>
 
@@ -772,7 +772,7 @@
     <p>denote the same <a data-lt="literal value">value</a>, but are not the
       same literal <a>RDF terms</a> and are not
       <a>term-equal</a> because their
-      <a>lexical form</a> differs.</p>
+      <a>lexical forms</a> differ.</p>
 
     <section id="section-text-direction" class="informative">
       <h3>Initial Text Direction</h3>
@@ -790,10 +790,10 @@
         is also provided.
         When this is provided to a user agent,
         for example using the HTML `dir` attribute,
-        it would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
+        it would be correctly presented as follows: <div lang="he" dir="rtl">פעילות הבינאום, W3C</div></p>
 
       <p class="note">The absence of a <a>base direction</a> does not necessarily imply that
-        the text has no initial text direct;
+        the text has no initial text direction;
         as described in [[[?UAX9]]],
         strings may be embedded within structures which establish an <em>embedding direction</em>,
         which determines the default bidirectional orientation of text.</p>
@@ -1535,7 +1535,7 @@
   <h2>Internationalization Considerations</h2>
   <p>Unicode provides a mechanism for signaling direction within a string
     (see [[[UAX9]]] [[UAX9]]).
-    RDF provides a machnism for specifying the <a>base direction</a>
+    RDF provides a mechanism for specifying the <a>base direction</a>
     of a <a>directional language-tagged string</a>
     to signal the initial text direction of a string.
     When a string has an overall base direction which cannot be determined by the
--- a/spec/index.html
+++ b/spec/index.html
@@ -74,9 +74,9 @@
 
   <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
-    RDF 1.2 also introduces a <a>base direction</a> element to <a>language-tagged strings</a>,
-    allowing the initial text direction to be specified when
-    presented in a user agent.</p>
+    RDF 1.2 also introduces <a>directional language-tagged strings</a>
+    containing a <a>base direction</a> element allowing the
+    initial text direction to be specified when presented in a user agent.</p>
 
   <p>RDF 1.2 Concepts introduces key concepts and terminology for RDF 1.2, discusses
     datatyping, and the handling of <a>fragment identifiers</a> in IRIs within
@@ -156,7 +156,9 @@
     <a>datatypes</a> that define the range of possible
     values, such as strings, numbers, and dates. Special kind of literals,
     <a>language-tagged strings</a>, denote
-    plain-text strings in a natural language.</p>
+    plain-text strings in a natural language, and
+    <a>directional language-tagged strings</a>, denote
+    plain-text strings in a natural language and include an initial text direction.</p>
 
     <p>Asserting an <a>RDF triple</a> says that <em>some relationship,
     indicated by the <a>predicate</a>, holds between the
@@ -688,17 +690,24 @@
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]].</li>
       <li>if and only if the <a>datatype IRI</a> is
-        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>,
-        <dfn>base direction</dfn> MUST be empty, `ltr` or `rtl`.</li>
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>,
+        a non-empty <a>language tag</a>
+        MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
+        of [[!BCP47]].
+        and <dfn>base direction</dfn> MUST be either `ltr` or `rtl`.</li>
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
-      is present. Lexical representations of language tags MAY be converted
-      to lower case. The value space of language tags is always in lower
-      case.
-      The fourth element, <a>base direction</a> MUST NOT be
-      present unless the <a>language tag</a> element is present.
-      The value space of <a>base direction</a> is either empty, or one of `ltr` or `rtl`.</p>
+      is present and the fourth element is not present.
+      Lexical representations of language tags MAY be converted
+      to lower case or formatted according to  the recommendation for formatting language tags
+      as described in <a data-cite="BCP47#section-2.1.1">section 2.1.1</a> of [[BCP47]].
+      The value of language tags is always treated as being in lower case.</p>
+
+    <p>A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
+      if both the third element and fourth elements are present.
+      Language tags are treated identically to <a>language-tagged string</a>,
+      and the fourth element , <a>base direction</a> MUST BE either `ltr` or `rtl`.</p>
 
     <p>Please note that concrete syntaxes MAY support
       <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
@@ -709,9 +718,10 @@
       <code>http://www.w3.org/2001/XMLSchema#string</code>
       (which is commonly abbreviated as <code>xsd:string</code>).
       Similarly, most concrete syntaxes represent
-      <a>language-tagged strings</a> without
-      the <a>datatype IRI</a> because it always equals
-      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>).</p>
+      <a>language-tagged strings</a> and <a>directional language-tagged strings</a> without
+      the <a>datatype IRI</a> because it always equals either
+      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>)
+      or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>).</p>
 
     <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is:</p>
 
@@ -719,8 +729,9 @@
       <li>If the literal is a <a>language-tagged string</a>,
         then the literal value is a pair consisting of its <a>lexical form</a>
         and its <a>language tag</a>, in that order.</li>
-      <li>if the literal is a <a>language-tagged string</a>, then the literal value is
-      a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>, likewise in that order.</li>
+      <li>if the literal is a <a>directional language-tagged string</a>, then the literal value is
+        a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
+        likewise in that order.</li>
 
       <li>If the literal's <a>datatype IRI</a> is in the set of
         <a>recognized datatype IRIs</a>, let <var>d</var> be the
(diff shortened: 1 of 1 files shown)


afs:
What about https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal?

iherman:
> See https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal
> 
> It uses "initial text direction" - does that capture your point @iherman?

Maybe, but it is not ideal. "Initial" suggests some sort of an ordering in time. Although I realize that I am on a slippery slope in terms of English terminology...

I am a bit worried by getting into unnecessary bike-shedding; the reason I proposed to pick the term used by the HTML spec is to avoid that... Part of the community has already picked that term, I am not sure if it is worth picking our own for something that is essentially the same.

spec/index.html
```diff
@@ -748,6 +760,26 @@ <h2>Literals</h2>
       same literal <a>RDF terms</a> and are not
       <a>term-equal</a> because their
       <a>lexical form</a> differs.</p>
+
+    <section id="section-text-direction" class="informative">
+      <h3>Text Direction</h3>
+
+      <p>The <a>text direction</a> of a <a>language-tagged string</a>
+        provides means of establishing the direction of text
+        written or mixed with right-to-left scripts.
```
afs, line 769:
```suggestion
        provides means of establishing the initial direction of text,
        including text which is mixture of right-to-left and left-to-right scripts.
```

spec/index.html
```diff
@@ -748,6 +760,26 @@ <h2>Literals</h2>
...
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the <a>language-tagged string</a> `"פעילות הבינאום, W3C"`
+        expressed with language tag `he` will be displayed incorrectly
+        unless the <a>text direction</a> of `rtl` (for right-to-left)
+        is also provided.
+        When this is provided to a user agent,
```
afs, line 779:
```suggestion
      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" may be displayed
      incorrectly unless the <a>text direction</a> of `rtl` (for right-to-left)
        is also provided.
```
or
```suggestion
      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" 
      will often be displayed incorrectly unless the <a>text direction</a>
      of `rtl` (for right-to-left) is also provided.
```

spec/index.html
```diff
@@ -1636,6 +1668,9 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
       and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
+    <li>Added the <a>text direction</a> element as part of
+      a <a>language-tagged string</a>,
+      and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
```
afs, line 1673:
```suggestion
    <li>Added the <a>text direction</a> element as part of 
      a <a>literal</a>,
      and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
```

afs:
> > It uses "initial text direction" - does that capture your point @iherman?
>
> Maybe, but it is not ideal. "Initial" suggests some sort of an ordering in time. Although I realize that I am on a slippery slope in terms of English terminology...

>the reason I proposed to pick the term used by the HTML spec is to avoid that... Part of the community has already picked that term, I am not sure if it is worth picking our own for something that is essentially the same.

Yes, it's tricky to get the right terminology.

When we talked to i18n, "text direction" was pointed out as suggesting "all this text" whereas this is meant as in "default".

In the "פעילות הבינאום, W3C" example there are multiple directions.

HTML also has `auto` which doesn't seem appropriate for RDF.

>I am a bit worried by getting into unnecessary bike-shedding; 

Yes :-)

afs:
> If there is going to be a partial solution provided in RDF, then I feel that it needs to be backward compatible with existing RDF systems. One way to do this is to use '-x-ltr' and '-x-rtl' at the end of the language tag. This produces valid language tags and is backwards compatible.

@pfps In https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal the proposed syntax is `@en--ltr`.  Not quite the same, but connected to the language tag.
```
LANGTAG ::=  "@" [a-zA-Z]+ ("-" [a-zA-Z0-9]+)*
```
RDF 1.2 gives us the opportunity for a syntax change while previous work has operated within the confines of RDF 1.1.

Being separate:
`LANG("foo"@en--ltr)` is proposed as "en", otherwise it would be `en-x-ltr`. Looking for "all literals with `en` continues to work. Similarly, the `LANGMATCHES` function continues to match (the algorithm is from RFC4647).

gkellogg:
@afs said:
> See https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal
> 
> It uses "initial text direction" - does that capture your point @iherman?

Sorry, there were a number of documents on (initial) text direction, and I based the PR off of https://github.com/w3c/rdf-star-wg/blob/main/docs/text-direction.md. We can reconcile the differences. I'm fine with "initial text direction", as that gets to the intent of the element. "Text Directionality" may have a subtly different meaning, as it describes the behavior of a display element, not a property of the text, but we can continue to discuss terminology either in this PR, or subsequently.

@pfps said:
> If text direction is to be supported in RDF let's support it in a general way. Proposals for adding an initial text direction to not meet this requirement as correct rendering of bidirectional text needs internal direction markers.

That's not my understanding of how bidirectional text works in Unicode. From [Unicode Bidirectional Algorithm basics](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics) each character already has its own directionality encoded, it's for cases where character classes are mixed that there is no a-priori way of knowing how to begin rendering the text. After setting off the initial direction, the Unicode algorithms handle any subsequent change in direction. Within that document **base direction** is used, so perhaps that would be a better term than "initial text direction" or "text directionality".

spec/index.html
```diff
@@ -73,7 +73,10 @@
   </ul>
 
   <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
-    which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.</p>
+    which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
+    RDF 1.2 also introduces a <a>text direction</a> element to <a>language-tagged strings</a>,
```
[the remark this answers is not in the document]
gkellogg: Yes, recall and see that now. Perhaps a topic for discussion, there is the danger of overloading concepts. Legacy language-tagged strings can be treated as having an empty (or `null`) (initial) text direction, so all language-tagged strings could be treated uniformly without necessarily introducing a new datatype. There are merits to both

Regarding getting the id of a definition or reference, we're obviously into ReSpec area here. As explored elsewhere, we can override the id assigned to a definition by giving it an `id` attribute, but people may make assumptions from the text of the definition. Once rendered by ReSpec, definitions will have an `id` attribute, and references will get an `href` based on that, or what is inferred from `data-lt`.

afs:
https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal is a write up based on issue #9.

spec/index.html
```diff
@@ -684,12 +687,18 @@ <h2>Literals</h2>
         language tag MUST be well-formed according to
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]].</li>
+      <li>if and only if the <a>datatype IRI</a> is
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>,
+        <dfn>text direction</dfn> MUST be empty, `ltr` or `rtl`.</li>
```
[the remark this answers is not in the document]
gkellogg: A literal is described as having two to four elements, and there are constraints on when it can have a language tag (or text direction). If there is an additional datatype `rdf:dirLangString` then we could say that language-tagged strings have three components and directional language-tagged-strings have four components, but at some additional cognitive load. If it is all `rdf:langString`, then there are four components, and text direction can have three possible values, empty (""), `ltr`, or `rtl`. SPARQL `LANGDIR` would return any of these three values for a language-tagged string.

spec/index.html
```diff
@@ -732,12 +741,15 @@ <h2>Literals</h2>
...
-      <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
-      without being the same RDF term. For example:</p>
+    <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
+      Two literals are term-equal (the same <a>RDF literal</a>)
+      if and only if the two <a>lexical forms</a>,
+      the two <a>datatype IRIs</a>,
+      the two <a>language tags</a> (if any), and
+      the two <a>text directions</a> (if any)
+      compare equal, character by character.
```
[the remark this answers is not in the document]
gkellogg: We previously say that the value space of `language tag` is lower case, is it redundant to say to use a case insensitive comparison here?

spec/index.html
gkellogg: I think you meant to replace the previous four lines, not the last. I'll create a substitution.

spec/index.html
```diff
@@ -748,6 +760,26 @@ <h2>Literals</h2>
...
+        written or mixed with right-to-left scripts.
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the <a>language-tagged string</a> `"פעילות הבינאום, W3C"`
+        expressed with language tag `he` will be displayed incorrectly
+        unless the <a>text direction</a> of `rtl` (for right-to-left)
+        is also provided.
```
gkellogg, line 778:
```suggestion
      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" may be displayed
        incorrectly unless the <a>text direction</a> of `rtl` (for right-to-left)
        is also provided.
```

gkellogg:
Applied some of @afs's suggestions, leaving the others for now pending further discussion.

iherman:
> Within that document **base direction** is used, so perhaps that would be a better term than "initial text direction" or "text directionality".

I am also fine with "base direction". Actually, when writing up my comment, my initial instinct was to propose that term but (to my surprise) that is not the term used by the HTML standard, and that is why I fell on the "directionality" side.  (No idea why that term was chosen for HTML.) Either way is fine with me.

pfps:
@gkellogg In https://www.w3.org/International/articles/inline-bidi-markup/ there are examples of strings that need embedded markup for correct rendering.  My takeaway is that a solution that only provides a language tag and a base direciton is insufficient.  The worst situation, I think, is including identifiers using strong ltr characters in rtl text, as in "[ARABIC TEXT] A7, B8, X" where the order of the identifiers is reversed from its correct order if there is no embedded markup.   Note that the language of this text is entirely Arabic - the identifiers are not English or any other language that uses ltr display.  I include an example with rtl identifiers inside ltr script.

Here are two identifiers using Hebrew script בבב, אא.  The first is בבב the second is אא

afs:
"base direction" works for me.

spec/index.html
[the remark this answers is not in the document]
TallTed: > is it redundant to say to use a case insensitive comparison here

I think it conforms to Postel's Law, will clearly reflect user intent, and will be better than case sensitivity-based errors.

TallTed:
I am also good with "base direction".

I don't like `-x-ltr` because it goes down the broken `-x` road. I am OK with `--ltr`.

pfps:
How is -x- broken?

spec/index.html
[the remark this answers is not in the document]
afs: I was jumping to conclusions. Some investigation: it isn't as simple as "case insensitive".

The text above this paragraph says "MAY lower case" the concrete string that is the language tag.

The text here is about term-equality and does not say it is a value-space comparison. (FWIW "value space" for language tags is a bit meaningless - "value spaces" involve datatypes but we are where we are.)

The RDF 1.1 text:

> the two language tags (if any) compare equal, character by character

does allow `"abc"@en` and `"abc"@EN` as different terms, whether that was intended or not. 

The root problem is that RDF has not used the canonical form for language tags. Some users do care about this.

At users' request Jena has options to leave as-is, always lower-case and always canonicalization.

Maybe better to leave as "character by character" because otherwise it is a implementation change.

spec/index.html
```diff
@@ -684,12 +687,18 @@ <h2>Literals</h2>
...
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
       is present. Lexical representations of language tags MAY be converted
       to lower case. The value space of language tags is always in lower
-      case.</p>
+      case.
+      The fourth element, <a>text direction</a> MUST NOT be
+      present unless the <a>language tag</a> element is present.
+      The value space of <a>text direction</a> is either empty, or one of `ltr` or `rtl`.</p>
```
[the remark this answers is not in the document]
TallTed:
```suggestion
      The value space of <a>base text direction</a> is either empty, or one of `ltr` or `rtl`.</p>
```

spec/index.html
```diff
@@ -708,8 +717,8 @@ <h2>Literals</h2>
 
     <ol>
       <li>If the literal is a <a>language-tagged string</a>,
-        then the literal value is a pair consisting of its <a>lexical form</a>
-        and its <a>language tag</a>, in that order.</li>
+        then the literal value is a tuple consisting of its <a>lexical form</a>,
+        its <a>language tag</a>, and its <a>text direction</a>, in that order.</li>
```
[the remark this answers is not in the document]
TallTed:
```suggestion
        then the literal value is a pair consisting of its <a>lexical form</a>
        and its <a>language tag</a>, in that order.</li>
      <li>if the literal is a <a>directional language string</a>, then the literal value is
      a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base text direction</a>, likewise in that order.</li>
```

spec/index.html
```diff
@@ -748,6 +760,26 @@ <h2>Literals</h2>
       same literal <a>RDF terms</a> and are not
       <a>term-equal</a> because their
       <a>lexical form</a> differs.</p>
+
+    <section id="section-text-direction" class="informative">
+      <h3>Text Direction</h3>
+
+      <p>The <a>text direction</a> of a <a>language-tagged string</a>
```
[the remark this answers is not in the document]
TallTed:
```suggestion
      <p>The <a>base text direction</a> of a <a>directional language string</a>
```

spec/index.html
```diff
@@ -748,6 +760,25 @@ <h2>Literals</h2>
...
+        provides means of establishing the initial direction of text,
```
TallTed, line 768:
```suggestion
        provides a means of establishing the initial direction of text,
```

spec/index.html
```diff
@@ -748,6 +760,25 @@ <h2>Literals</h2>
...
+        including text which is mixture of right-to-left and left-to-right scripts.
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" may be displayed
+        incorrectly unless the <a>text direction</a> of `rtl` (for right-to-left)
```
TallTed, line 776:
```suggestion
        incorrectly unless the <a>base text direction</a> of `rtl` (for right-to-left)
```

spec/index.html
```diff
@@ -748,6 +760,25 @@ <h2>Literals</h2>
...
+
+      <p>The <a>text direction</a> of a <a>language-tagged string</a>
+        provides means of establishing the initial direction of text,
+        including text which is mixture of right-to-left and left-to-right scripts.
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" may be displayed
```
TallTed, line 775:
```suggestion
      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" (for Hebrew) may be displayed
```

spec/index.html
```diff
@@ -748,6 +760,25 @@ <h2>Literals</h2>
...
+        incorrectly unless the <a>text direction</a> of `rtl` (for right-to-left)
+        is also provided.
+        When this is provided to a user agent,
+        for example using the HTML `dir` attribute,
+        this would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
```
TallTed, line 780:
```suggestion
        it would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
```

spec/index.html
```diff
@@ -1636,6 +1667,9 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
       and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
+    <li>Added the <a>text direction</a> element as part of
```
TallTed, line 1670:
```suggestion
    <li>Added the <a>base text direction</a> element as part of 
```

TallTed (changes requested):

spec/index.html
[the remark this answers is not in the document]
afs: The cognitive load is introducing a new concept.

`rdf:langString` is currently three slots.

There isn't a datatype class of all strings with a base direction if it is a new facet of `rdf:langString`.

SPARQL, or any toolkit API, can present this in a context suitable way but let's make the core data model exact.

spec/index.html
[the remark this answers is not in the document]
TallTed: Hm, yes, I see what you mean. Accidental inconsistency is a rarely vanquished hobgoblin, especially across specs developed separately over years. We have our work cut out for us, in trying to bring consistency to all these docs that we're simultaneously trying to upgrade/update.

"Character by character" is at least clear, and we can include a note that advises deployers of the potential need to enact Jena's options — i.e., keep original langtag casing, make all langtags lower-case, or (whatever you meant by "always canonicalization").

TallTed:
> How is -x- broken?

See [RFC#6648: Deprecating the "X-" Prefix and Similar Constructs in Application Protocols](https://datatracker.ietf.org/doc/html/rfc6648)

pfps:
I think we should get approval from the i18n group before adding any incomplete solutions to text directions - and this PR is, so far, incomplete.

afs:
There is an extensions mechanism which uses singleton subtags, which is not private tags.
https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.6.
I don't know if there is much deployed use of these.

They have to be registered. https://www.rfc-editor.org/rfc/rfc5646.html#section-3.7

I don't think singletons will work.

pfps:
There is one such subtag - for character ordering.   See https://www.w3.org/International/questions/qa-choosing-language-tags#extension

Getting a subtag for base direction would be a good idea, but might be a lot of work and might not be approved.   The subtag would be 'd' and the possible values 'auto', 'ltr', and 'rtl', I guess.

spec/index.html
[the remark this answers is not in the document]
afs: I don't believe it is transparent as implied by "less cognitive burden".

Simply ignoring direction changes cardinality but RDF graphs are sets (the usual effect of projection).

Used with `skos:prefLabel` there is a pattern of one label per language tag:
https://www.w3.org/TR/skos-reference/#S14

Finding literals with datatype `rdf:langString` may now find multiple instances if searching on `LANGMATCHES` or `DATATYPE`.

spec/index.html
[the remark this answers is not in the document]
gkellogg: We need something to reconcile the notions of a lower-case value (space) for the language-tag and the fact that it's compared character by character (code point by code point?). Are there systems where `"abc"@en` and `"abc"@EN` are not considered the same term.

It might say something like "two language tags (if any) compare equal after normalizing to lower case".

The sentence "The value space of language tags is always in lower case" might be changed to "The value of language tags is always treated as being in lower case".

gkellogg:
> There is one such subtag - for character ordering. See https://www.w3.org/International/questions/qa-choosing-language-tags#extension
> 
> Getting a subtag for base direction would be a good idea, but might be a lot of work and might not be approved. The subtag would be 'd' and the possible values 'auto', 'ltr', and 'rtl', I guess.

This was considered in [RDF Literals and Base Directions](https://w3c.github.io/rdf-dir-literal) in [Extend language tags with `-d-*`](https://w3c.github.io/rdf-dir-literal/#extending-bcp).

The pros and cons section noted some barriers to adoption and negative reaction in an [email thread](https://mailarchive.ietf.org/arch/browse/ietf-languages/?gbt=1&index=FNDxsLMAPdb8I1Bw6iyGlWZvuGM).

spec/index.html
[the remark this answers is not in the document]
afs: We should avoid "value" because that is about datatype literals.

"compare equal as if normalized to lower case".

which does not imply they are converted (the earlier text is MAY).

> Are there systems where "abc"@en and "abc"@EN
Jena keeps them apart but they are the same term (yes - that's contradictory)

It's about meeting the user expectation of round-trip with no change.

spec/index.html
```diff
@@ -684,12 +687,18 @@ <h2>Literals</h2>
...
+      <li>if and only if the <a>datatype IRI</a> is
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>,
+        <dfn>base direction</dfn> MUST be empty, `ltr` or `rtl`.</li>
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
       is present. Lexical representations of language tags MAY be converted
       to lower case. The value space of language tags is always in lower
-      case.</p>
+      case.
```
gkellogg, line 698: This also suggests standard formatting in BCP47 and update the sentence about language-tag values.

```suggestion
      to lower case,
      or formatted according to  the recommendation for formatting language tags
      as described in <a data-cite="BCP47#section-2.1.1">section 2.1.1</a> of [[BCP47]].
      The value of language tags is always treated as being in lower case
```

spec/index.html
```diff
@@ -732,12 +743,15 @@ <h2>Literals</h2>
...
-      <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
-      without being the same RDF term. For example:</p>
+    <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
+      Two literals are term-equal (the same <a>RDF literal</a>)
+      if and only if the two <a>lexical forms</a>,
+      the two <a>datatype IRIs</a>,
+      the two <a>language tags</a> (if any), and
+      the two <a>base directions</a> (if any)
+      compare equal, character by character.
```
gkellogg, line 752: This is redundant for base direction, but it's easier to apply the comparison criteria for both language tags and base directions.

```suggestion
      compare equal, as if normalized to lower case.
```

gkellogg:
Without further opinions expressed in the next couple of days, I'll apply @afs's suggestions for using two different datatypes (`rdf:langString` and `rdf:dirLangString`) to define `directional language-tagged string` in addition to `language-tagged string`.

gkellogg:
Updated based on @afs's suggested of defining "directional language-tagged string".

spec/index.html
```diff
@@ -684,12 +689,25 @@ <h2>Literals</h2>
...
+        MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
+        of [[!BCP47]].
+        and <dfn>base direction</dfn> MUST be either `ltr` or `rtl`.</li>
     </ul>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
-      is present. Lexical representations of language tags MAY be converted
-      to lower case. The value space of language tags is always in lower
-      case.</p>
+      is present and the fourth element is not present.
```
Tpt, line 701: I am a bit not at ease with this definition. Intuitively directional language-tagged strings are a special case of language-tagged string with an extra piece of information and this definition prevents to state that `rdf:dirLangString` is a sub-datatype of `rdf:langString`. It's why I would be personally more in favor of just adding an optional base direction to `rdf:langString`.

afs: We adding a new concept and it's about human language so it's not surprising there isn't a single, natural choice of design.

RDF itself doesn't have subclass/superclass. 

In XMLScheme there is a more complex system for derived datatypes - by extension and by restriction. `xsd:integer` is a restriction of `xsd:decimal`. A subdatatype for `rdf:dirLangString` is "by extension" adding a new facet which would be new to RDF core. Equally it could be a superdatatype with `rdf:langString` derived-by-restriction.

sub/super impacts the choice of `DATATYPE("abc"@en--rtl)` and `DATATYPE(?x) = rdf:langString`.

[Proposal for SPARQL functions](https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal#functions)

The choices for SPARQL functions don't give a connection.
SKOS labels seem to be the most sensitive to the choice in RDF concepts and is also the sensitive to the SPARQL choices.

It seems to come down to a choice of no new datatype or a separate one.

The idea that there isn't a datatype/class for literals with base direction doesn't feel like RDF to me.
A toolkit can of course present the RDF in a convenient way but if the information isn't available (one datatype) it can't.

afs:
> Without further opinions expressed in the next couple of days, I'll apply @afs's suggestions for using two different datatypes (`rdf:langString` and `rdf:dirLangString`) to define `directional language-tagged string` in addition to `language-tagged string`.

Minor.
As long a it's clear we aren't changing the definition of language tags - that's not within our scope.

aphillips:
There's a lot going on in this thread. I'll add a few comments from the I18N and BCP47 perspective... and add this to I18N WG's agenda to discuss.

Regarding the term to use for the direction of a field, the I18N Glossary defines two terms: "base direction" and "paragraph direction". Either is acceptable. "Initial text direction" (from HTML) is also acceptable. The reason we prefer "base" or "paragraph" is that "initial" could imply that it only applies to the start of the string, whereas the value is actually the paragraph direction for the whole string.

Regarding matching language tags, we shouldn't assume that tags will always be canonicalized to lowercase but comparison "character-by-character" (it is "code point by code point", but the tags are limited to a subset of ASCII, for which there is no ambiguity) will be unreliable unless tags are canonicalized (BCP47 defines a mixed-case canonicalization) or ASCII lowercased--or compared case-insensitively, which might be more difficult in the RDF space.

There are two extensions to BCP47 in production: the `-t-` extension for transliteration/transcription and the `-u-` extension for (Unicode) locales. The locale extension is used by JavaScript and in several libraries and programming languages and is used to modify things such as sorting preferences, number presentation, and the like. Note that canonicalized language tags order the extensions ASCII-betically, i.e. the `-d-` extension should appear before the `-t-` or `-u-` extensions, whereas we probably want `-d` on the end so that the value can be truncated away. Also, if a `-d-` extension were created, it wouldn't be possible to tell if RDF processing had added the value or if the value were part of the language tag's data.

Adding an extension for base direction has been mooted before, but I18N and others don't necessarily think this is the best solution (I think @gkellogg has a link above to the discussion). Base direction really is separate from the language of content. It might be that such an extension turns out to be the most convenient way to solve the base direction problem, but it feels "hacky" (trying to smuggle additional data through an existing channel, if you will). Applications would have to process language tags (including decorating the language tag with the direction and removing it from the language tag--without disturbing other extensions--when deserializing content). It also adds to the length of the tag, which might run up against truncation requirements for some limited implementations.

Using a custom separator such as `--` could work. Note that JSON-LD has a serialization [here](https://www.w3.org/TR/json-ld/#the-i18n-namespace), which uses `_` (illegal in language tags and thus unambigous).

afs:
@aphillips - thank you for the comments especially about the `-d-` singleton and canonical order.

gkellogg:
@aphillips said:

> Regarding the term to use for the direction of a field, the I18N Glossary defines two terms: "base direction" and "paragraph direction". Either is acceptable. "Initial text direction" (from HTML) is also acceptable. The reason we prefer "base" or "paragraph" is that "initial" could imply that it only applies to the start of the string, whereas the value is actually the paragraph direction for the whole string.

We settled on **base direction**, which is in the PR now.

> Regarding matching language tags, we shouldn't assume that tags will always be canonicalized to lowercase but comparison "character-by-character" (it is "code point by code point", but the tags are limited to a subset of ASCII, for which there is no ambiguity) will be unreliable unless tags are canonicalized (BCP47 defines a mixed-case canonicalization) or ASCII lowercased--or compared case-insensitively, which might be more difficult in the RDF space.

The description of matching language tags now says "... the two [language tags](http://localhost:8000/rdf-concepts/spec/#dfn-language-tag) (if any) compare equal, as if normalized to lower case ...". The use of "character" rather than "code point" is fairly pervasive, and should be addressed in a separate PR.

Note that a previous statement "language tags MAY be converted to lower case" has an additional clause "or formatted according to the recommendation for formatting language tags as described in [section 2.1.1](https://www.rfc-editor.org/rfc/rfc5646#section-2.1.1) of [[BCP47](http://localhost:8000/rdf-concepts/spec/#bib-bcp47)]". This is, in fact, done by many implementations, but converting to lower case is also common. Note that these are non-normative statements. Perhaps we could put more emphasis on converting to BCP47 formatting.

Additionally, @pfps has questioned (https://github.com/w3c/rdf-concepts/pull/48#issuecomment-1602980228) if adding base directions to strings is sufficient to meet the needs of I18N. It would be good to get a specific response to this from the I18N group to see if the solution offered here is adequate or could be further improved.

Thanks for the comments!

gkellogg:
> > Without further opinions expressed in the next couple of days, I'll apply @afs's suggestions for using two different datatypes (`rdf:langString` and `rdf:dirLangString`) to define `directional language-tagged string` in addition to `language-tagged string`.
> 
> Minor. As long a it's clear we aren't changing the definition of language tags - that's not within our scope.

The only "change" is an additional informative suggestion to format per BCP47 recommendations. Normalizing to lower case probably isn't going away. The choice of formatting is separate from the text for comparison, which now considers the strings as converted to lower case for comparison purposes.

pfps:
I read over https://www.w3.org/TR/international-specs/ and as far as I can see there are lots of best practices there that are not currently met in RDF (such as https://www.w3.org/TR/international-specs/#lang_inline).   is there going to be an analysis to determine how to satisfy these other best practices?

gkellogg:
@aphillips in https://github.com/w3c/rdf-concepts/pull/48#issuecomment-1609678186 you said you were adding this to the I18N group agenda for discussion. Do you have a target for when we can expect feedback?

The proposed section 3.3.1 is titled "Initial Text Direction", and the text says it provides a means of establishing the initial direction of text; from your comments. I'm not sure "paragraph direction" works in the context of a directional language-tagged string, but there may be some room for further improvement in this description.

spec/index.html
Tpt: Sorry, I missed your reply. Sorry for that.

> It seems to come down to a choice of no new datatype or a separate one.

Yes, I agree that having `rdf:dirLangString` is useful and we should have it.

> The choices for SPARQL functions don't give a connection.

I am not sure to fully understand what you are saying. Sorry for that. Isn't it already the case with XSD numeric datatypes? They extend one an other but `DATATYPE` only returns the most specific datatype (i.e. `DATATYPE("1"^^xsd:int) = xsd:int` and not `DATATYPE("1"^^xsd:int) = xsd:integer`)

> is "by extension" adding a new facet which would be new to RDF core.

Yes, that's true. However if we do not make `rdf:dirLangString` some kind of "subdatatype" of `rdf:langString` then it means for example that all vocabularies having `rdfs:range rdf:langString` would have to be updated to state that the range is the union of `rdf:langString` and `rdf:dirLangString`. This seems quite cumbersome to me.
I would tend to think that `"foo"^^en--ltr` entailing `"foo"^^en` is something we might want to have in some entailment regimes.
But it's probably a topic for  RDF-semantics or RDF-schema and not for RDF-concepts. So, I am convinced by the current PR. Thank you for your reply!

spec/index.html
```diff
@@ -732,12 +754,15 @@ <h2>Literals</h2>
...
-      RDF literal) if and only if the two <a>lexical forms</a>,
-      the two <a>datatype IRIs</a>, and the two
-      <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
-      without being the same RDF term. For example:</p>
+    <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
+      Two literals are term-equal (the same <a>RDF literal</a>)
+      if and only if the two <a>lexical forms</a>,
+      the two <a>datatype IRIs</a> compare equal, character by character,
+      the two <a>language tags</a> (if any) compare equal, as if normalized to lower case, and
```
pchampin, line 761:
```suggestion
      the two <a>language tags</a> (if any) compare equal after being normalized to lower case, and
```

spec/index.html
```diff
@@ -748,6 +773,31 @@ <h2>Literals</h2>
       same literal <a>RDF terms</a> and are not
       <a>term-equal</a> because their
       <a>lexical form</a> differs.</p>
```
pchampin, line 775:
```suggestion
      <a>lexical forms</a> differ.</p>
```

spec/index.html
```diff
@@ -732,12 +754,15 @@ <h2>Literals</h2>
...
-      the two <a>datatype IRIs</a>, and the two
-      <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
-      without being the same RDF term. For example:</p>
+    <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
+      Two literals are term-equal (the same <a>RDF literal</a>)
+      if and only if the two <a>lexical forms</a>,
+      the two <a>datatype IRIs</a> compare equal, character by character,
+      the two <a>language tags</a> (if any) compare equal, as if normalized to lower case, and
+      the two <a>base directions</a> (if any) compare equal, character by character.
```
pchampin, line 762: the mention of "character to character" is odd, because currently, it seems to apply only to datatype IRI and direction, and not to lexical form nor language tag...

I propose the current change (which overrides my proposed change above for line 761)
```suggestion
      if and only if the two <a>lexical forms</a>,
      the two <a>datatype IRIs</a>,
      the two <a>language tags</a> (if any) normalized to lower case, and
      the two <a>base directions</a> (if any), compare equal, character by character.
```

It might be argued to be ambiguous (must they *all* be equal with each other?!) but that ambiguity existed in the text from RDF 1.1, and didn't seem to bother anyone :stuck_out_tongue:

spec/index.html
```diff
@@ -748,6 +773,31 @@ <h2>Literals</h2>
...
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" (for Hebrew) may be displayed
+        incorrectly unless the <a>base direction</a> of `rtl` (for right-to-left)
+        is also provided.
+        When this is provided to a user agent,
+        for example using the HTML `dir` attribute,
+        it would be correctly presented as <br/><span lang="he" dir="rtl">פעילות הבינאום, W3C</span>.</p>
```
pchampin, line 793:
```suggestion
        it would be correctly presented as follows: <div lang="he" dir="rtl">פעילות הבינאום, W3C</div></p>
```
with a `<div>`, the text is actually justified to the right, which makes it more obvious that it is RTL (and probably more correct).

spec/index.html
```diff
@@ -748,6 +773,31 @@ <h2>Literals</h2>
...
+
+      <p class="note">The absence of a <a>base direction</a> does not necessarily imply that
+        the text has no initial text direct;
```
pchampin, line 796:
```suggestion
        the text has no initial text direction;
```
??

spec/index.html
```diff
@@ -1483,17 +1533,14 @@ <h2>Security Considerations</h2>
...
-  <p>RDF is restricted to representing string values with left-to-right or right-to-left direction indicators.
-    RDF provides a mechanism for specifying the language associated with
-    a string (<a>language-tagged string</a>),
-    but does not provide a means of indicating the base direction of the string.</p>
-
   <p>Unicode provides a mechanism for signaling direction within a string
-    (see [[[UAX9]]] [[UAX9]]),
-    however, when a string has an overall base direction which cannot be determined by the
+    (see [[[UAX9]]] [[UAX9]]).
+    RDF provides a machnism for specifying the <a>base direction</a>
```
pchampin, line 1538:
```suggestion
    RDF provides a mechanism for specifying the <a>base direction</a>
```

pchampin (approved):
I made a number of suggestion to improve this PR.

I share @Tpt's unease about having two datatypes `rdf:langString` and `rdf:dirLangString`, but I believe it would be better to merge this PR first, and then use it as a starting point for further discussions.

spec/index.html
```diff
@@ -684,12 +689,25 @@ <h2>Literals</h2>
...
+      is present and the fourth element is not present.
+      Lexical representations of language tags MAY be converted
+      to lower case or formatted according to  the recommendation for formatting language tags
+      as described in <a data-cite="BCP47#section-2.1.1">section 2.1.1</a> of [[BCP47]].
+      The value of language tags is always treated as being in lower case.</p>
+
+    <p>A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
+      if both the third element and fourth elements are present.
+      Language tags are treated identically to <a>language-tagged string</a>,
+      and the fourth element , <a>base direction</a> MUST BE either `ltr` or `rtl`.</p>
```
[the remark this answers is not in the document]
gkellogg: If we were to settle on just using `rdf:langString` for both, then the bullet point wouldn't be necessary, but there seem to be reasons that it is necessary. Can be discussed further after merge.

spec/index.html
gkellogg: The spec does not requiring normalizing to lower case, only for comparison purpose. Your suggested change would seem to suggest that they are normalized to lower case. It's only the comparison which is case insensitive.

spec/index.html
gkellogg: IIRc, I had something similar to this originally, but after feedback from @afs, spelled it out for each element.

TallTed:
```suggestion
      if and only if the two <a>lexical forms</a>,
      the two <a>datatype IRIs</a>,
      the two <a>language tags</a> (if any) normalized to lower case, and
      the two <a>base directions</a> (if any),
      all compare equal, character by character.
```

TallTed: or
```suggestion
      if and only if all of
      the two <a>lexical forms</a>,
      the two <a>datatype IRIs</a>,
      the two <a>language tags</a> (if any) normalized to lower case, and
      the two <a>base directions</a> (if any),
      compare equal, character by character.
```

gkellogg: There should probably be a colon after "if all of", as in "if and only  if all of:", as it's enumerating a set of clauses. I'm otherwise fine with this suggestion, but just be be clear, the language tags are normalized to lower case only for comparison purposes and the language tag elements, themselves, are not considered to necessarily be represented in lower case.

afs (approved):

afs (commented):
There is a substantive change in term-equality for language tags within this PR.

Original text:
"Thus, two literals can have the same value without being the same RDF term. "
New text:
"language tags ... normalized to lower case, ... compare equal, character by character."

pfps:
The change to term equality should probably be split out into a separate PR.

afs:
Yes - let's split it out and get the base direction text into the document.

TallTed:
Is there significant controversy about the term-equality language in this PR (which hasn't been noted here yet, but could be)? If so, then yes, it should probably be split out... but lacking such controversy, I don't see the need for a split. Changes can still be made via another PR...

spec/index.html
```diff
@@ -732,12 +754,16 @@ <h2>Literals</h2>
...
-      without being the same RDF term. For example:</p>
+    <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
+      Two literals are term-equal (the same <a>RDF literal</a>)
+      if and only if the two <a>lexical forms</a>,
+      the two <a>datatype IRIs</a>,
+      the two <a>language tags</a> (if any) normalized to lower case, and
+      the two <a>base directions</a> (if any),
+      all compare equal, character by character.
+      Thus, two literals can have the same value
+      without being the same <a>RDF term</a>.
```
TallTed, line 765: This is the text containing the change in term-equality, cited by @afs

gkellogg: There's an additional bit about potentially normalizing language tags based on BCP47 rules that can be moved to the new PR as well.

gkellogg: See https://github.com/w3c/rdf-concepts/pull/48/commits/c45d9470785a9c681c1ad2cf8cf47906b4d62c75.

gkellogg:
> There is a substantive change in term-equality for language tags within this PR.
> 
> Original text: "Thus, two literals can have the same value without being the same RDF term. " New text: "language tags ... normalized to lower case, ... compare equal, character by character."

See https://github.com/w3c/rdf-concepts/pull/48/commits/c45d9470785a9c681c1ad2cf8cf47906b4d62c75, which removes  the comparison after normalization.

TallTed (approved):
Looks right.

afs (approved):

spec/index.html
```diff
@@ -1694,6 +1694,8 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
     <li>Added the <a>base direction</a> element as part of 
       a <a>literal</a>,
       and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
+    <li>Changed reference from DOM4, which was not a recommendation at the time, to [[DOM]],
+      making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
```
afs, line 1698: "normative" - that has two readings (1) the definition is normative as possible recognized dadatypes (2) the spec makes them normative/required.

```suggestion
      making the definitions of <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
```
(given other discussions, we need to clear that the RDF Data Model does not mandate implementation and all the consequence of these or any other datatypes)

gkellogg: @afs, this slipped before subsequent update; I've added the change you suggested, but any other discussion about not mandating implementation of these datatypes probably needs to go in another issue/PR.

afs: Thank you for adding the change.

RDF 1.1 did not mandate implementation of any datatypes (except, arguably, `rdf:langString`).

gkellogg:
I was going to include it in the first slot proposed for I18N, along with the Unicode cleanup (if necessary) and discussion of BCP47 case sensitivity when it comes to literal equality, and thus if triples differing in case are the same, or not. I'm separately working on slides for this section.

spec/index.html
```diff
@@ -748,7 +773,32 @@ <h2>Literals</h2>
...
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+      <p>For example, the text `"פעילות הבינאום, W3C"` with language tag "he" (for Hebrew) may be displayed
+        incorrectly unless the <a>base direction</a> of `rtl` (for right-to-left)
+        is also provided.
+        When this is provided to a user agent,
+        for example using the HTML `dir` attribute,
+        it would be correctly presented as follows: <div lang="he" dir="rtl">פעילות הבינאום, W3C</div></p>
```
aphillips, line 794: Do you intend for this to display with a base direction of LTR here so that the later tagged example rendering differently is obvious? Or did you omit the <span dir=rtl> by not observing it?

Note that our specs are translated, so if your intention is the make this show the LTR display for contrast, you might want to do <bdi dir=ltr translate=no> around it.

If you want to show logical order (which won't make sense) would require an `lro` markup, but would have to be explained (by using the code points).

The simplest is probably to say something like:

```suggestion
<p>For example, some text with a language tag "he" might be displayed in a left-to-right context (such as an English web page) as <bdi dir=ltr>פעילות הבינאום, W3C</bdi>, which is incorrect. When provided to a user agent including base direction information (such as using HTML's `dir` attribute) it can then be correctly presented as:</p>
<div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
```

spec/index.html
```diff
@@ -748,7 +773,32 @@ <h2>Literals</h2>
...
+
+      <p class="note">The absence of a <a>base direction</a> does not necessarily imply that
+        the text has no initial text direction;
+        as described in [[[?UAX9]]],
+        strings may be embedded within structures which establish an <em>embedding direction</em>,
+        which determines the default bidirectional orientation of text.</p>
```
aphillips, line 800: This is slightly misleading. The bidi algorithm determines the base direction in any case. And "embedding" is an overloaded term in the bidi algorithm (strings can be "embedded", but "embedding" in bidi refers to stacking bidirectional states...) 

I'm not sure what the note is trying to convey. Are you trying to say "if the direction is not provided as metadata, the string can still be rendered"? Generally, what we say is either (a) when there is no base direction provided for a given string, the `auto` (first-strong detection) direction should be used; or (b) when the base direction is not provided, the direction of the enclosing document (or content??) is used

spec/index.html
```diff
@@ -1492,17 +1542,14 @@ <h2>Security Considerations</h2>
...
-    however, when a string has an overall base direction which cannot be determined by the
+    (see [[[UAX9]]] [[UAX9]]).
+    RDF provides a mechanism for specifying the <a>base direction</a>
+    of a <a>directional language-tagged string</a>
+    to signal the initial text direction of a string.
+    When a string has an overall base direction which cannot be determined by the
     beginning of the string, an external indicator is required,
-    such as the [[HTML]] <a data-cite="HTML/dom.html#the-dir-attribute">dir attribute</a>,
-    which currently has no counterpart for <a>RDF literals</a>.</p>
+    such as the [[HTML]] <a data-cite="HTML/dom.html#the-dir-attribute">dir attribute</a>.</p>
```
aphillips, line 1552: Not just that: strings which can be detected still need bidi isolation. In many formats (and particularly HTML) this is done by applying the base direction to the inserted string value. That is, the `dir` attribute does more than just setting the paragraph direction and saving the user-agent the trouble of finding the first-strong character.

The implication of this text is that only some strings need base direction. In practice many natural language strings would benefit from the metadata (not just the RTL or corner-case ones)

aphillips (changes requested):
Reviewed in the I18N TPAC meeting as prep for tomorrow. Some comments included.

spec/index.html
gkellogg, line 800: Within the context of our specifications, this note really isn't very useful. The only syntax we define where it could possibly apply is RDF/XML, for which XML 1.1 doesn't support base direction.

```suggestion
```

spec/index.html
gkellogg: Only literals having a language tag could benefit from having a base direction, for which we create a new datatype: `rdf:dirLangString`. By definition, base direction is not necessary for a literal with any other datatype.

In the case of RDF, there is substantial data in the wild that does not include any notion of base direction. Based on what you say, arguably, new data should consider the use of directional language-tagged strings rather than simple language-tagged strings.

Is there some suggested change for this language?

spec/index.html
```diff
@@ -786,18 +786,8 @@ <h3>Initial Text Direction</h3>
...
-        for example using the HTML `dir` attribute,
-        it would be correctly presented as follows: <div lang="he" dir="rtl">פעילות הבינאום, W3C</div></p>
-
-      <p class="note">The absence of a <a>base direction</a> does not necessarily imply that
-        the text has no initial text direction;
-        as described in [[[?UAX9]]],
-        strings may be embedded within structures which establish an <em>embedding direction</em>,
-        which determines the default bidirectional orientation of text.</p>
+<p>For example, some text with a language tag "he" might be displayed in a left-to-right context (such as an English web page) as <bdi dir=ltr>פעילות הבינאום, W3C</bdi>, which is incorrect. When provided to a user agent including base direction information (such as using HTML's `dir` attribute) it can then be correctly presented as:</p>
+<div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
```
TallTed, line 790:
```suggestion
       <p>For example, some text with a language tag "`he`" might be displayed
         in a left-to-right context (such as an English web page) as
         <bdi dir=ltr>פעילות הבינאום, W3C</bdi>, which is incorrect. When provided
         to a user agent including base direction information (such as using
         HTML's `dir` attribute) it can then be correctly presented as:
       </p>
       <div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
```

spec/index.html
aphillips: The challenge here is that this implies that only strings with indeterminate direction need base direction metadata. I agree that only language strings need base direction and I18N's suggestion would be that, in general, it's better to use the new `dirLangString` to just `langString`. I recognize that `langString` has to continue to exist.

A different, perhaps more correct, formulation might be something like:

```suggestion
For most human language strings, but particularly for those
whose base direction cannot be accurately determined from the 
string content, is it valuable to have an external indicator in order
to get the proper display and isolation of the value.
One example of such an indicator is
the [[HTML]] <a data-cite="HTML/dom.html#the-dir-attribute">dir attribute</a>.
```

You might also consider adding a pointer to String-Meta here?

gkellogg:
Rebased, after merging in #59.

pchampin:
This was discussed during the TPAC 2023 meeting:
https://www.w3.org/2023/09/12-rdf-star-minutes.html#t03

TallTed (changes requested):

spec/index.html
```diff
@@ -779,7 +801,27 @@ <h2>Literals</h2>
...
+        provides a means of establishing the initial direction of text,
+        including text which is mixture of right-to-left and left-to-right scripts.
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
+
+       <p>For example, some text with a language tag "`he`" might be displayed
+         in a left-to-right context (such as an English web page) as
+         <bdi dir=ltr>פעילות הבינאום, W3C</bdi>, which is incorrect. When provided
```
TallTed, line 819:
```suggestion
         <bdi dir="ltr">פעילות הבינאום, W3C</bdi>, which is incorrect. When provided
```

TallTed (changes requested):

spec/index.html
```diff
@@ -764,12 +785,13 @@ <h2>Literals</h2>
 
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>: Two literals are term-equal (the same
       RDF literal) if and only if the two <a>lexical forms</a>,
-      the two <a>datatype IRIs</a>, and the two
-      <a>language tags</a> (if any) compare equal,
-      using <a data-cite="i18n-glossary#dfn-case-sensitive">case sensitive matching</a>
+      the two <a>datatype IRIs</a>,
```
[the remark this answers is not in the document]
gkellogg: In the abstract syntax, all literals have datatypes. It's only in concrete syntaxes where a "simple literal" only implicitly has a datatype.

spec/index.html
```diff
@@ -779,7 +801,27 @@ <h2>Literals</h2>
     <p>denote the same <a data-lt="literal value">value</a>, but are not the
       same literal <a>RDF terms</a> and are not
       <a>term-equal</a> because their
-      <a>lexical form</a> differs.</p>
+      <a>lexical forms</a> differ.</p>
```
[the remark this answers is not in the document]
gkellogg: It's the lexical values that are the same, and the distinction of the datatype only exists in concrete syntaxes, in the abstract syntax, these would all have the datatype `http://www.w3.org/2001/XMLSchema#string`.

spec/index.html
```diff
@@ -1744,13 +1788,16 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
...
+      a <a>literal</a>,
+      and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
     <li>Improved the use of IRI terminology,
       and added <a href="#iri-abnf" class="sectionRef"></a>.
       This improves the language using <a>relative IRI references</a>
       and clarifies that, in the abstract syntax, IRIs are resolved,
       avoiding the incorrect use of "absolute IRI".</li>
     <li>Changed reference from DOM4, which was not a recommendation at the time, to [[DOM]],
-      making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
+      making thedefinitions of <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
```
Tpt, line 1800: Typo: `thedefinitions` -> `the definitions`

gkellogg:
```suggestion
      making the definitions of <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
```

spec/index.html
```diff
@@ -779,7 +802,27 @@ <h2>Literals</h2>
...
+    <section id="section-text-direction" class="informative">
+      <h3>Initial Text Direction</h3>
+
+      <p>The <a>base direction</a> of a <a>directional language-tagged string</a>
+        provides a means of establishing the initial direction of text,
+        including text which is a mixture of right-to-left and left-to-right scripts.
+        The [[[?UAX9]]] [[?UAX9]] provides support for automatically rendering
+        a sequence of characters in logical order,
+        so that they are visually ordered as expected.
+        However, this is not sufficient to correctly render bidirectional text.</p>
```
TallTed, line 816:
```suggestion
        so that they are visually ordered as expected,
        but this is not sufficient to correctly render bidirectional text.</p>
```

TallTed (changes requested):

PR #56, opened 2023-08-07
Update rdf:HTML and rdf:XMLLiteral definitions with DOM reference.

gkellogg:
This also makes the sections normative, as the text previously said they were informative because DOM4 (now simply DOM) was not a recommendation.

Note some minor changes in DOM method notation to facilitate definition references.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1198,15 +1198,13 @@
       are not required to support either of these facilities.</p>
   </section>
 
-  <section id="section-html" class="informative">
+  <section id="section-html">
     <h3>The <code>rdf:HTML</code> Datatype</h3>
 
     <p>RDF provides for HTML content as a possible <a>literal value</a>.
       This allows markup in literal values. Such content is indicated
       in an <a>RDF graph</a> using a <a>literal</a> whose <a>datatype</a>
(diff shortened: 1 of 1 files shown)


domel (approved):

Tpt (approved):

hartig (approved):

afs:
The changes to the section look good - there is other text that needs updating.

In the intro of section 5.2:

Drop "Furthermore, the following IRIs are allocated for non-normative datatypes:"

Replace by:
```
</p>
<p>The following three datatype IRIs are defined in this section:</p>
```
Add `rdf:JSON` to the list.

(or remove the "furthermore" and drop the list because it's clear enough because they are in the TOC.)

Maybe move the "Semantic extensions" note down on paragraph so that immediately after the list of rdf:HTML,rdf:XMLLIteral,rdf:JSON, is the sentence "RDF processors are not required to recognize datatype IRIs." 

This is important because clarifies that "normative" does not mean "required support" for RDF conformance. RDF terms with bad lexical forms due to datatypes can introduce contradictions (rdf:XMLLIteral used to do this).

TallTed (changes requested):

afs (approved):

PR #57, opened 2023-08-09
Adds a definition for "IRI reference".

gkellogg:
We had defined "IRI" and "relative IRI reference", but not "IRI reference. This allows other specs to refer to RDF Concepts for all such terms.

For #20.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/57.html" title="Last updated on Aug 17, 2023, 8:09 PM UTC (5e3cbe4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/57/bf2d6ea...5e3cbe4.html" title="Last updated on Aug 17, 2023, 8:09 PM UTC (5e3cbe4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -601,7 +601,7 @@
         [[RFC3986]] that permits a wider range of Unicode characters.
         Every absolute URI and URL is an IRI, but not every IRI is an URI.
         In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
-        <a data-cite="RFC3997#section-1.3">section 1.3</a>.
+        <a data-cite="RFC3987#section-1.3">section 1.3</a>.
         An IRI reference is common usage of an Internationalized Resource Identifier.
         An IRI reference may be absolute or relative.
         However, the <a>IRI</a> that results from such a reference
--- a/spec/index.html
+++ b/spec/index.html
@@ -462,7 +462,7 @@
       <a>RDF dataset</a>, for example through the use of
       <a>namespace prefixes</a>,
       <a>IRI references</a>, <a>blank node identifiers</a>,
-      and different ordering of statements. While these aspects can have great
+      and different ordering of triples. While these aspects can have great
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
   </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -603,10 +603,9 @@
         In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
         <a data-cite="RFC3987#section-1.3">section 1.3</a>.
         An IRI reference is common usage of an Internationalized Resource Identifier.
-        An IRI reference may be absolute or relative.
-        However, the <a>IRI</a> that results from such a reference
-        only includes absolute IRIs; any <a>relative IRI references</a> are
-        resolved to their absolute form.
+        An IRI reference refers to either <a>IRI</a> or <a>relative IRI reference</a>,
+        as described by the <em>IRI-reference</em> production in <a href="#iri-abnf" class="sectionRef"></a>.
+        The abstract syntax uses only fully resolved <a>IRIs</a>.
         When IRIs are used in operations that are only
         defined for URIs, they must first be converted according to
         the mapping defined in
--- a/spec/index.html
+++ b/spec/index.html
@@ -461,7 +461,7 @@
       many different ways to encode the same <a>RDF graph</a> or
       <a>RDF dataset</a>, for example through the use of
       <a>namespace prefixes</a>,
-      <a>relative IRI references</a>, <a>blank node identifiers</a>,
+      <a>IRI references</a>, <a>blank node identifiers</a>,
       and different ordering of statements. While these aspects can have great
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
@@ -600,7 +600,13 @@
         <dfn data-lt="URI" data-lt-noDefault><abbr title="Uniform Resource Identifier">URI</abbr>s</dfn>
         [[RFC3986]] that permits a wider range of Unicode characters.
         Every absolute URI and URL is an IRI, but not every IRI is an URI.
-        In RDF, IRIs are used as <em>IRI references</em>, as defined in [[RFC3987]].
+        In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
+        <a data-cite="RFC3997#section-1.3">section 1.3</a>.
+        An IRI reference is common usage of an Internationalized Resource Identifier.
+        An IRI reference may be absolute or relative.
+        However, the <a>IRI</a> that results from such a reference
+        only includes absolute IRIs; any <a>relative IRI references</a> are
+        resolved to their absolute form.
         When IRIs are used in operations that are only
         defined for URIs, they must first be converted according to
         the mapping defined in


spec/index.html
```diff
@@ -600,7 +600,13 @@ <h3>IRIs</h3>
         <dfn data-lt="URI" data-lt-noDefault><abbr title="Uniform Resource Identifier">URI</abbr>s</dfn>
         [[RFC3986]] that permits a wider range of Unicode characters.
         Every absolute URI and URL is an IRI, but not every IRI is an URI.
-        In RDF, IRIs are used as <em>IRI references</em>, as defined in [[RFC3987]].
+        In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
+        <a data-cite="RFC3997#section-1.3">section 1.3</a>.
+        An IRI reference is common usage of an Internationalized Resource Identifier.
```
afs, line 605: What is this trying to say with "common usage"?

 In RDF (the data model), there are only IRIs.

"IRI references" are syntax and resolved by parsers.

spec/index.html
```diff
@@ -600,7 +600,13 @@ <h3>IRIs</h3>
         <dfn data-lt="URI" data-lt-noDefault><abbr title="Uniform Resource Identifier">URI</abbr>s</dfn>
         [[RFC3986]] that permits a wider range of Unicode characters.
         Every absolute URI and URL is an IRI, but not every IRI is an URI.
-        In RDF, IRIs are used as <em>IRI references</em>, as defined in [[RFC3987]].
+        In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
+        <a data-cite="RFC3997#section-1.3">section 1.3</a>.
```
afs, line 604:
```suggestion
        <a data-cite="RFC3987#section-1.3">section 1.3</a>.
```

spec/index.html
```diff
@@ -600,7 +600,13 @@ <h3>IRIs</h3>
...
+        An IRI reference is common usage of an Internationalized Resource Identifier.
+        An IRI reference may be absolute or relative.
+        However, the <a>IRI</a> that results from such a reference
+        only includes absolute IRIs; any <a>relative IRI references</a> are
+        resolved to their absolute form.
```
afs, line 609: Not true - there's a catch.

https://datatracker.ietf.org/doc/html/rfc3986#appendix-D.1

An IRI reference is either an "IRI" or a " relative reference".
```
   URI-reference = URI / relative-ref
```

By spec, absolute means "no fragment" - i.e. the form used HTTP request lines.

The correct term is simply "IRI"

```
   URI-reference = URI / relative-ref
   absolute-URI  = scheme ":" hier-part [ "?" query ]
   relative-ref  = relative-part [ "?" query ] [ "#" fragment ]
```

afs (changes requested):

spec/index.html
gkellogg: The "common usage" came out of RFC3987, and in order to make the definition in RDF Concepts, it made sense to include that language (to me, anyway). We used the term "IRI reference" later in the note, and it is also used in JSON-LD, so better to keep the definition localized, so specs don't need to reference both RDF Concepts and RFC3987.

Even though, as a concept, an "IRI reference" is syntax, and resolved before it makes it into the RDF data model, it is a commonly used term, as is "relative IRI reference" which we define later on.

spec/index.html
gkellogg: From [RFC3987 section 1.3](https://datatracker.ietf.org/doc/html/rfc3987#section-1.3):

> IRI reference: Denotes the common usage of an Internationalized
      Resource Identifier.  An IRI reference may be absolute or
      relative.  However, the "IRI" that results from such a reference
      only includes absolute IRIs; any relative IRI references are
      resolved to their absolute form.  Note that in [[RFC2396](https://datatracker.ietf.org/doc/html/rfc2396)] URIs did
      not include fragment identifiers, but in [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)] fragment
      identifiers are part of URIs.

Given that the term and usage of "IRI reference" isn't going to just go away, we should say something about it, rather than leaving it to RFC3987. By definition an "IRI" is absolute, and a "relative IRI reference" is relative. When syntaxes allow either form, the term "IRI reference" is useful. Do we need further clarification that an IRI reference ultimately resolves to an (absolute) IRI?

```
IRI            = scheme ":" ihier-part [ "?" iquery ]
                     [ "#" ifragment ]
IRI-reference  = IRI / irelative-ref
irelative-ref  = irelative-part [ "?" iquery ] [ "#" ifragment ]
```


These all allow fragments.

spec/index.html
afs: This (in RDF 1.1) 
"IRIs in the RDF abstract syntax MUST be absolute, and MAY contain a fragment identifier."

is not correct. "absolute" means something else.

spec/index.html
afs: RFC 3987 is inconsistent. The grammar of 3987 says: `absolute-IRI   = scheme ":" ihier-part [ "?" iquery ]`. That's the nearest to a definition of absolute IRI. It is only the text in 3987 "to their absolute form" that suggests otherwise. 

RFC 3987 does not replace 3986. 3987 does not claim to do more than widen the characters in certain places.

[RFC 3986 section-4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3) is a section about absolute URIs.

[RFC 3986 D.1](https://datatracker.ietf.org/doc/html/rfc3986#appendix-D.1)

> An ABNF rule for URI has been introduced to correspond to one common
> usage of the term: an absolute URI with optional fragment.



> Do we need further clarification that an IRI reference ultimately resolves to an (absolute) IRI?

An IRI reference resolves to an IRI.
```
IRI-reference  = IRI / irelative-ref
```

spec/index.html
```diff
@@ -600,7 +600,13 @@ <h3>IRIs</h3>
...
         [[RFC3986]] that permits a wider range of Unicode characters.
         Every absolute URI and URL is an IRI, but not every IRI is an URI.
-        In RDF, IRIs are used as <em>IRI references</em>, as defined in [[RFC3987]].
+        In RDF, IRIs are used as <dfn data-lt="iri reference">IRI references</dfn>, as defined in [[RFC3987]]
+        <a data-cite="RFC3987#section-1.3">section 1.3</a>.
+        An IRI reference is common usage of an Internationalized Resource Identifier.
+        An IRI reference may be absolute or relative.
+        However, the <a>IRI</a> that results from such a reference
+        only includes absolute IRIs; any <a>relative IRI references</a> are
+        resolved to their absolute form.
```
gkellogg, line 609: Perhaps if we simplify it to the following:

```suggestion
        An IRI reference refers to either <a>IRI</a> or <a>relative IRI reference</a>,
        as described by the <em>IRI-reference</em> production in <a href="#iri-abnf" class="sectionRef"></a>.
        The abstract syntax uses only fully resolved <a>IRIs</a>.
```

TallTed:
```suggestion
        An IRI reference refers to either an absolute <a>IRI</a> or a <a>relative IRI reference</a>,
        as described by the <em>IRI-reference</em> production in <a href="#iri-abnf" class="sectionRef"></a>.
        The abstract syntax uses only fully resolved, absolute <a>IRIs</a>.
```

spec/index.html
gkellogg: Clarified in https://github.com/w3c/rdf-concepts/pull/57/commits/d26528cf4070ebb1932c6b3d0e8946d7e1943441.

spec/index.html
gkellogg: Clarified in https://github.com/w3c/rdf-concepts/pull/57/commits/d26528cf4070ebb1932c6b3d0e8946d7e1943441.

spec/index.html
```diff
@@ -461,7 +461,7 @@ <h2>RDF Documents and Syntaxes</h2>
       many different ways to encode the same <a>RDF graph</a> or
       <a>RDF dataset</a>, for example through the use of
       <a>namespace prefixes</a>,
-      <a>relative IRI references</a>, <a>blank node identifiers</a>,
+      <a>IRI references</a>, <a>blank node identifiers</a>,
       and different ordering of statements. While these aspects can have great
```
afs, line 465: Unrelated: should that be "triples"?

afs (approved):

spec/index.html
TallTed: @afs -- Which "that"?

afs: "statements"

TallTed: I wouldn't argue with changing `ordering of statements` but given that this is talking about `ways to encode the same <a>RDF graph</a> or <a>RDF dataset</a>`, I think it must become `ordering of triples and/or quads` or possibly `ordering of triples (and/or quads)`.

I see no strong reason not to make that tiny additional change within this PR, but if anyone objects to that, I ask that you create an issue referencing this discussion.

gkellogg, line 465:
```suggestion
      and different ordering of triples. While these aspects can have great
```

gkellogg: Changed to "triples". "quads" aren't appropriate in this context, and N-Quads actually uses "statements" in some circumstances where N-Triples uses "triples". In this context, "triples" is correct.

afs: That would be tricky. Quads only appear in non-normative 4.3.

The definition of RDF Datasets is about graphs, and those graphs are sets of triples.

PR #59, opened 2023-08-18
Improve Unicode terminology and term references.

### LINKED ISSUE #51: Improve Unicode terminology
~Some of the use of "characters" in the spec should be replaced with "code points" to conform with standard recommendations. For example the phrase "compare equal, character by character" should probably be "compare equal, code point by code point".~

Improve the use of Unicode terminology throughout the document. Better reference terminology from i18n-glossary. (Note, there is no convenient externally defined term for "unicode string").

Within a string, constrain code points to be unicode scalar values, which excludes surrogates.

gkellogg:
This tries to improve our use of Unicode terminology. Note that there is no good referencable term for "unicode string" other than [[UNICODE]]. The RDF Literal definition is updated to exclude surrogate code points by referencing "unicode scalar value", which is a normative change.

Most of the Unicode-specific terminology comes from [i18n-glossary](https://w3c.github.io/i18n-glossary).

Fixes #51.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/59.html" title="Last updated on Sep 12, 2023, 9:27 AM UTC (a6bcd68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/59/e4b1b92...a6bcd68.html" title="Last updated on Sep 12, 2023, 9:27 AM UTC (a6bcd68)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -496,6 +496,25 @@
     and still the subject of group discussion. An alternative to conformance
     levels, "profiles", may be adopted instead, abandoned, or described in 
     another specification.</p>
+
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
+      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      in any <a data-cite="i18n-glossary#dfn-character-set">character set</a>
+      or <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
+      However, note that most <a>concrete RDF syntaxes</a>
+      are restricted to the UTF-8 character encoding [[!RFC3629]].
+      Strings are further restricted to those matching the
+      <a data-cite="XML10#charsets">Char</a> production [[XML10]].</p>
+
+    <p>A string is identical to another string if it consists of the same sequence of code points.
+      An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
+      using the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a> form
+      (UTF-8 or UTF-16) without decoding the string into a <a data-cite="i18n-glossary#dfn-scalar-value">scalar value</a> sequence.</p>
+  </section>
 </section>
 
 
@@ -577,7 +596,7 @@
 
     <p>An <dfn data-lt="iri"><abbr title="Internationalized Resource Identifier">IRI</abbr></dfn>
       (Internationalized Resource Identifier) within an RDF graph
-      is a Unicode string [[!UNICODE]] that conforms to the syntax
+      is a <a>string</a> that conforms to the syntax
       defined in RFC 3987 [[!RFC3987]].</p>
 
     <p class="note">For convenience, a complete [[ABNF]] grammar
@@ -738,7 +757,8 @@
       RDF literal) if and only if the two <a>lexical forms</a>,
       the two <a>datatype IRIs</a>, and the two
       <a>language tags</a> (if any) compare equal,
-      using <a data-cite="i18n-glossary#dfn-case-sensitive">case sensitive matching</a>.
+      using <a data-cite="i18n-glossary#dfn-case-sensitive">case sensitive matching</a>
+      (see description of string comparison in <a href="#rdf-strings" class="sectionRef"></a>).
       Thus, two literals can have the same value
       without being the same RDF term. For example:</p>
 
@@ -981,7 +1001,7 @@
     a <a>value space</a> and a <a>lexical-to-value mapping</a>, and
     is denoted by one or more <a>IRIs</a>.</p>
 
-  <p>The <dfn>lexical space</dfn> of a datatype is a set of Unicode strings [[!UNICODE]].</p>
+  <p>The <dfn>lexical space</dfn> of a datatype is a set of <a>strings</a>.</p>
 
   <p>The <dfn>lexical-to-value mapping</dfn> of a datatype is a set of
     pairs whose first element belongs to the <a>lexical space</a>,
@@ -1064,7 +1084,7 @@
       <caption>A list of the RDF-compatible XSD types, with short descriptions</caption>
       <tr><th></th><th>Datatype</th><th>Value space (informative)</th></tr>
 
-      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode strings [[UNICODE]])</td></tr>
+      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode <a>strings</a> [[UNICODE]])</td></tr>
       <tr><td><a data-cite="xmlschema11-2#boolean"><code>xsd:boolean</code></a></td><td>true, false</td></tr>
       <tr><td><a data-cite="xmlschema11-2#decimal"><code>xsd:decimal</code></a></td><td>Arbitrary-precision decimal numbers</td></tr>
       <tr><td><a data-cite="xmlschema11-2#integer"><code>xsd:integer</code></a></td><td>Arbitrary-size integer numbers</td></tr>
@@ -1222,9 +1242,6 @@
       <dt>The IRI denoting this datatype</dt>
       <dd>is <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#HTML</code>.</dd>
 
-      <dt>The lexical space</dt>
-      <dd>is the set of Unicode strings [[!UNICODE]].</dd>
-
       <dt>The value space</dt>
       <dd>is a set of DOM
         <a data-cite="DOM#interface-documentfragment"><code>DocumentFragment</code></a>
@@ -1278,11 +1295,11 @@
       <dd>is <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral</code>.</dd>
 
       <dt id="XMLLiteral-lexical-space">The <a>lexical space</a></dt>
-      <dd>is the set of all Unicode strings [[UNICODE]] which are well-balanced, self-contained
-        <a data-cite="XML10#NT-content">XML content</a>
-        [[XML10]]; and for which embedding between an arbitrary
-        XML start tag and an end tag yields a document conforming to
-       [[[XML-NAMES]]] [[XML-NAMES]].</dd>
+      <dd>is the set of all <a>strings</a> which are well-balanced,
+        self-contained <a data-cite="XML10#NT-content">XML content</a> [[XML10]];
+        and for which embedding between an arbitrary
+        XML start tag and an end tag yields
+        a document conforming to [[[XML-NAMES]]] [[XML-NAMES]].</dd>
 
       <dt id="XMLLiteral-value-space">The <a>value space</a></dt>
         <dd>is a set of DOM
@@ -1462,7 +1479,7 @@
     Security/privacy protocols must be imposed which reflect the sensitivity of the embedded information.</p>
 
   <p>RDF can express data which is presented to the user, such as RDF Schema labels.
-    Applications rendering strings retrieved from untrusted RDF documents,
+    Applications rendering <a>strings</a> retrieved from untrusted RDF documents,
     or using unescaped characters,
     SHOULD use warnings and other appropriate means to limit the possibility
     that malignant strings might be used to mislead the reader.
@@ -1494,7 +1511,7 @@
 
 <section id="internationalization" class="appendix informative">
   <h2>Internationalization Considerations</h2>
-  <p>RDF is restricted to representing Unicode string [[UNICODE]] values with left-to-right or right-to-left direction indicators.
+  <p>RDF is restricted to representing Unicode <a>string</a> [[UNICODE]] values with left-to-right or right-to-left direction indicators.
     RDF provides a mechanism for specifying the language associated with
     a string (<a>language-tagged string</a>),
     but does not provide a means of indicating the base direction of the string.</p>
@@ -1655,9 +1672,10 @@
     <li>Changed reference from DOM4, which was not a recommendation at the time, to [[DOM]],
       making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
     <li>Clarify Unicode terminology,
-      using a data-lt="code point" class="lint-ignore">Unicode code points</a>,
+      using <a data-lt="code point" class="lint-ignore">Unicode code points</a>,
       and restriction to the XML <a data-cite="XML10#charsets">Char</a> production.
-      Also removes obsolete recommendations for the use of Normalization Form C in literals.</li>
+      Also removes obsolete recommendations for the use of Normalization Form C in literals.
+      Adds a definition of <a>string</a> that can be used in other RDF documents.</li>
   </ul>
 
   <p class="note">A detailed overview of the differences between RDF versions&nbsp;1.0
--- a/spec/index.html
+++ b/spec/index.html
@@ -509,7 +509,7 @@ of the UTF-8 character encoding [[!RFC3629]],
 which also prevents the use of certain non-character values.
       are restricted to the UTF-8 character encoding [[!RFC3629]].
       Strings are further restricted to those matching the
-      <a data-cite="XML10#charsets">Char</a> production [[XML10]].</p>
+      <a data-cite="XML11#charsets">Char</a> production [[XML11]].</p>
 
     <p>A string is identical to another string if it consists of the same sequence of code points.
       An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
@@ -1298,7 +1298,7 @@ which also prevents the use of certain non-character values.
 
       <dt id="XMLLiteral-lexical-space">The <a>lexical space</a></dt>
       <dd>is the set of all <a>strings</a> which are well-balanced,
-        self-contained <a data-cite="XML10#NT-content">XML content</a> [[XML10]];
+        self-contained <a data-cite="XML11#NT-content">XML content</a> [[XML11]];
         and for which embedding between an arbitrary
         XML start tag and an end tag yields
         a document conforming to [[[XML-NAMES]]] [[XML-NAMES]].</dd>
@@ -1675,7 +1675,7 @@ which also prevents the use of certain non-character values.
       making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
     <li>Clarify Unicode terminology,
       using <a data-lt="code point" class="lint-ignore">Unicode code points</a>,
-      and restriction to the XML <a data-cite="XML10#charsets">Char</a> production.
+      and restriction to the XML <a data-cite="XML11#charsets">Char</a> production.
       Also removes obsolete recommendations for the use of Normalization Form C in literals.
       Adds a definition of <a>string</a> that can be used in other RDF documents.</li>
   </ul>
--- a/spec/index.html
+++ b/spec/index.html
@@ -695,7 +695,7 @@
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
         <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, thus, they do not contain Unicode surrogate code points.</li>
+        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, and therefore do not contain Unicode surrogate code points.</li>
       <li>a <dfn>datatype IRI</dfn>, being an <a>IRI</a>
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>, and</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1650,7 +1650,8 @@
       making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
     <li>Clarify Unicode terminology,
       using a data-lt="code point" class="lint-ignore">Unicode code points</a>,
-      and restriction to the XML <a data-cite="XML10#charsets">Char</a> production.</li>
+      and restriction to the XML <a data-cite="XML10#charsets">Char</a> production.
+      Also removes obsolete recommendations for the use of Normalization Form C in literals.</li>
   </ul>
 
   <p class="note">A detailed overview of the differences between RDF versions&nbsp;1.0
--- a/spec/index.html
+++ b/spec/index.html
@@ -587,7 +587,7 @@
 
     <p><dfn>IRI equality</dfn>:
       Two IRIs are the same if and only if they compare as identical when
-      considered as character strings, as in Simple String Comparison in
+      considered as character strings (unicode strings), as in Simple String Comparison in
       <a data-cite="rfc3987#section-5.3.1">section 5.3.1</a>
       of [[!RFC3987]].
       (This is done in the abstract syntax, so the IRIs are absolute
@@ -738,7 +738,8 @@
       RDF literal) if and only if the two <a>lexical forms</a>,
       the two <a>datatype IRIs</a>, and the two
       <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
+      using <a data-cite="i18n-glossary#dfn-case-sensitive">character sensitive matching</a>.
+      Thus, two literals can have the same value
       without being the same RDF term. For example:</p>
 
     <pre>
--- a/spec/index.html
+++ b/spec/index.html
@@ -503,8 +503,7 @@
     <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
       is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a>
-      in any <a data-cite="i18n-glossary#dfn-character-set">character set</a>
-      or <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
+      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
       However, note that most <a>concrete RDF syntaxes</a>
       are restricted to the UTF-8 character encoding [[!RFC3629]].
       Strings are further restricted to those matching the
--- a/spec/index.html
+++ b/spec/index.html
@@ -197,7 +197,8 @@
       </figcaption>
     </figure>
 
-    <p>A variation on the graph shown in <a href="#fig-quoted-triple"></a> can be described where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
+    <p>A variation on the graph shown in <a href="#fig-quoted-triple"></a> can be described
+      where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
 
     <figure id="fig-quoted-asserted-triple">
       <a href="quoted-asserted-triple.svg">
@@ -208,7 +209,8 @@
              aria-describedby="fig-quoted-asserted-triple-alt"/>
       </a>
       <figcaption id="fig-quoted-asserted-triple-alt">
-        An RDF graph containing two triples where one of them contains the other as the subject; hence, that other triple is both quoted and asserted.
+        An RDF graph containing two triples where one of them contains the other as the subject;
+        hence, that other triple is both quoted and asserted.
       </figcaption>
     </figure>
 
@@ -503,7 +505,11 @@
     <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
       or <a data-lt="string">RDF string</a>,
-      `is used to describe an ordered` sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>. Unicode scalar values do not include the <a data-cite="i18n-glossary#dfn-surrogate"  class="lint-ignore">surrogate code points</a>.
+      `is used to describe an ordered` sequence of zero or more
+      <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
+      Unicode scalar values do not include the
+      <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
       Note that most <a>concrete RDF syntaxes</a> require the use
       of the UTF-8 character encoding [[!RFC3629]], 
       which requires that certain non-character values be expressed
@@ -511,9 +517,11 @@
 </p>
 
     <p>A string is identical to another string if it consists of the same sequence of code points.
-      An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
+      An implementation MAY determine string equality by comparing the
+      <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
       that use the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a>
-      (UTF-8 or UTF-16) without decoding the string into a <a data-lt="code point" class="lint-ignore">Unicode code point</a> sequence.</p>
+      (UTF-8 or UTF-16) without decoding the string into a
+      <a data-lt="code point" class="lint-ignore">Unicode code point</a> sequence.</p>
   </section>
 </section>
 
@@ -695,7 +703,9 @@
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
         <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, and therefore do not contain <a data-cite="i18n-glossary#dfn-surrogate"  class="lint-ignore">Unicode surrogate code points</a>.</li>
+        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>,
+        and therefore do not contain
+        <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">Unicode surrogate code points</a>.</li>
       <li>a <dfn>datatype IRI</dfn>, being an <a>IRI</a>
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>, and</li>
@@ -852,8 +862,9 @@
 
     <p id="section-graph-equality">Two
       <a>RDF graphs</a> <var>G</var> and <var>G'</var> are
-      <dfn data-lt="graph isomorphism|isomorphic" data-lt-noDefault class="export">isomorphic</dfn> (that is, they have an identical
-      form) if there is a bijection <var>M</var> between the sets of nodes of the two
+      <dfn data-lt="graph isomorphism|isomorphic" data-lt-noDefault class="export">isomorphic</dfn>
+      (that is, they have an identical form)
+      if there is a bijection <var>M</var> between the sets of nodes of the two
       graphs, such that:</p>
 
     <ol>
--- a/spec/index.html
+++ b/spec/index.html
@@ -506,15 +506,15 @@
       in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
       Note that most <a>concrete RDF syntaxes</a> require the use
       of the UTF-8 character encoding [[!RFC3629]], 
-      which also prevents the use of certain non-character values.
+      which requires that certain non-character values be expressed
+      using the `\u0000` or `\U00000000` forms.
       are restricted to the UTF-8 character encoding [[!RFC3629]].
-      Strings are further restricted to those matching the
-      <a data-cite="XML11#charsets">Char</a> production [[XML11]].</p>
+</p>
 
     <p>A string is identical to another string if it consists of the same sequence of code points.
       An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
       using the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a> form
-      (UTF-8 or UTF-16) without decoding the string into a <a data-cite="i18n-glossary#dfn-scalar-value">scalar value</a> sequence.</p>
+      (UTF-8 or UTF-16) without decoding the string into a <a data-lt="code point" class="lint-ignore">Unicode code point</a> sequence.</p>
   </section>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -503,8 +503,7 @@
     <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
       or <a data-lt="string">RDF string</a>,
-      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>, which do not include Unicode surrogate code points.
-      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
+      `is used to describe an ordered` sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>. Unicode scalar values do not include the <a data-cite="i18n-glossary#dfn-surrogate"  class="lint-ignore">surrogate code points</a>.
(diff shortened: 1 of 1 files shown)


Tpt (approved):

domel (approved):

hartig (approved):

spec/index.html
```diff
@@ -1646,6 +1648,7 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
       and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
     <li>Changed reference from DOM4, which was not a recommendation at the time, to [[DOM]],
       making the <a>rdf:HTML</a> and <a>rdf:XMLLiteral</a> datatypes normative.</li>
+    <li>Clarify Unicode terminology.</li>
```
gkellogg, line 1651:
```suggestion
    <li>Clarify Unicode terminology. Excludes <a data-cite="i18n-glossary#dfn-surrogate">surrogate code points</a> from literals.</li>
```

spec/index.html
```diff
@@ -576,7 +576,7 @@ <h3>IRIs</h3>
 
     <p>An <dfn data-lt="iri"><abbr title="Internationalized Resource Identifier">IRI</abbr></dfn>
       (Internationalized Resource Identifier) within an RDF graph
-      is a Unicode string [[!UNICODE]] that conforms to the syntax
+      is a unicode string [[!UNICODE]] that conforms to the syntax
```
TallTed, line 579: Are we going with `Unicode` (which appears to me to be correct) or `unicode`? And are we using one form throughout the documents, or changing for some reason(s) that are not clear to me?

This PR currently has a mix of `Unicode` and `unicode` in the human-facing text.

I think we should be using one form throughout the documents, and that it should have the capital `U`, i.e., `Unicode` (as I note is used throughout the [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/), not to mention [the Unicode website](https://home.unicode.org/)).

afs: Isn't it because, in this case, "unicode string" is informal (local) terminology?

afs (approved):

spec/index.html
TallTed: The proximal `[[!UNICODE]]` citation suggests it's not so informal.

In such case, I would hope for a glossary entry or similar for "unicode string" that would differentiate it from "Unicode" and/or "Unicode other things"...

I also note [The Unicode® Consortium Name and Trademark Usage Policy](https://www.unicode.org/policies/logo_policy.html) with which I think we should comply, unless there's a strong argument not to (and W3C legal OKs such non-compliance).

gkellogg: It is unfortunate that there is no good reference for a "string" term; i18n-glossary says its intended to be used along INFRA, but there a string is defined based on JavaScript, which isn't helpful.

I agree with @TallTed that we should be consistent and use "Unicode string" uniformly; I can add a commit to do so here. In lieu of a term reference, we can make sure we're using [[UNICODE]] in addition, where appropriate.

TallTed (approved):

spec/index.html
TallTed: I think we can mark this thread resolved.

aphillips (changes requested):
I'm going to leave a comment on #51 in a second. I made some suggestions below, but mostly I have questions about what you're trying to accomplish (how I18N can best guide your terminology choices) and what you mean by "Unicode string"

spec/index.html
```diff
@@ -658,8 +658,7 @@ <h3>IRIs</h3>
           “<code>%3f</code>”)</li>
         <li>Punycode-encoding of Internationalized Domain Names
           in IRIs</li>
-        <li>IRIs that are not in Unicode Normalization
-          Form C [[NFC]]</li>
+        <li>IRIs that are not in Unicode <a data-cite="i18n-glossary#dfn-unicode-normalization-form-c">Normalization Form&nbsp;C</a> [[UNICODE]]</li>
```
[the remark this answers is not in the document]
gkellogg: Thanks, I did attempt to do that, but ran into some ReSpec issues that made it more difficult than you would think. Same for adding RDF12-CONCEPTS to `xref` in other specs. Needs to be sorted out at some point.

spec/index.html
```diff
@@ -673,8 +672,11 @@ <h2>Literals</h2>
       elements:</p>
 
     <ul>
-      <li>a <dfn>lexical form</dfn>, being a Unicode [[!UNICODE]] string,
-        which SHOULD be in Normal Form&nbsp;C [[!NFC]],</li>
+      <li>a <dfn>lexical form</dfn>, being a Unicode string [[!UNICODE]],
+        which SHOULD be in <a data-cite="i18n-glossary#dfn-unicode-normalization-form-c">Normalization Form&nbsp;C</a> [[!UNICODE]]
+        and where all <a data-cite="i18n-glossary#dfn-code-point">code points</a> MUST be
+        <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>
+        (i.e., not include <a data-cite="i18n-glossary#dfn-surrogate">surrogate code points</a>).</li>
```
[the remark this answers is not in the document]
gkellogg: Quick response:

"USVString" might be an alternative to "Unicode string" for us to consider, although  referencing "USVString" would create a normative reference to WebIDL, which doesn't seem right. The specs go back a long ways, and the use of "Unicode string" is pretty pervasive, so we were attempting to try to limit the downstream changes needed. 

Even though we use the term, it doesn't seem proper to define a "Unicode string" term formally in this spec. (Something concept in i18n-glossary would be useful for a generic sequence of code units might be useful).

Binding the abstract data model to UTF-16, particularly when most concrete forms favor UTF-8, doesn't seem like the right way to go for an abstract syntax. The INFRA definition of "string" would also bind us too closely to UTF-16.

Literals generally derive from `xsd:string`, so the XML "Char" production might be a useful way to restrict the representation. Generally, other than for IRIs, the main concept is for the [lexical form](https://www.w3.org/TR/rdf12-concepts/#dfn-lexical-form) of a literal, which we could possibly define as an "RDF string" term.

The NFC bits are also historic, and removing that would be helpful.

spec/index.html
```diff
@@ -587,7 +587,7 @@ <h3>IRIs</h3>
 
     <p><dfn>IRI equality</dfn>:
       Two IRIs are the same if and only if they compare as identical when
-      considered as character strings, as in Simple String Comparison in
+      considered as character strings (Unicode strings), as in Simple String Comparison in
```
gkellogg, line 590:
```suggestion
      Two IRIs are the same if and only if they consist of the same sequence of <a data-cite="i18n-glossary#dfn-code-point">Unicode code points</a>, as in Simple String Comparison in
```

spec/index.html
```diff
@@ -736,7 +738,8 @@ <h2>Literals</h2>
       RDF literal) if and only if the two <a>lexical forms</a>,
       the two <a>datatype IRIs</a>, and the two
       <a>language tags</a> (if any) compare equal,
-      character by character. Thus, two literals can have the same value
+      using <a data-cite="i18n-glossary#dfn-case-sensitive">character sensitive matching</a>.
```
gkellogg, line 741:
```suggestion
      using <a data-cite="i18n-glossary#dfn-case-sensitive">case sensitive matching</a>.
```

spec/index.html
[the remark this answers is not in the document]
aphillips: Thanks. I'd be glad to hear if there are changes to our instructions that are needed that would make lives easier.

spec/index.html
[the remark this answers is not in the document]
aphillips: @gkellogg Thanks for this! I agree that there is room for improvement in I18N's materials, given that the INFRA stuff isn't really intended for the XML (and friends) world. Your reply clarifies for me what the intention is, which makes it easier to recommend stuff.

Food for thought: I think it is fine to lean into the "code point string" "character-by-character" definition of a Unicode string and then, where necessary, include some normative gunk that allows implementations to do string identity without having to decode the scalar value. Code unit comparison of two strings in the same Unicode encoding (byte-by-byte in UTF-8 or 16-bit code unit-by-code unit in UTF-16) work _exactly the same_ as comparing the scalar values, except that an isolated surrogate in either encoding "decodes" to U+FFFD 🤷

spec/index.html
```diff
@@ -1062,7 +1084,7 @@ <h3>The XML Schema Built-in Datatypes</h3>
       <caption>A list of the RDF-compatible XSD types, with short descriptions</caption>
       <tr><th></th><th>Datatype</th><th>Value space (informative)</th></tr>
 
-      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode character strings)</td></tr>
+      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode <a>strings</a> [[UNICODE]])</td></tr>
```
aphillips, line 1087: I'm still not sure what the parenthetical is meant to do. Is it meant to call attention to the fact that `xsd:string` is restricted to `Char`?

spec/index.html
```diff
@@ -1492,12 +1511,12 @@ <h2>Security Considerations</h2>
 
 <section id="internationalization" class="appendix informative">
   <h2>Internationalization Considerations</h2>
-  <p>RDF is restricted to representing string values with left-to-right or right-to-left direction indicators.
+  <p>RDF is restricted to representing Unicode <a>string</a> [[UNICODE]] values with left-to-right or right-to-left direction indicators.
```
aphillips, line 1514: This sentence calls for more attention for reasons other than its use of `string` as a term. I think it is factually inaccurate or maybe some words were left out. Perhaps:

```suggestion
<p>RDF does not provide a means of associating <a>base direction</a> with a <a>string</a> other than via the inclusion of Unicode [[UNICODE]] bidirectional control characters inside the string itself.
```

aphillips (changes requested):
My comments are mostly minor and, in one case, slightly orthogonal to this PR's goal. Overall I think this looks good.

Note that I18N gave me an action to add XML-related string definitions to the glossary, best practices to our best practices doc, and a quotable reference for the string equality bits to charmod-norm in today (2023-08-31)'s teleconference. That will take a little time to land by might help you in the future by providing linkable stuff you can just reference.

spec/index.html
```diff
@@ -495,6 +496,25 @@ <h2>RDF Documents and Syntaxes</h2>
...
     levels, "profiles", may be adopted instead, abandoned, or described in 
     another specification.</p>
+
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
+      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      in any <a data-cite="i18n-glossary#dfn-character-set">character set</a>
```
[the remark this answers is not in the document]
aphillips: (Note: "this line" refers to:

> in any <a data-cite="i18n-glossary#dfn-character-set">character set</a>

spec/index.html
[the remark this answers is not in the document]
gkellogg: This is resolved by using your suggested language, along with a reference to the XML Char production.

spec/index.html
```diff
@@ -495,6 +496,25 @@ <h2>RDF Documents and Syntaxes</h2>
...
+      or <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
```
gkellogg, line 507: I think it is gratuitous, and we can just limit it to any character encoding.

```suggestion
      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
```

spec/index.html
```diff
@@ -495,6 +496,25 @@ <h2>RDF Documents and Syntaxes</h2>
...
+      However, note that most <a>concrete RDF syntaxes</a>
```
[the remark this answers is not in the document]
gkellogg: This is defined in the section on [RDF Documents and Syntaxes](https://www.w3.org/TR/rdf12-concepts/#rdf-documents)

> An RDF document is a document that encodes an [RDF graph](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-graph) or [RDF dataset](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-dataset) in a concrete RDF syntax, such as Turtle [[RDF12-TURTLE](https://www.w3.org/TR/rdf12-concepts/#bib-rdf12-turtle)], RDFa [[RDFA-PRIMER](https://www.w3.org/TR/rdf12-concepts/#bib-rdfa-primer)], JSON-LD [[JSON-LD11](https://www.w3.org/TR/rdf12-concepts/#bib-json-ld11)], or TriG [[RDF12-TRIG](https://www.w3.org/TR/rdf12-concepts/#bib-rdf12-trig)]. RDF documents enable the exchange of RDF graphs and RDF datasets between systems.

It's not explicit here that this is a byte stream, but, for example, an [N-Triples document](https://www.w3.org/TR/rdf12-n-triples/#dfn-n-triples-document) _is_ defined as a Unicode string. (That language will need to be updated after this PR is done). And later: "The content encoding of N-Triples is always UTF-8."

spec/index.html
gkellogg: This is a legacy description, that I don't really understand either (maybe @afs does?) I think it could just say "Character <a>strings</a>", as I'm not clear on what strings could be represented that are not Unicode, particularly given the restriction on the XML Char production.

spec/index.html
gkellogg: Note that there is a long-pending PR #48 on adding base direction, which I believe touches this. That PR has been held up for a long time due to lack of consensus and understanding on the requirements for addressing text direction in RDF. I believe this needs to be resolved on a call with interested parties and hope that time at TPAC can be reserved to address this.

gkellogg:
@aphillips said:

> Note that I18N gave me an action to add XML-related string definitions to the glossary, best practices to our best practices doc, and a quotable reference for the string equality bits to charmod-norm in today (2023-08-31)'s teleconference. That will take a little time to land by might help you in the future by providing linkable stuff you can just reference.

I'd much rather reference something in  i18n-glossary than define our on "string" term 😄.

pfps:
It seems wrong for RDF concepts to even discuss character encodings here.  Character sets or character encodings do not matter at all for RDF strings, just the Unicode code points.

As far as I can tell well-formed Unicode code unit sequences cannot encode surrogate code points.  And this is the same distinction that forbids values that are not Unicode code points at all.  (From D84 in The Unicode Standard Version 15.0 - Core Specification - "Any code unit sequence that would correspond to a code point outside the defined rante of Unicode scalar values would, for example, be ill-formed.")

So I don't think it is acceptable to say that an RDF string is a sequence of Unicode code points, as just about any system that handles Unicode may only accept well-formed Unicode code unit sequences.  Instead of Unicode code points Unicode scalar values much be used.

gkellogg:
> It seems wrong for RDF concepts to even discuss character encodings here. Character sets or character encodings do not matter at all for RDF strings, just the Unicode code points.

Character encodings are discussed as relates to concrete syntaxes. If you think it would improve the text, we could remove "... in any character encoding" from the first paragraph of Strings in RDF.

> As far as I can tell well-formed Unicode code unit sequences cannot encode surrogate code points. And this is the same distinction that forbids values that are not Unicode code points at all. (From D84 in The Unicode Standard Version 15.0 - Core Specification - "Any code unit sequence that would correspond to a code point outside the defined rante of Unicode scalar values would, for example, be ill-formed.")

Note that strings are also restricted to the XML Char production, which excludes surrogates.

> So I don't think it is acceptable to say that an RDF string is a sequence of Unicode code points, as just about any system that handles Unicode may only accept well-formed Unicode code unit sequences. Instead of Unicode code points Unicode scalar values much be used.

What is accepted is a subject for concrete syntaxes, which are also restricted to the XML Char production. Code points seems like a reasonable representation for an abstract syntax.  @aphillips discussed his thought process on using code points vs. code units above. As we're not experts in Unicode, shouldn't we defer to such expert guidance?

pfps:
s ar

pfps:
As far as I can tell, @aphillips did not express a preference between Unicode code point and Unicode scalar value.  But, as you say, there is a further restriction of literal values (which is what I'm more interested in, but which also restricts RDF strings) to XML Char (currently the old one that excludes most control characters, but I think the reference is supposed to be updated to 1.1 which only excludes three scalar values).   So I guess things are fine then. It is a bit strange to exclude these three scalar values and not the other noncharacter scalar values - but not worth worrying about.

spec/index.html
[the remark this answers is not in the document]
aphillips: Thanks. That's helpful.

So this refers to a document _format_ that, when serialized on the wire, will almost always these days be required to be serialized in UTF-8 (and thus, the resulting byte string will be a UTF-8 code unit string... how thick is my jargon? It has the consistency of concrete...)

I'd suggest a tweak here as a result. I'd move the "Strings are further..." sentence in front of this one and modify this to say something like:

```suggestion
Note that most <a>concrete RDF syntaxes</a> require the use
of the UTF-8 character encoding [[!RFC3629]], 
which also prevents the use of certain non-character values.
```

spec/index.html
aphillips: I'm guessing it's the XML-Char limitation. If so, it would probably be better to just say that?

spec/index.html
aphillips: Totally. I'll send an email to the various chairs to see if we can align schedules.

spec/index.html
gkellogg:
```suggestion
      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (see <a>string</a>)</td></tr>
```

TallTed (approved):

spec/index.html
```diff
@@ -674,8 +694,9 @@ <h2>Literals</h2>
       elements:</p>
 
     <ul>
-      <li>a <dfn>lexical form</dfn>, being a Unicode [[!UNICODE]] string,
-        which SHOULD be in Normal Form&nbsp;C [[!NFC]],</li>
+      <li>a <dfn>lexical form</dfn> consisting of a sequence of
+        <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
+        matching the <a data-cite="XML10#charsets">Char</a> production [[XML10]].</li>
```
pfps, line 699: Concepts already uses XML 1.1.   This reference needs to be updated.

pfps, line 699:
```suggestion
        matching the <a data-cite="XML11#charsets">Char</a> production [[XML11]].</li>
       Because RDF lexical forms match the Char production of XML they do not contain Unicode surrogate code points.
```

spec/index.html
```diff
@@ -495,6 +496,26 @@ <h2>RDF Documents and Syntaxes</h2>
     and still the subject of group discussion. An alternative to conformance
     levels, "profiles", may be adopted instead, abandoned, or described in 
     another specification.</p>
+
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
+      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a>
```
pfps, line 505: Because the lexical forms of RDF literals do not contain surrogate code points, neither do RDF strings.
```suggestion
      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> matching the <a data-cite="XML11#charsets">Char</a> production [[XML11]], and thus do not contain Unicode surrogate code points.
```

spec/index.html
gkellogg: There are a couple of specs which still reference XML10, that need to be updated to XML11.

spec/index.html
```diff
@@ -495,6 +496,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
+      Note that most <a>concrete RDF syntaxes</a> require the use
+      of the UTF-8 character encoding [[!RFC3629]], 
+      which also prevents the use of certain non-character values.
+      are restricted to the UTF-8 character encoding [[!RFC3629]].
+      Strings are further restricted to those matching the
+      <a data-cite="XML11#charsets">Char</a> production [[XML11]].</p>
```
afs, line 512: See comment on https://github.com/w3c/rdf-concepts/issues/51#issuecomment-1707886347

Lexical forms in the abstract data model are not limited to displayable text. Just because thios WG doesn't see a usage for certain code points, it doesn't mean they should be excluded from the abstract data model. Discouragement or advice to avoid certain forms can be given as it is current - use normal form C.

```suggestion
</p>
```

spec/index.html
```diff
@@ -495,6 +496,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+
+    <p>A string is identical to another string if it consists of the same sequence of code points.
+      An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
+      using the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a> form
+      (UTF-8 or UTF-16) without decoding the string into a <a data-cite="i18n-glossary#dfn-scalar-value">scalar value</a> sequence.</p>
```
afs, line 517:
```suggestion
      (UTF-8 or UTF-16) without decoding the string into a <a>Unicode code point</a> sequence.</p>
```

spec/index.html
```diff
@@ -495,6 +496,26 @@ <h2>RDF Documents and Syntaxes</h2>
...
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
+      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
+      Note that most <a>concrete RDF syntaxes</a> require the use
+      of the UTF-8 character encoding [[!RFC3629]], 
+      which also prevents the use of certain non-character values.
```
afs, line 509: They can be inserted using `\u` and `\U`.

afs, line 509: They can be inserted using `\u` and `\U`. This sentence should be removed.

```suggestion
```

spec/index.html
afs: Turtle:
```
[39]  STRING_LITERAL_QUOTE  ::=  '"' ([^#x22#x5C#x0A#x0D] | ECHAR | UCHAR)* '"'
```

Or use  `\u` or `\U` escapes.

```
:subject :predicate "\u0000" 
```

afs: XML Char production does not include U+0000, the "noncharacters" FFFE, FFFF ("noncharacters").
It advises against certain other characters which we would effectively inherit - I think makes it not suitable for RDF Concepts.

spec/index.html
TallTed: tweaking @pfps's language
```suggestion
      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which matches the <a data-cite="XML11#charsets">Char</a> production of [[XML11]], and thus does not contain Unicode surrogate code points.
```

spec/index.html
TallTed: Might it be better amended? Perhaps as --
```suggestion
      Note that most <a>concrete RDF syntaxes</a> require the use
      of the UTF-8 character encoding [[!RFC3629]], 
      which requires that certain non-character values be expressed
      using the `\u0000` or `\U00000000` forms.
```

spec/index.html
```diff
@@ -504,12 +504,12 @@ <h2>Strings in RDF</h2>
...
-Note that most <a>concrete RDF syntaxes</a> require the use
-of the UTF-8 character encoding [[!RFC3629]], 
-which also prevents the use of certain non-character values.
+      Note that most <a>concrete RDF syntaxes</a> require the use
+      of the UTF-8 character encoding [[!RFC3629]], 
+      which also prevents the use of certain non-character values.
       are restricted to the UTF-8 character encoding [[!RFC3629]].
       Strings are further restricted to those matching the
-      <a data-cite="XML10#charsets">Char</a> production [[XML10]].</p>
+      <a data-cite="XML11#charsets">Char</a> production [[XML11]].</p>
```
TallTed, line 512:
```suggestion
      <a data-cite="XML11#charsets">Char</a> production of [[XML11]].</p>
```

spec/index.html
```diff
@@ -696,7 +696,8 @@ <h2>Literals</h2>
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
         <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        matching the <a data-cite="XML10#charsets">Char</a> production [[XML10]].</li>
+        matching the <a data-cite="XML11#charsets">Char</a> production [[XML11]].
+       Because RDF lexical forms match the Char production of XML they do not contain Unicode surrogate code points.</li>
```
TallTed, line 700:
```suggestion
        matching the <a data-cite="XML11#charsets">Char</a> production of [[XML11]].
        Since RDF lexical forms match the <a data-cite="XML11#charsets">Char</a>
        production of [[XML11]], they do not contain Unicode surrogate code points.</li>
```

TallTed (changes requested):

spec/index.html
gkellogg: I believe @afs is objecting to the use of the Char production in RDF Concepts, and limiting it to just describe a sequence of zero or more Unicode code points without further restriction.

spec/index.html
gkellogg: If we remove the restriction for the Char production, literals may contain surrogates.

pfps: It would be possible to limit the "lexical form" of language-tagged strings - to scalar values, for example -  but make the space of lexical forms larger - code points or even 32-bit integers.

spec/index.html
gkellogg:
```suggestion
is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>, and thus does not contain Unicode surrogate code points.
```

spec/index.html
gkellogg:
```suggestion
        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, thus, they do not contain Unicode surrogate code points.</li>
```

spec/index.html
```diff
@@ -1062,7 +1086,7 @@ <h3>The XML Schema Built-in Datatypes</h3>
       <caption>A list of the RDF-compatible XSD types, with short descriptions</caption>
       <tr><th></th><th>Datatype</th><th>Value space (informative)</th></tr>
 
-      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (but not all Unicode character strings)</td></tr>
+      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character strings (see <a>string</a>)</td></tr>
```
gkellogg, line 1089: @afs, @pfps is this correct with the updated definition? I'm not clear on what `xsd:string` literals would not now be considered a string.

pfps: It's the other way around.  The value space for xsd:string (either version) is a proper subset of Unicode scalar values.  So pointing to the new section on strings gives the impression that xsd:string is larger than it really is.

spec/index.html
TallTed: Once more into the breach...
```suggestion
      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>, which do not include Unicode surrogate code points.
```

spec/index.html
```diff
@@ -506,15 +506,15 @@ <h2>Strings in RDF</h2>
       in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
       Note that most <a>concrete RDF syntaxes</a> require the use
       of the UTF-8 character encoding [[!RFC3629]], 
-      which also prevents the use of certain non-character values.
+      which requires that certain non-character values be expressed
+      using the `\u0000` or `\U00000000` forms.
       are restricted to the UTF-8 character encoding [[!RFC3629]].
```
TallTed, line 511: This line appears to be a relic that no longer belongs here.
```suggestion
```

spec/index.html
```diff
@@ -506,15 +506,15 @@ <h2>Strings in RDF</h2>
...
-      Strings are further restricted to those matching the
-      <a data-cite="XML11#charsets">Char</a> production [[XML11]].</p>
+</p>
 
     <p>A string is identical to another string if it consists of the same sequence of code points.
       An implementation MAY determine string equality by comparing the <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
       using the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a> form
```
TallTed, line 516:
```suggestion
      that use the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a>
```

spec/index.html
gkellogg:
```suggestion
      <tr><th rowspan="4">Core types</th><td><a data-cite="xmlschema11-2#string"><code>xsd:string</code></a></td><td>Character <a>strings</a> matching the <a data-cite="XML11#charsets">Char</a> production from [[XML11]]</td></tr>
```

spec/index.html
```diff
@@ -696,8 +695,7 @@ <h2>Literals</h2>
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
         <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        matching the <a data-cite="XML11#charsets">Char</a> production [[XML11]].
-       Because RDF lexical forms match the Char production of XML they do not contain Unicode surrogate code points.</li>
+        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, thus, they do not contain Unicode surrogate code points.</li>
```
TallTed, line 698:
```suggestion
        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, and therefore do not contain Unicode surrogate code points.</li>
```

spec/index.html
afs: This is better wording but remove the link to `<a>string</a>`

A general (RDF) "string" as used for lexical forms should be independent of XML, and as comprehensive as possible. Anything excluded should have a reason.

Just "Character strings", no link is enough given the link on the previous column.

Or

"Character sequence matching the ..."

spec/index.html
```diff
@@ -495,6 +496,25 @@ <h2>RDF Documents and Syntaxes</h2>
...
     levels, "profiles", may be adopted instead, abandoned, or described in 
     another specification.</p>
+
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
+      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>, which do not include Unicode surrogate code points.
+      in any <a data-cite="i18n-glossary#dfn-character-encoding">character encoding</a>.
```
afs, line 506:
```suggestion
      is used to describe an ordered sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>. Unicode scalar values do not include the high-surrogate or low-surrogate code points.
```
This fixes some sentence damage in the working text - "code points. in any"

The language in the Unicode spec seems to be e.g. "high-surrogate and low-surrogate code points are designated for surrogate code units"

The "in any character encoding" is confusing given UTF-16. The key point about UTF-8 is covered by the next sentence. Encoding for XML is format sis left to XML (it doesn't even have to be unicode).

afs: I've found places in the Uniocde spec that say "surrogate code points" (D15). While most places say "high-surrogate ... low-surrogate", 

Suggestion chnaged:
--
do not
include the high-surrogate or low-surrogate code points.
=>
do not
include the surrogate code points.
--

spec/index.html
```diff
@@ -495,6 +496,25 @@ <h2>RDF Documents and Syntaxes</h2>
     and still the subject of group discussion. An alternative to conformance
     levels, "profiles", may be adopted instead, abandoned, or described in 
     another specification.</p>
+
+  <section id="rdf-strings">
+    <h2>Strings in RDF</h2>
+
+    <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
+      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>
```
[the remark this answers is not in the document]
gkellogg: One definition, but another reference. Note that the fragment is `#dfn-rdf-string` in any case.

```suggestion
      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
      or <a data-lt="string">RDF string</a>,
```

spec/index.html
```diff
@@ -674,8 +694,9 @@ <h2>Literals</h2>
       elements:</p>
 
     <ul>
-      <li>a <dfn>lexical form</dfn>, being a Unicode [[!UNICODE]] string,
-        which SHOULD be in Normal Form&nbsp;C [[!NFC]],</li>
+      <li>a <dfn>lexical form</dfn> consisting of a sequence of
+        <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
+        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, and therefore do not contain Unicode surrogate code points.</li>
```
gkellogg, line 699:
```suggestion
        which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>, and therefore do not contain <a data-cite="i18n-glossary#dfn-surrogate"  class="lint-ignore">Unicode surrogate code points</a>.</li>
```

spec/index.html
```diff
@@ -503,17 +505,23 @@ <h2>Strings in RDF</h2>
     <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
       or <a data-lt="string">RDF string</a>,
-      `is used to describe an ordered` sequence of zero or more <a data-lt="code point" class="lint-ignore">Unicode code points</a> which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>. Unicode scalar values do not include the <a data-cite="i18n-glossary#dfn-surrogate"  class="lint-ignore">surrogate code points</a>.
+      `is used to describe an ordered` sequence of zero or more
```
TallTed, line 508: Why is this phrase styled as code?
```suggestion
      is used to describe an ordered sequence of zero or more
```

gkellogg: I don't think the style was intended; I'm not sure where it was introduced.

spec/index.html
```diff
@@ -495,6 +498,31 @@ <h2>RDF Documents and Syntaxes</h2>
...
+      or <a data-lt="string">RDF string</a>,
+      is used to describe an ordered sequence of zero or more
+      <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
+      Unicode scalar values do not include the
+      <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
+      Note that most <a>concrete RDF syntaxes</a> require the use
+      of the UTF-8 character encoding [[!RFC3629]], 
+      which requires that certain non-character values be expressed
+      using the `\u0000` or `\U00000000` forms.
```
aphillips, line 516: the UTF-8 character encoding doesn't define escape sequeneces. These would be JS escapes. Other formats use a different format, such as HTML/XML's NCRs (`&#x20ac;`).

aphillips (changes requested):
This is almost done. One minor comment about escapes.

spec/index.html
afs: I think it is OK because it's covered by the text "Note that most concrete RDF syntaxes" at the start of the sentence.
The RDF formats Turtle, TriG, N-triples and N-quads, and SPARQL, all have `\u0000` and `\U00000000`.

aphillips: Fair enough, although the sentence actually says that UTF-8 requires this ;-).

gkellogg, line 516: This might clarify that the UCHAR forms are part of the concrete RDF syntaxes, and not UTF-8.

```suggestion
      Note that most <a>concrete RDF syntaxes</a> require the use
      of the UTF-8 character encoding [[!RFC3629]], 
      and use the `\u0000` or `\U00000000` forms to express certain non-character values.

File: spec/index.html [hunk 28]
```diff
@@ -659,8 +687,7 @@ <h3>IRIs</h3>
           “<code>%3f</code>”)</li>
         <li>Punycode-encoding of Internationalized Domain Names
           in IRIs</li>
-        <li>IRIs that are not in Unicode Normalization
-          Form C [[NFC]]</li>
+        <li>IRIs that are not in Unicode <a class="lint-ignore">Normalization Form C</a> [[UNICODE]]</li>
```
woutermont, line 690: Might immediately refer to the specific annex here.

```suggestion
        <li>IRIs that are not in Unicode <a class="lint-ignore">Normalization Form C</a> [[UAX15]]</li>
```

spec/index.html
aphillips: One nit: some of the "non-character values" are not non-characters. For example, NULL (U+0000) is a real Unicode character. I might not change the text here, though, because the characters that require escapes are non-XML characters and many of these are _also_ non-character code points. (Also, the escapes can be used for "normal" characters and can be especially helpful for e.g. invisible characters such as the bidi controls)

@pfps I read the above proposed change as two useful things: most syntaxes require UTF-8 (which is a fact), thus eliminating non-Unicode character encoding forms. And most syntaxes use escapes to express the above mentioned code points.

spec/index.html
```diff
@@ -659,8 +687,7 @@ <h3>IRIs</h3>
           “<code>%3f</code>”)</li>
         <li>Punycode-encoding of Internationalized Domain Names
```
domel, line 688: add [[RFC3492]]

spec/index.html
```diff
@@ -659,8 +687,7 @@ <h3>IRIs</h3>
...
           in IRIs</li>
```
gkellogg, line 689:
```suggestion
          in IRIs [[RFC3492]]</li>
```

spec/index.html
aphillips: You might consider [[RFC5891]] instead, as that's listed as obsoleting (or rather, updating) 3492. Note that the punycode spec itself is incomplete for encoding domain names: it also requires stuff like 3491 (nameprep). 

I will ask @kleinsin (who is a member of our WG *and* the author of 5891) what the most appropriate current reference is. I don't think there's a handy stable shorthand like BCP47 or STDxx for this yet, but I may be wrong.

domel: I think that we refer to Punycode algorithm here and RFC5891 do not update it.

pchampin:
This was discussed during the TPAC 2023 meeting:
https://www.w3.org/2023/09/12-rdf-star-minutes.html#t03

spec/index.html
domel: @gkellogg Yes, I can. But the question is do we want it? RFC5891 do _not_  obsolete RFC5891. It just updates it.

klensin: > You might consider [[RFC5891]] instead, as that's listed as obsoleting (or rather, updating) 3492. Note that the punycode spec itself is incomplete for encoding domain names: it also requires stuff like 3491 (nameprep).
> 
> I will ask @kleinsin (who is a member of our WG _and_ the author of 5891) what the most appropriate current reference is. I don't think there's a handy stable shorthand like BCP47 or STDxx for this yet, but I may be wrong.

The real issue is that this is probably not just about updating references.  With the warning that I have not studied this thread enough to be confident that I fully understand the issues involved, let me try to explain.

If the discussion is about internatiionalized domain names (IDNs), there should be no references to  RFC 3490 or 3491 (aka "IDNA2003") at all.  Such references should be replaced by references to one or more of RFC 5890ff and the surrounding text should be checked carefully to make sure it is not dependent on assumptions or conventions of IDNA2003 (RFC 3490/3491) that were changed in IDNA2008 (RFC 5890ff),

Now, as far as Punycode is concerned, IDNA2008 generally, and 5891 in particular, do not change the algorithm itself _at all_ (see RFC 5891 Section 4.4).  What it changes are the references inside 3942 that point to IDNA2003 that are updatedto point to IDNA2008.  However, depending on how you are using "punycode" terminology, that is fairly important.  The issues goes back to the core differences between IDNA2003 and IDNA2008.   Conceptually, IDNA2008 is not a tuned and updated version of IDNA2003.  It is as much of a change as the change of the default CCS for HTML from ISO 8859-1 to Unicode.  In the particular case of IDNs, one of those conceptual differences is that IDNA2003 was quite permissive as to what names were valid, even allowing different Unicode strings to convert to the same string after Punycode conversion although (obviously) the reverse conversion could only produce one of those strings.  IDNA2008 is much more focused on identifier integrity and restrictions intended to reduce or eliminate ambiguity, mapping problems, etc.  As a result, the "to_asciii" and "to_unicode" terminology and, if one needs to be precise, "Punycode string" terminology, are no longer useful and, IIR, do not appear in the IDNA2008 documents at all.   And strings that can be generated by the Punycode algorithm that are not valid IDNA2008 labels are, to use a technical term, garbage: there is no way to talk about their validity.  So, in general, contemporary documents should not be referring to Punycode at all, especially as a type of string, but should be using the IDNA2008 terminology of "A-labels" and "U-labels", defined in RFC 5890.

I hope that helps.
  john

Issue #14, opened 2023-02-16
Add the rdf:JSON datatype


domel:
The rdf:JSON datatype is originally defined in [Section 10.2 The rdf:JSON Datatype](https://w3.org/TR/JSON-LD11/#the-rdf-json-datatype) in [JSON-LD 1.1](https://w3.org/TR/JSON-LD11/). This issue is to determine whether to promote that definition to RDF Concepts (and RDF Schema), which would bring it in line with the definition of other datatypes in the RDF namespace.

-----
Original text:
I think that we should "legalize" `rdf:JSON rdf:type rdfs:Datatype .` in the same manner as `rdf:HTML`.

See also:
https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype

pfps:
If there is going to be some official status for rdf:JSON then there should also be the same status for rdf:YAML.

gkellogg:
I don’t think a canonical form of YAML has been specified, as there has for JSON.

pfps:
Where does canonical form fit in here?

domel:
I believe it's about defining [lexical space, value space, and lexical-to-value mapping](https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype).

domel:
I'm not against `rdf:YAML`, but `rdf:JSON` has already been included in RDF Schema (only information in the specification is missing). We can create a separate issue for YAML.

gkellogg:
As [defined in JSON-LD](https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype) it becomes an issue for the L2V mapping. JCS had not reached a normatively CI table state at the time of production, but due to variations in representation it is important for determining equivalent values. YAML has much more potential for variation, but has no similar scheme for normalizing representations.

pfps:
I don't see rdf:JSON in https://www.w3.org/TR/rdf-schema/

pfps:
So maybe the datatype defined in JSON-LD needs a string canonical form.  But that's only an issue for that datatype.  Moreover, having the canonical form being a string when the meaning of the value is something else is not a good idea.

gkellogg:
> I don't see rdf:JSON in https://www.w3.org/TR/rdf-schema/

[...]

PR #62, opened 2023-09-13
Moves datatype definitions to an appendix and adds rdf:JSON datatype

hartig:
[As agreed during the FSF meeting at TPAC](https://www.w3.org/2023/09/12-rdf-star-minutes.html#t07), this PR
1. adds a new appendix,
2. moves the sections that define the `rdf:HTML` and `rdf:XMLLiteral` datatypes into that appendix, and
3. copies the definition of the `rdf:JSON` datatype from [Section 10.2 of [JSON_LD11]](https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype) into this new appendix as well.

Regarding the latter, I have indeed just copied the definition of the `rdf:JSON` datatype. When doing so, there were two things that I was wondering:
1. The sentence that defines the lexical space contains two different links for "JSON Grammar". The first one points to Sec.2 of RFC4627, and the second one points to Sec.2 of RFC8259. I was wondering, why these two different links?
4. There is an issue box about the JSON Canonicalization Scheme (JCS) which I also copied (note, in this copy I had to make a slight change because JSON-LD11 explicitly defines [JSON Literal](https://www.w3.org/TR/json-ld11/#dfn-json-literal), which we don't have in RDF Concepts). However, I have to admit that I do not fully understand the last sentence in this issue box ("Users are cautioned from depending on the [JSON literal](https://www.w3.org/TR/json-ld11/#dfn-json-literal) lexical representation as an [RDF literal](https://www.w3.org/TR/rdf11-concepts/#dfn-literal), as the specifics of serialization may change in a future revision of this document."). In particular, I wonder:
    1. What "serialization" does the last part of his sentence refer to?
    2. What is meant by "this document"? (the JSON-LD11 spec or the JCS spec?)
    5. Also, is this whole comment something specific to JSON-LD?

([GitHack preview](https://raw.githack.com/w3c/rdf-concepts/Issue14/spec/index.html))

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/62.html" title="Last updated on Sep 21, 2023, 6:38 PM UTC (40c1c01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/62/fe28411...40c1c01.html" title="Last updated on Sep 21, 2023, 6:38 PM UTC (40c1c01)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1498,7 +1498,7 @@
         in order to avoid side effects with existing specifications.
       </dd>
 
-      <dt id="JSON-mapping>The <a>lexical-to-value mapping</a></dt>
+      <dt id="JSON-mapping">The <a>lexical-to-value mapping</a></dt>
       <dd>
         maps every element of the lexical space to the result of
         <ol>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1249,7 +1249,7 @@
     The semantics of fragment identifiers is
     <a data-cite="rfc3986#section-3.5">defined in
     RFC 3986</a> [[RFC3986]]: They identify a secondary resource
-    that is usually a part of, view of, defined in, or described in
+    that is usually a part of, a view of, defined in, or described in
     the primary resource, and the precise semantics depend on the set
     of representations that might result from a retrieval action
     on the primary resource.</p>
@@ -1257,11 +1257,11 @@
   <p>This section discusses the handling of fragment identifiers
     in representations that encode <a>RDF graphs</a>.</p>
 
-  <p>In RDF-bearing representations of a primary resource
-    <code>&lt;foo&gt;</code>,
-    the secondary resource identified by a fragment <code>bar</code>
+  <p>In RDF-bearing representations of a primary resource,
+    <code>&lt;https://example.com/foo&gt;</code>,
+    the secondary resource identified by a fragment identifier, <code>bar</code>,
     is the <a>resource</a> <a>denoted</a> by the
-    full <a>IRI</a> <code>&lt;foo#bar&gt;</code> in the <a>RDF graph</a>.
+    full <a>IRI</a> <code>&lt;https://example.com/foo#bar&gt;</code> in the <a>RDF graph</a>.
     Since IRIs in RDF graphs can denote anything, this can be
     something external to the representation, or even external
     to the web.</p>
@@ -1274,7 +1274,7 @@
     <a>fragment identifiers</a> in RDF-bearing representations, the encoded
     <a>RDF graph</a> should use fragment identifiers in a way that is consistent
     with these constraints. For example, in an HTML+RDFa document [[HTML-RDFA]],
-    the fragment <code>chapter1</code> may identify a document section
+    a fragment identifier such as <code>chapter1</code> may identify a document section
     via the semantics of HTML's <code>@name</code> or <code>@id</code>
     attributes. The <a>IRI</a> <code>&lt;#chapter1&gt;</code> should
     then be taken to <a>denote</a> that same section in any RDFa-encoded
@@ -1282,7 +1282,7 @@
     Similarly, fragment identifiers should be used consistently in resources
     with multiple representations that are made available via
     <a data-cite="webarch/#frag-coneg">content negotiation</a>
-    [[WEBARCH]]. For example, if the fragment <code>chapter1</code> identifies a
+    [[WEBARCH]]. For example, if the fragment identifier <code>chapter1</code> identifies a
     document section in an HTML representation of the primary resource, then the
     <a>IRI</a> <code>&lt;#chapter1&gt;</code> should be taken to
     <a>denote</a> that same section in all RDF-bearing representations of the
@@ -1293,19 +1293,19 @@
   <h2>Generalized RDF Triples, Graphs, and Datasets</h2>
 
   <p>It is sometimes convenient to loosen the requirements
-    on <a>RDF triple</a>s.  For example, the completeness
+    on <a>RDF triples</a>.  For example, the completeness
     of the RDFS entailment rules is easier to show with a
     generalization of RDF triples.</p>
 
   <p>A <dfn class="export">generalized RDF triple</dfn> is a triple having a subject,
-    a predicate, and object, where each can be an <a>IRI</a>, a
+    a predicate, and an object, where each can be an <a>IRI</a>, a
     <a>blank node</a> or a
     <a>literal</a>. A
     <dfn class="export">generalized RDF graph</dfn>
     is a set of generalized RDF triples. A
     <dfn class="export">generalized RDF dataset</dfn>
     comprises a distinguished generalized RDF graph, and zero
-    or more pairs each associating an IRI, a blank node or a literal
+    or more pairs each associating an IRI, a blank node, or a literal
     to a generalized RDF graph.</p>
 
 
@@ -1314,14 +1314,14 @@
     <a data-lt="RDF graph">graphs</a>, and
     <a data-lt="RDF dataset">datasets</a> only
     by allowing <a>IRIs</a>,
-    <a>blank nodes</a> and
+    <a>blank nodes</a>, and
     <a>literals</a> to appear
-    in any position, i.e., as subject, predicate, object or graph names.</p>
+    in any position, i.e., as subject, predicate, object, or graph name.</p>
 
-  <p class="note" id="note-generalized-rdf"> Any users of
-    generalized RDF triples, graphs or datasets need to be
+  <p class="note" id="note-generalized-rdf">Any user of
+    generalized RDF triples, graphs, or datasets needs to be
     aware that these notions are non-standard extensions of
-    RDF and their use may cause interoperability problems.
+    RDF, and their use may cause interoperability problems.
     There is no requirement on the part of any RDF tool to
     accept, process, or produce anything beyond standard RDF
     triples, graphs, and datasets. </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -33,7 +33,7 @@
         { name: "Brian McBride" }
       ],
 
-      xref: ["i18n-glossary"],
+      xref: ["i18n-glossary", "infra"],
       github: "https://github.com/w3c/rdf-concepts/",
       group:           "rdf-star",
       doJsonLd:     true,
@@ -140,7 +140,7 @@
     <dfn data-lt="resource">resources</dfn>. Anything can be a resource,
     including physical things, documents, abstract concepts, numbers
     and strings; the term is synonymous with "entity" as it is used in
-    the RDF Semantics specification [[RDF12-SEMANTICS]].
+    [[[RDF12-SEMANTICS]]] [[RDF12-SEMANTICS]].
     The resource denoted by an IRI is called its <a>referent</a>, and the
     resource denoted by a literal is called its
     <a>literal value</a>. Literals have
@@ -407,7 +407,7 @@
       simple <dfn class="export">logical expression</dfn>, or claim about the world.
       An <a>RDF graph</a> is the conjunction (logical <em>AND</em>) of
       its triples.   The precise details of this meaning of RDF triples and graphs are
-      the subject of the RDF Semantics specification [[RDF12-SEMANTICS]], which yields the
+      the subject of [[[RDF12-SEMANTICS]]] [[RDF12-SEMANTICS]], which yields the
       following relationships between <a>RDF graph</a>s:</p>
 
     <dl>
@@ -506,7 +506,7 @@
       Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
       or <a data-lt="string">RDF string</a>,
       is used to describe an ordered sequence of zero or more
-      <a data-lt="code point" class="lint-ignore">Unicode code points</a>
+      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>
       which are <a data-cite="i18n-glossary#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
       Unicode scalar values do not include the
       <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
@@ -520,7 +520,7 @@
       <a data-cite="i18n-glossary#dfn-code-unit">code units</a> of two strings
       that use the same <a data-cite="i18n-glossary#dfn-character-encoding">Unicode character encoding</a>
       (UTF-8 or UTF-16) without decoding the string into a
-      <a data-lt="code point" class="lint-ignore">Unicode code point</a> sequence.</p>
+      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code point</a> sequence.</p>
   </section>
 </section>
 
@@ -615,7 +615,7 @@
 
     <p><dfn>IRI equality</dfn>:
       Two IRIs are the same if and only if they consist of the same sequence of
-      <a data-lt="code point" class="lint-ignore">Unicode code points</a>,
+      <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>,
       as in Simple String Comparison in
       <a data-cite="rfc3987#section-5.3.1">section 5.3.1</a> of [[!RFC3987]].
       (This is done in the abstract syntax, so the IRIs are resolved
@@ -701,7 +701,7 @@
 
     <ul>
       <li>a <dfn>lexical form</dfn> consisting of a sequence of
-        <a data-lt="code point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
+        <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
         which are <a data-cite="i18n-glossary#dfn-scalar-value">Unicode scalar values</a>,
         and therefore do not contain
         <a data-cite="i18n-glossary#dfn-surrogate" class="lint-ignore">Unicode surrogate code points</a>.</li>
@@ -1002,7 +1002,7 @@
     [[!XMLSCHEMA11-2]]. Any datatype definition that conforms
     to this abstraction MAY be used in RDF, even if not defined
     in terms of XML Schema. RDF re-uses many of the XML Schema
-    built-in datatypes, and defines two additional non-normative datatypes,
+    built-in datatypes, and defines three additional datatypes,
     <code><a>rdf:JSON</a></code>, <code><a>rdf:HTML</a></code>, and <code><a>rdf:XMLLiteral</a></code>.
     The list of datatypes supported by an implementation is determined
     by its <a>recognized datatype IRIs</a>.</p>
@@ -1086,7 +1086,7 @@
       listed in the following table are the
       <dfn>RDF-compatible XSD types</dfn>. Their use is RECOMMENDED.</p>
 
-    <p>Readers might note that the xsd:hexBinary and xsd:base64Binary
+    <p>Readers might note that the `xsd:hexBinary` and `xsd:base64Binary`
       datatypes are the only safe datatypes for transferring binary
       information.</p>
 
@@ -1223,8 +1223,8 @@
 
     <p class="note">Semantic extensions of RDF might choose to
       recognize other datatype IRIs
-      and require them to refer to a fixed datatype.  See the RDF
-      Semantics specification [[RDF12-SEMANTICS]] for more information on
+      and require them to refer to a fixed datatype.
+      See [[[RDF12-SEMANTICS]]] [[RDF12-SEMANTICS]] for more information on
       semantic extensions.</p>
 
     <p class="note" id="note-custom-datatypes">The Web Ontology Language
@@ -1460,60 +1460,56 @@
       <dd>is <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#JSON</code>.</dd>
 
       <dt id="JSON-lexical-space">The <a>lexical space</a></dt>
-      <dd>is the set of all UNICODE [[UNICODE]] strings that conform to the
+      <dd>is the set of all <a>strings</a> that conform to the
         <a data-cite="RFC4627#section-2">JSON Grammar</a> as described in
         <a data-cite="RFC8259#section-2">Section&nbsp;2 JSON Grammar</a> of
         [[RFC8259]].</dd>
 
       <dt id="JSON-value-space">The <a>value space</a></dt>
       <dd>
-        is the set of all UNICODE [[UNICODE]] strings that conform to the
-        <a data-cite="RFC4627#section-2">JSON Grammar</a> as described in
-        <a data-cite="RFC8259#section-2">Section&nbsp;2 JSON Grammar</a> of
-        [[RFC8259]], and furthermore comply with the following constraints:
+        is the space defined by
+        the <a data-cite="JSON-LD11#dfn-internal-representation">JSON-LD internal representation</a> [[JSON-LD11]],
+        being the <a data-cite="RFC4627#section-3">JSON values</a> <a data-lt="list">array</a>,
+        <a>map</a>,
+        <a>string</a>,
+        <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+        <a>boolean</a>, or
+        <a data-cite="INFRA#nulls">null</a>.
         <ul>
-          <li>They MUST NOT contain any unnecessary whitespace.</li>
-          <li>Keys in objects MUST be ordered lexicographically.</li>
-          <li>Native Numeric values MUST be serialized according to
-            <a data-cite="ECMASCRIPT#sec-tostring-applied-to-the-number-type">Section&nbsp;7.1.12.1</a>
-            of [[ECMASCRIPT]].</li>
-          <li>Strings MUST be serialized with Unicode codepoints from `U+0000`
-            through `U+001F` using lower case hexadecimal Unicode notation
-            (`\uhhhh`) except for the set of predefined JSON control characters —
-            `U+0008`, `U+0009`, `U+000A`, `U+000C`, and `U+000D` — which SHOULD be
-            serialized as `\b`, `\t`, `\n`, `\f`, and `\r`, respectively. All
-            other Unicode characters SHOULD be serialized "as is", except
-            `U+005C` (`\`) and `U+0022` (`"`), which SHOULD be serialized as
-            `\\` and `\"`, respectively.</li>
+          <li><a data-lt="list">Array</a> <a data-cite="INFRA#list-item">entries</a> may be
+            <a>map</a>,
+            <a>string</a>,
+            <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+            <a>boolean</a>, or
+            <a data-cite="INFRA#nulls">null</a>.</li>
+          <li><a>Map</a> <a data-cite="INFRA#map-key">keys</a> are <a>strings</a> with
+            <a data-cite="INFRA#map-value">values</a> may be any of the above
+            <a data-cite="RFC4627#section-3">JSON values</a>.</li>
         </ul>
-        <div class="issue">The JSON Canonicalization Scheme (JCS) [[RFC8785]]
-          is an emerging standard for JSON canonicalization. This <code>rdf:JSON</code> specification
-          will likely be updated to require such a canonical representation.
-          Users are cautioned against depending on the lexical representation of
-          literals with the <code>rdf:JSON</code> datatype as RDF literals,
-          as the specifics of serialization may change in a future revision of
-          this document.</div>
-        Despite being defined as a set of strings, this value space is
-        considered distinct from the value space of <code>xsd:string</code>,
-        in order to avoid side effects with existing specifications.
+
+        Two <a data-cite="RFC4627#section-3">JSON values</a> are considered equal
+        if they are the same if they are the same <a>string</a>,
+        <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+        <a>boolean</a>, or
+        <a data-cite="INFRA#nulls">null</a>,
+        if they are both an <a data-lt="list">array</a> with equal <a data-cite="INFRA#list-item">entries</a>, or
+        if they are both a <a>map</a> with equal <a data-cite="INFRA#map-entry">map entries</a>.
       </dd>
 
       <dt id="JSON-mapping">The <a>lexical-to-value mapping</a></dt>
       <dd>
-        maps every element of the lexical space to the result of
-        <ol>
-          <li>parsing it into an internal representation consistent with the
-            [[ECMASCRIPT]] representation created by using the <code>JSON.parse</code>
-            function as defined in <a data-cite="ECMASCRIPT#sec-json-object">Section&nbsp;24.5 The JSON Object</a>
-            of [[ECMASCRIPT]],</li>
-          <li>then serializing it into the JSON format [[RFC8259]] in compliance
-            with the constraints of the value space described above.</li>
-        </ol>
+        maps every element of the lexical space to the result of parsing it into an
+        <a data-cite="JSON-LD11#dfn-internal-representation">internal representation</a>
+        consistent with the [[ECMASCRIPT]] representation created by using the <code>JSON.parse</code>
+        function as defined in <a data-cite="ECMASCRIPT#sec-json-object">Section&nbsp;24.5 The JSON Object</a>
+        of [[ECMASCRIPT]].
       </dd>
 
       <dt id="JSON-canonical">The canonical mapping</dt>
-      <dd>maps any element of the value space to the identical string in the
-        lexical space.</dd>
+      <dd>maps the <a data-cite="JSON-LD11#dfn-internal-representation">internal representation</a>
+        to a <a>string</a> conforming to the <a data-cite="RFC4627#section-2">JSON Grammar</a>
+        using the mechanism described in [[[RFC8785]] [[RFC8785]].
+      </dd>
     </dl>
   </section>
 
@@ -1758,7 +1754,7 @@
       from <a data-cite="?JSON-LD11#the-rdf-json-datatype">Section&nbsp;10.2 The `rdf:JSON` Datatype</a>
       in [[?JSON-LD11]].</li>
     <li>Clarify Unicode terminology,
-      using <a data-lt="code point" class="lint-ignore">Unicode code points</a>,
+      using <a data-cite="i18n-glossary#dfn-code-point" class="lint-ignore">Unicode code points</a>,
       and restriction to the XML <a data-cite="XML11#charsets">Char</a> production.
       Also removes obsolete recommendations for the use of Normalization Form C in literals.
       Adds a definition of <a>string</a> that can be used in other RDF documents.</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1199,7 +1199,8 @@
       MUST refer to the RDF-compatible XSD type named <code>xsd:xxx</code> for
       every XSD type listed in <a href="#xsd-datatypes">section 5.1</a>.</p>
 
-    <p>The following three datatype IRIs are defined in Appendix&nbsp;<a href="#section-additional-datatypes" class="sectionRef"></a>:</p>
+    <p>The datatypes identified by the three IRIs below are defined in
+    Appendix&nbsp;<a href="#section-additional-datatypes" class="sectionRef"></a>:</p>
 
     <ul>
       <li>The IRI <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral</code>
@@ -1257,11 +1258,11 @@
   <p>This section discusses the handling of fragment identifiers
     in representations that encode <a>RDF graphs</a>.</p>
 
-  <p>In RDF-bearing representations of a primary resource,
+  <p>In RDF-bearing representations of a primary resource, e.g.,
     <code>&lt;https://example.com/foo&gt;</code>,
-    the secondary resource identified by a fragment identifier, <code>bar</code>,
+    the secondary resource identified by a fragment identifier, e.g., <code>bar</code>,
     is the <a>resource</a> <a>denoted</a> by the
-    full <a>IRI</a> <code>&lt;https://example.com/foo#bar&gt;</code> in the <a>RDF graph</a>.
+    full <a>IRI</a> in the <a>RDF graph</a>, which would be <code>&lt;https://example.com/foo#bar&gt;</code> in this case.
     Since IRIs in RDF graphs can denote anything, this can be
     something external to the representation, or even external
     to the web.</p>
@@ -1276,7 +1277,7 @@
     with these constraints. For example, in an HTML+RDFa document [[HTML-RDFA]],
     a fragment identifier such as <code>chapter1</code> may identify a document section
     via the semantics of HTML's <code>@name</code> or <code>@id</code>
-    attributes. The <a>IRI</a> <code>&lt;#chapter1&gt;</code> should
+    attributes. Such an <a>IRI</a>, e.g., <code>&lt;#chapter1&gt;</code>, should
     then be taken to <a>denote</a> that same section in any RDFa-encoded
     <a>triples</a> within the same document.
     Similarly, fragment identifiers should be used consistently in resources
@@ -1322,7 +1323,7 @@
     generalized RDF triples, graphs, or datasets needs to be
     aware that these notions are non-standard extensions of
     RDF, and their use may cause interoperability problems.
-    There is no requirement on the part of any RDF tool to
+    There is no requirement for any RDF tool to
     accept, process, or produce anything beyond standard RDF
     triples, graphs, and datasets. </p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1462,14 +1462,14 @@
 
       <dt id="JSON-lexical-space">The <a>lexical space</a></dt>
       <dd>is the set of all <a data-lt="string">RDF strings</a> that conform to the
-        <a data-cite="RFC4627#section-2">JSON Grammar</a> as described in
+        <a data-cite="RFC8259#section-2">JSON Grammar</a> as described in
         <a data-cite="RFC8259#section-2">Section&nbsp;2 JSON Grammar</a> of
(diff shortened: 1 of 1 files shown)


domel:
> The sentence that defines the lexical space contains two different links for "JSON Grammar". The first one points to Sec.2 of RFC4627, and the second one points to Sec.2 of RFC8259. I was wondering, why these two different links?

I think that's an oversight. RFC8259 obsoleted RFC7159 obsoleted RFC4627. So I see no reason to refer to the old document when the new one is valid.

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+  <h2>Fragment Identifiers</h2>
+
+  <p>RDF uses <a>IRIs</a>, which may include
+    <span id="dfn-fragment-identifiers"><!-- obsolete term--></span>
+    <dfn data-lt="fragment identifier" class="no-export">fragment identifiers</dfn>,
+    as resource identifiers.
+    The semantics of fragment identifiers is
+    <a data-cite="rfc3986#section-3.5">defined in
+    RFC 3986</a> [[RFC3986]]: They identify a secondary resource
+    that is usually a part of, view of, defined in, or described in
```
TallTed, line 1252:
```suggestion
    that is usually a part of, a view of, defined in, or described in
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    the primary resource, and the precise semantics depend on the set
+    of representations that might result from a retrieval action
+    on the primary resource.</p>
+
+  <p>This section discusses the handling of fragment identifiers
+    in representations that encode <a>RDF graphs</a>.</p>
+
+  <p>In RDF-bearing representations of a primary resource
+    <code>&lt;foo&gt;</code>,
+    the secondary resource identified by a fragment <code>bar</code>
```
TallTed, line 1262:
```suggestion
    the secondary resource identified by a fragment identifier, <code>bar</code>,
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    RFC 3986</a> [[RFC3986]]: They identify a secondary resource
+    that is usually a part of, view of, defined in, or described in
+    the primary resource, and the precise semantics depend on the set
+    of representations that might result from a retrieval action
+    on the primary resource.</p>
+
+  <p>This section discusses the handling of fragment identifiers
+    in representations that encode <a>RDF graphs</a>.</p>
+
+  <p>In RDF-bearing representations of a primary resource
```
TallTed, line 1260:
```suggestion
  <p>In RDF-bearing representations of a primary resource,
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    <code>&lt;foo&gt;</code>,
```
TallTed, line 1261:
```suggestion
    <code>&lt;https://example.com/foo&gt;</code>,
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    the secondary resource identified by a fragment <code>bar</code>
+    is the <a>resource</a> <a>denoted</a> by the
+    full <a>IRI</a> <code>&lt;foo#bar&gt;</code> in the <a>RDF graph</a>.
```
TallTed, line 1264:
```suggestion
    full <a>IRI</a> <code>&lt;https://example.com/foo#bar&gt;</code> in the <a>RDF graph</a>.
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+
+  <p>In this way, the RDF-bearing representation acts as an intermediary
+    between the web-accessible primary resource, and some set of possibly
+    non-web or abstract entities that the <a>RDF graph</a> may describe.</p>
+
+  <p>In cases where other specifications constrain the semantics of
+    <a>fragment identifiers</a> in RDF-bearing representations, the encoded
+    <a>RDF graph</a> should use fragment identifiers in a way that is consistent
+    with these constraints. For example, in an HTML+RDFa document [[HTML-RDFA]],
+    the fragment <code>chapter1</code> may identify a document section
```
TallTed, line 1277:
```suggestion
    a fragment identifier, <code>chapter1</code>, may identify a document section
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    via the semantics of HTML's <code>@name</code> or <code>@id</code>
+    attributes. The <a>IRI</a> <code>&lt;#chapter1&gt;</code> should
```
TallTed, line 1279:
```suggestion
    attributes. The <a>IRI</a>, <code>&lt;#chapter1&gt;</code>, should
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    then be taken to <a>denote</a> that same section in any RDFa-encoded
+    <a>triples</a> within the same document.
+    Similarly, fragment identifiers should be used consistently in resources
+    with multiple representations that are made available via
+    <a data-cite="webarch/#frag-coneg">content negotiation</a>
+    [[WEBARCH]]. For example, if the fragment <code>chapter1</code> identifies a
+    document section in an HTML representation of the primary resource, then the
+    <a>IRI</a> <code>&lt;#chapter1&gt;</code> should be taken to
```
TallTed, line 1287:
```suggestion
    [[WEBARCH]]. For example, if the fragment identifier, <code>chapter1</code>, identifies a
    document section in an HTML representation of the primary resource, then the
    <a>IRI</a>, <code>&lt;#chapter1&gt;</code>, should be taken to
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    <a>denote</a> that same section in all RDF-bearing representations of the
+    same primary resource.</p>
+</section>
+
+<section id="section-generalized-rdf" class="informative">
+  <h2>Generalized RDF Triples, Graphs, and Datasets</h2>
+
+  <p>It is sometimes convenient to loosen the requirements
+    on <a>RDF triple</a>s.  For example, the completeness
```
TallTed, line 1296:
```suggestion
    on <a>RDF triples</a>.  For example, the completeness
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    of the RDFS entailment rules is easier to show with a
+    generalization of RDF triples.</p>
+
+  <p>A <dfn class="export">generalized RDF triple</dfn> is a triple having a subject,
+    a predicate, and object, where each can be an <a>IRI</a>, a
```
TallTed, line 1301:
```suggestion
    a predicate, and an object, where each can be an <a>IRI</a>, a
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    <a>blank node</a> or a
+    <a>literal</a>. A
+    <dfn class="export">generalized RDF graph</dfn>
+    is a set of generalized RDF triples. A
+    <dfn class="export">generalized RDF dataset</dfn>
+    comprises a distinguished generalized RDF graph, and zero
+    or more pairs each associating an IRI, a blank node or a literal
```
TallTed, line 1308:
```suggestion
    or more pairs each associating an IRI, a blank node, or a literal
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    to a generalized RDF graph.</p>
+
+
+  <p>Generalized RDF triples, graphs, and datasets differ
+    from normative RDF <a>triples</a>,
+    <a data-lt="RDF graph">graphs</a>, and
+    <a data-lt="RDF dataset">datasets</a> only
+    by allowing <a>IRIs</a>,
+    <a>blank nodes</a> and
```
TallTed, line 1317:
```suggestion
    <a>blank nodes</a>, and
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    <a>literals</a> to appear
+    in any position, i.e., as subject, predicate, object or graph names.</p>
```
TallTed, line 1319:
```suggestion
    in any position, i.e., as subject, predicate, object, or graph name.</p>
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+
+  <p class="note" id="note-generalized-rdf"> Any users of
+    generalized RDF triples, graphs or datasets need to be
+    aware that these notions are non-standard extensions of
+    RDF and their use may cause interoperability problems.
+    There is no requirement on the part of any RDF tool to
+    accept, process, or produce anything beyond standard RDF
+    triples, graphs, and datasets. </p>
```
TallTed, line 1327:
```suggestion
  <p class="note" id="note-generalized-rdf">Any user of
    generalized RDF triples, graphs, or datasets needs to be
    aware that these notions are non-standard extensions of
    RDF, and their use may cause interoperability problems.
    There is no requirement on the part of any RDF tool to
    accept, process, or produce anything beyond standard RDF
    triples, graphs, and datasets. </p>
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
-    the secondary resource identified by a fragment <code>bar</code>
-    is the <a>resource</a> <a>denoted</a> by the
-    full <a>IRI</a> <code>&lt;foo#bar&gt;</code> in the <a>RDF graph</a>.
-    Since IRIs in RDF graphs can denote anything, this can be
-    something external to the representation, or even external
-    to the web.</p>
+    <p>RDF provides for JSON content as a possible <a>literal value</a>.
+      This allows markup in literal values. Such content is indicated in an
+      <a>RDF graph</a> using a <a>literal</a> whose <a>datatype</a> is set
+      to <code><dfn>rdf:JSON</dfn></code>.</p>
```
TallTed, line 1454:
```suggestion
    <p>RDF provides for JSON content as a possible <a>literal value</a>.
      This includes allowing markup in literal values. Such content is indicated in an
      <a>RDF graph</a> as a <a>literal</a> whose <a>datatype</a> is set
      to <code><dfn>rdf:JSON</dfn></code>.</p>
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+            <a data-cite="ECMASCRIPT#sec-tostring-applied-to-the-number-type">Section&nbsp;7.1.12.1</a>
+            of [[ECMASCRIPT]],</li>
+          <li>Strings MUST be serialized with Unicode codepoints from `U+0000`
+            through `U+001F` using lower case hexadecimal Unicode notation
+            (`\uhhhh`) unless in the set of predefined JSON control characters
+            `U+0008`, `U+0009`, `U+000A`, `U+000C` or `U+000D` which SHOULD be
+            serialized as `\b`, `\t`, `\n`, `\f` and `\r`, respectively. All
+            other Unicode characters SHOULD be serialized "as is", other than
+            `U+005C` (`\`) and `U+0022` (`"`) which SHOULD be serialized as
+            `\\` and `\"`, respectively.</li>
```
TallTed, line 1487:
```suggestion
          <li>They MUST NOT contain any unnecessary whitespace.</li>
          <li>Keys in objects MUST be ordered lexicographically.</li>
          <li>Native Numeric values MUST be serialized according to
            <a data-cite="ECMASCRIPT#sec-tostring-applied-to-the-number-type">Section&nbsp;7.1.12.1</a>
            of [[ECMASCRIPT]].</li>
          <li>Strings MUST be serialized with Unicode codepoints from `U+0000`
            through `U+001F` using lower case hexadecimal Unicode notation
            (`\uhhhh`) except for the set of predefined JSON control characters —
            `U+0008`, `U+0009`, `U+000A`, `U+000C`, and `U+000D` — which SHOULD be
            serialized as `\b`, `\t`, `\n`, `\f`, and `\r`, respectively. All
            other Unicode characters SHOULD be serialized "as is", except
            `U+005C` (`\`) and `U+0022` (`"`), which SHOULD be serialized as
            `\\` and `\"`, respectively.</li>
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+        </ul>
+        <div class="issue">The JSON Canonicalization Scheme (JCS) [[RFC8785]]
+          is an emerging standard for JSON canonicalization. This specification
+          will likely be updated to require such a canonical representation.
+          Users are cautioned from depending on the lexical representation of
```
TallTed, line 1492:
```suggestion
          Users are cautioned against depending on the lexical representation of
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+          literals with the <code>rdf:JSON</code> datatype as an RDF literal,
```
TallTed, line 1493:
```suggestion
          literals with the <code>rdf:JSON</code> datatype as RDF literals,
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+            through `U+001F` using lower case hexadecimal Unicode notation
+            (`\uhhhh`) unless in the set of predefined JSON control characters
+            `U+0008`, `U+0009`, `U+000A`, `U+000C` or `U+000D` which SHOULD be
+            serialized as `\b`, `\t`, `\n`, `\f` and `\r`, respectively. All
+            other Unicode characters SHOULD be serialized "as is", other than
+            `U+005C` (`\`) and `U+0022` (`"`) which SHOULD be serialized as
+            `\\` and `\"`, respectively.</li>
+        </ul>
+        <div class="issue">The JSON Canonicalization Scheme (JCS) [[RFC8785]]
+          is an emerging standard for JSON canonicalization. This specification
```
TallTed, line 1490:
```suggestion
          is an emerging standard for JSON canonicalization. This <code>rdf:JSON</code> specification
```

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
-    in any position, i.e., as subject, predicate, object or graph names.</p>
+      <dt>The <a>lexical-to-value mapping</a></dt>
+      <dd>
+        maps every element of the lexical space to the result of
+        <ol>
+          <li>parsing it into an internal representation consistent with the
+            [[ECMASCRIPT]] representation created by using the <code>JSON.parse</code>
+            function as defined in <a data-cite="ECMASCRIPT#sec-json-object">Section&nbsp;24.5 The JSON Object</a>
+            of [[ECMASCRIPT]],</li>
+          <li>then serializing it in the JSON format [[RFC8259]] in compliance
```
TallTed, line 1509:
```suggestion
          <li>then serializing it into the JSON format [[RFC8259]] in compliance
```

hartig:
Looking at @TallTed's change suggestions, I see only now that the diff of this PR contains the complete Section 6 (Fragment Identifiers) and Section 7 (Generalized RDF Triples, Graphs, and Datasets) as new text, and then, later, as removed text snippets. I have not touched these sections at all. It seems that the Github diff functionality did not correctly recognize the actual changes. This is strange in particular because, when I did `git diff` locally before committing the changes, the diff looked correct. Hence, the diff functionality of git recognized the changes correctly, in contrast to the one of Github.

Long story short, at least half of @TallTed's change suggestion in this PR now are for text that was not supposed to be changed at all by this PR. Therefore, I am a bit unsure what to do with these suggestions. My idea is the following. I will merge right away the suggestions for the parts of the text that this PR was meant to change. Regarding the suggestions for the other parts (i.e., in Sections 6 and 7), I will keep them open for a few days and, unless I hear otherwise from any of the other editors, will then merge them later (mid next week).

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
-    comprises a distinguished generalized RDF graph, and zero
-    or more pairs each associating an IRI, a blank node or a literal
-    to a generalized RDF graph.</p>
+      <dt>The <a>lexical space</a></dt>
+      <dd>is the set of all UNICODE [[UNICODE]] strings that conform to the
+        <a data-cite="RFC4627#section-2">JSON Grammar</a> as described in
+        <a data-cite="RFC8259#section-2">Section&nbsp;2 JSON Grammar</a> of
+        [[RFC8259]].</dd>
 
+      <dt id="XMLLiteral-value-space">The <a>value space</a></dt>
```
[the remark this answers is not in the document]
gkellogg: As discussed at TPAC, we can consider updating the definition of the value space based in [INFRA](https://infra.spec.whatwg.org):

> The value space defined by the [JSON-LD internal representation](https://www.w3.org/TR/json-ld11-api/#dfn-internal-representation) of the lexical value, being an [array](https://infra.spec.whatwg.org/#list), [map](https://infra.spec.whatwg.org/#ordered-map), [string](https://infra.spec.whatwg.org/#javascript-string), [number](https://tc39.es/ecma262/#sec-terms-and-definitions-number-value), [booleans](https://infra.spec.whatwg.org/#boolean), or [null](https://infra.spec.whatwg.org/#nulls).
> * [Array](https://infra.spec.whatwg.org/#list) entries may be [map](https://infra.spec.whatwg.org/#ordered-map), [string](https://infra.spec.whatwg.org/#javascript-string), [number](https://tc39.es/ecma262/#sec-terms-and-definitions-number-value), [booleans](https://infra.spec.whatwg.org/#boolean), or [null](https://infra.spec.whatwg.org/#nulls).
> * [Map](https://infra.spec.whatwg.org/#ordered-map) [keys](https://infra.spec.whatwg.org/#map-key) are [strings](https://infra.spec.whatwg.org/#javascript-string).
> * [Map](https://infra.spec.whatwg.org/#ordered-map) [values](https://infra.spec.whatwg.org/#map-value) may be any of the above types.

spec/index.html
```diff
@@ -1350,106 +1445,81 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+      <dd>
+        maps every element of the lexical space to the result of
+        <ol>
+          <li>parsing it into an internal representation consistent with the
+            [[ECMASCRIPT]] representation created by using the <code>JSON.parse</code>
+            function as defined in <a data-cite="ECMASCRIPT#sec-json-object">Section&nbsp;24.5 The JSON Object</a>
+            of [[ECMASCRIPT]],</li>
+          <li>then serializing it into the JSON format [[RFC8259]] in compliance
+            with the constraints of the value space described above.</li>
+        </ol>
```
gkellogg, line 1511:
```suggestion
        parsing the lexical value into the [internal representation](https://www.w3.org/TR/json-ld11-api/#dfn-internal-representation).
```

gkellogg (changes requested):
Numerous small things; I'll follow on with a direct commit giving some of the rdf:HTML elements identifiers.

@TallTed's comments should not be lost, but probably belong in their own PR.

The suggested update to the rdf:JSON value space is a suggestion. It could be based directly on INFRA and ECMAScript (for number), but why repeat the work done in JSON-LD? I can push a more concrete update if agreed.

spec/index.html
```diff
@@ -1350,106 +1445,77 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
-    <dfn class="export">generalized RDF graph</dfn>
-    is a set of generalized RDF triples. A
-    <dfn class="export">generalized RDF dataset</dfn>
-    comprises a distinguished generalized RDF graph, and zero
-    or more pairs each associating an IRI, a blank node or a literal
-    to a generalized RDF graph.</p>
+      <dt id="JSON-value-space">The <a>value space</a></dt>
+      <dd>
+        is the space defined by
+        the <a data-cite="JSON-LD11#dfn-internal-representation">JSON-LD internal representation</a> [[JSON-LD11]],
```
gkellogg, line 1471: This probably works on its own, without reference to the JSON-LD internal representation.

TallTed:
>> [@hartig] at least half of @TallTed's change suggestion in this PR now are for text that was not supposed to be changed at all by this PR

> [@gkellogg] @TallTed's comments should not be lost, but probably belong in their own PR.

I have no objection to those comments being moved to a different PR (with or without an issue, as I don't think those comments are controversial).  It would be super if you could create that PR, as I'm not certain which comments you mean, and I don't want to lose any.

spec/index.html
TallTed: It's not clear to me what "this" is that "probably works on its own". I suggest a change request, to make it clear what you mean.

gkellogg: My suggestion was that we could probably drop the reference to JSON-LD internal representation, as we pretty much describe the INFRA/ECMAScript types we use directly. There was a concern raised at TPAC about layering specs.

gkellogg (approved):

spec/index.html
hartig:
```suggestion
    a fragment identifier such as <code>chapter1</code> may identify a document section
```

spec/index.html
hartig:
```suggestion
    [[WEBARCH]]. For example, if the fragment identifier <code>chapter1</code> identifies a
    document section in an HTML representation of the primary resource, then the
    <a>IRI</a> <code>&lt;#chapter1&gt;</code> should be taken to
```

hartig:
> > > [@hartig] at least half of @TallTed's change suggestion in this PR now are for text that was not supposed to be changed at all by this PR

> > [@gkellogg] @TallTed's comments should not be lost, but probably belong in their own PR.

> [@TallTed] I have no objection to those comments being moved to a different PR (with or without an issue, as I don't think those comments are controversial). It would be super if you could create that PR, as I'm not certain which comments you mean, and I don't want to lose any.

Since all of these small changes have been minor (mostly missing commas, correct use of plural vs singular, etc), I don't think it is worth the effort creating a separate PR and copying them all over. Therefore, I have merged them into this PR now.

spec/index.html
hartig: @TallTed  In this case, I don't think that putting commas is correct.

pfps:
The rdf:JSON datatype being added here appears to be very different from the rdf:JSON datatype in https://www.w3.org/TR/json-ld/#the-rdf-json-datatype

spec/index.html
gkellogg: I'm inclined to agree that the commas don't look right in this case, but I typically defer to @TallTed's judgement. Potentially, there's a way to re-phrase so it seems more natural.

spec/index.html
gkellogg: Looks like the comment is on an old commit; the current text includes "identifier".

hartig:
> The rdf:JSON datatype being added here appears to be very different from the rdf:JSON datatype in https://www.w3.org/TR/json-ld/#the-rdf-json-datatype

I see that too now. Just to clarify: my original changes regarding the rdf:JSON datatype in this PR were only to insert an almost verbatim copy of the text of https://www.w3.org/TR/json-ld/#the-rdf-json-datatype into the rdf:JSON section in this document (rdf-concepts). Yet, it seems that @gkellogg's commits to this PR have edited the section quite a bit.

spec/index.html
```diff
@@ -1350,106 +1445,75 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+      <dt id="JSON-value-space">The <a>value space</a></dt>
+      <dd>
+        is a single <a data-cite="RFC4627#section-3">JSON value</a> in the form of an <a data-lt="list">array</a>,
+        <a>map</a>,
+        <a>string</a>,
+        <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+        <a>boolean</a>, or
+        <a data-cite="INFRA#nulls">null</a>.
+        <ul>
+          <li><a data-lt="list">Array</a> <a data-cite="INFRA#list-item">entries</a> may be
```
afs, line 1477: Why can't arrays contain arrays?

spec/index.html
```diff
@@ -1350,106 +1445,75 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
-    is a set of generalized RDF triples. A
-    <dfn class="export">generalized RDF dataset</dfn>
-    comprises a distinguished generalized RDF graph, and zero
-    or more pairs each associating an IRI, a blank node or a literal
-    to a generalized RDF graph.</p>
+      <dt id="JSON-value-space">The <a>value space</a></dt>
+      <dd>
+        is a single <a data-cite="RFC4627#section-3">JSON value</a> in the form of an <a data-lt="list">array</a>,
+        <a>map</a>,
+        <a>string</a>,
```
afs, line 1472: That links to RDF string - is that right given that the other defs go to in Infra.

An infra string is "A string is a sequence of unsigned 16-bit integers, also known as code units"

pfps:
There is no section 24.5 in the current ECMASCRIPT document.

spec/index.html
hartig: @gkellogg yes, the comment is on an old commit. Notice that there are actually two separate change suggestions here: one from @TallTed and one from me. The one from me is a copy of Ted's suggestion but without the added commas, because also here I think that adding the commas is not correct (exactly as in the other case).

spec/index.html
```diff
@@ -1498,15 +1496,15 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
 
       <dt id="JSON-mapping">The <a>lexical-to-value mapping</a></dt>
       <dd>
-        maps every element of the lexical space to the result of parsing it into an
-        <a data-cite="JSON-LD11#dfn-internal-representation">internal representation</a>
+        maps every element of the lexical space to the result of parsing it into a
+        <a data-cite="RFC4627#section-3">JSON value</a>
```
afs, line 1500: Wrong reference? Section 3 is "Encoding"

spec/index.html
```diff
@@ -1350,106 +1445,75 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+            <a data-cite="INFRA#map-value">values</a> may be any of the above
+            <a data-cite="RFC4627#section-3">JSON values</a>.</li>
+        </ul>
 
+        Two <a data-cite="RFC4627#section-3">JSON values</a> are considered equal
+        if they are the same if they are the same <a>string</a>,
+        <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+        <a>boolean</a>, or
+        <a data-cite="INFRA#nulls">null</a>,
+        if they are both an <a data-lt="list">array</a> with equal <a data-cite="INFRA#list-item">entries</a>, or
```
afs, line 1493:
```suggestion
        if they are both an <a data-lt="list">array</a> with equal <a data-cite="INFRA#list-item">entries</a>
        in the same order, or
```
It's not the only way to say this e.g. saying they are pairwise equal.

afs:
> it seems that @gkellogg's commits to this PR have edited the section quite a bit.

For reference - commit 8b5d09c56d and later commits.
The `<a>string</a>` reference changes meaning.

Why is the value space not the RFC8785 canonical mapping?

That also makes other changes e.g. i18n - probably inconsequential but mixed in.

afs:
Why is the value space not the RFC8785 canonical mapping?

spec/index.html
gkellogg: Actually, they can. They can't in JSON-LD (other than as JSON values); I had the wrong hat on when thinking about this; I'll correct.

spec/index.html
gkellogg: Number doesn't go to INFRA, which doesn't really have the concept. But, in this context, "string" should probably point to INFRA.

gkellogg:
We discussed at TPAC that the value space of rdf:JSON being the canonical string representation was ill-considered by the JSON-LD WG, and the value space really should be a [JSON value](https://datatracker.ietf.org/doc/html/rfc4627#section-2.1) (reference may need correcting, number, string, boolean, null, array, or map). That is the motivation behind these changes.

Indeed, "string" should probably reference https://infra.spec.whatwg.org/#string rather than our own RDF string.

spec/index.html
gkellogg:
```suggestion
        <a data-cite="RFC4627#section-2.1">JSON value</a>
```

afs:
> We discussed at TPAC that the value space of rdf:JSON being the canonical string representation was ill-considered by the JSON-LD WG and the value space really should be a [JSON value](https://datatracker.ietf.org/doc/html/rfc4627#section-2.1) (

Whose "we" because the RDF-star minutes don't reflect that.

gkellogg:
> > We discussed at TPAC that the value space of rdf:JSON being the canonical string representation was ill-considered by the JSON-LD WG and the value space really should be a [JSON value](https://datatracker.ietf.org/doc/html/rfc4627#section-2.1) (
> 
> Whose "we" because the RDF-star minutes don't reflect that.

I recall the discussion, both at TPAC, and from @pfps comments before. I had suggested that we "correct" the value space based on this notion. Of course, we can revert these changes and go back to the JSON-LD wording if there's not consensus. However, it was my understanding that using the JSON Canonicalization Scheme as the value space was wrong.

> Don't we then deviate from the canonical form algorithm? (is there variance in the RFC canonical form for the same JSON structure?)

We don't deviate (intentionally) from the canonical form defined by JCS/RFC8785.

afs:
> from @pfps comments before

Links please. I don't see anything on #14.

Let's execute on this PR as originally given - which is editorial because the adoption of text from JSON-LD11 was already signallled. 

That's "editorial". It can have all the non-rdf:JSON items thet have also been added.

Then a second PR which is "substantive", "needs discussion".

I am finding it very hard to get a detailed understanding of the substantive change on a fragmented PR.

While I'm not that enthusiastic about defining the a datatype value space based on canonical forms generally, it does work, uses the JSON community RFC 8785, and, for rdf:JSON, has been published.

I think that a new data model in RDF concepts for JSON has long term risks of differences to definitions elsewhere. I would like to see evidence that it does match the RFC 8785 canonicalization.

spec/index.html
afs: JSON Number is a JS number value https://tc39.es/ecma262/#sec-terms-and-definitions-number-value excluding NaN, +Inf, -Inf.

spec/index.html
```diff
@@ -1199,7 +1199,7 @@ <h3>Datatype IRIs</h3>
       MUST refer to the RDF-compatible XSD type named <code>xsd:xxx</code> for
       every XSD type listed in <a href="#xsd-datatypes">section 5.1</a>.</p>
 
-    <p>The following three datatype IRIs are defined in this section:</p>
+    <p>The following three datatype IRIs are defined in Appendix&nbsp;<a href="#section-additional-datatypes" class="sectionRef"></a>:</p>
```
TallTed, line 1202:
```suggestion
    <p>The datatypes identified by the three IRIs below are defined in
    Appendix&nbsp;<a href="#section-additional-datatypes" class="sectionRef"></a>:</p>
```

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    on the primary resource.</p>
+
+  <p>This section discusses the handling of fragment identifiers
+    in representations that encode <a>RDF graphs</a>.</p>
+
+  <p>In RDF-bearing representations of a primary resource,
+    <code>&lt;https://example.com/foo&gt;</code>,
+    the secondary resource identified by a fragment identifier, <code>bar</code>,
+    is the <a>resource</a> <a>denoted</a> by the
+    full <a>IRI</a> <code>&lt;https://example.com/foo#bar&gt;</code> in the <a>RDF graph</a>.
```
TallTed, line 1264:
```suggestion
  <p>In RDF-bearing representations of a primary resource, e.g.,
    <code>&lt;https://example.com/foo&gt;</code>,
    the secondary resource identified by a fragment identifier, e.g., <code>bar</code>,
    is the <a>resource</a> <a>denoted</a> by the
    full <a>IRI</a> in the <a>RDF graph</a>, which would be <code>&lt;https://example.com/foo#bar&gt;</code> in this case.
```

spec/index.html
TallTed: @hartig @gkellogg — I've revised the suggestion. I was moving too quickly while underslept. :-)

spec/index.html
```diff
@@ -1237,6 +1237,101 @@ <h3>Datatype IRIs</h3>
...
+    by allowing <a>IRIs</a>,
+    <a>blank nodes</a>, and
+    <a>literals</a> to appear
+    in any position, i.e., as subject, predicate, object, or graph name.</p>
+
+  <p class="note" id="note-generalized-rdf">Any user of
+    generalized RDF triples, graphs, or datasets needs to be
+    aware that these notions are non-standard extensions of
+    RDF, and their use may cause interoperability problems.
+    There is no requirement on the part of any RDF tool to
```
TallTed, line 1325:
```suggestion
    There is no requirement that any RDF tool
```

spec/index.html
```diff
@@ -1350,106 +1445,72 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+        <ul>
+          <li><a data-cite="INFRA#list">Array</a> <a data-cite="INFRA#list-item">entries</a> may be
+            any of the above <a data-cite="RFC4627#section-2.1">JSON values</a>.</li>
+          <li><a data-cite="INFRA#ordered-map">Map</a> <a data-cite="INFRA#map-key">keys</a> are <a data-cite="INFRA#string">strings</a> with
+            <a data-cite="INFRA#map-value">values</a>, which may be any of the above
+            <a data-cite="RFC4627#section-2.1">JSON values</a>.</li>
+        </ul>
 
+        Two <a data-cite="RFC4627#section-2.1">JSON values</a> are considered equal
+        if they are the same if they are the same <a data-cite="INFRA#string">string</a>,
```
TallTed, line 1485:
```suggestion
        if they are the same <a data-cite="INFRA#string">string</a>,
```

spec/index.html
```diff
@@ -1350,106 +1445,72 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+        <a data-cite="ECMASCRIPT#sec-terms-and-definitions-number-value">number</a>,
+        <a data-cite="INFRA#boolean">boolean</a>, or
+        <a data-cite="INFRA#nulls">null</a>,
+        if they are both <a data-cite="INFRA#list">arrays</a> with <a data-cite="INFRA#list-item">entries</a>
+        which are pairwise equal,
```
TallTed, line 1490:
```suggestion
        <a data-cite="INFRA#nulls">null</a>;
        if they are both <a data-cite="INFRA#list">arrays</a> with <a data-cite="INFRA#list-item">entries</a>
        which are pairwise equal;
```

spec/index.html
```diff
@@ -1350,106 +1445,72 @@ <h3>The <code>rdf:XMLLiteral</code> Datatype</h3>
...
+        or if they are both a <a data-cite="INFRA#ordered-map">map</a> with equal <a data-cite="INFRA#map-entry">map entries</a>.
```
TallTed, line 1491:
```suggestion
        or if they are both <a data-cite="INFRA#ordered-map">maps</a> with equal <a data-cite="INFRA#map-entry">map entries</a>.
```

gkellogg:
@afs said:

> I think that a new data model in RDF concepts for JSON has long term risks of differences to definitions elsewhere. I would like to see evidence that it does match the RFC 8785 canonicalization.

I reverted the changes to the rdf:JSON value space (although it does need to be updated to address the issue on using JCS, at least), and preserved the original commits for another PR. See 39cdf497f0fe9c03fcdaea49e2f3f9b3bf0b9448.

As I said in the commit, the value space could be re-defined (without going to an INFRA representation) by being an RDF string representation conforming to the formatting requirements of the JSON Canonicalization Scheme [[RFC8785](http://localhost:8000/rdf-concepts/spec/#bib-rfc8785)].

Given differences in recollection of discussions, and likely more pressing matters, the WG should consider next steps. See #65.

spec/index.html
```diff
@@ -1467,44 +1467,53 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
...
+        <a data-cite="RFC8259#section-2">Section&nbsp;2 JSON Grammar</a> of
+        [[RFC8259]], and furthermore comply with the following constraints:
         <ul>
-          <li><a data-cite="INFRA#list">Array</a> <a data-cite="INFRA#list-item">entries</a> may be
-            any of the above <a data-cite="RFC4627#section-2.1">JSON values</a>.</li>
-          <li><a data-cite="INFRA#ordered-map">Map</a> <a data-cite="INFRA#map-key">keys</a> are <a data-cite="INFRA#string">strings</a> with
-            <a data-cite="INFRA#map-value">values</a>, which may be any of the above
-            <a data-cite="RFC4627#section-2.1">JSON values</a>.</li>
+          <li>They MUST NOT contain any unnecessary whitespace.</li>
+          <li>Keys in objects MUST be ordered lexicographically.</li>
```
TallTed, line 1476: uh-oh. `lexicographically`, again?

gkellogg: The section can still be improved, but I was trying to just go back to the JSON-LD 1.1 definition. We can either improve that, or go in the native value-space suggested previously and in #65. At this point, the fewer moving parts the better.

pfps:
Some minor comments on rdf:JSON datatype (definition mostly copied from JSON-LD).  This is not an endorsement of the value space, as opposed to either having the value space just be the lexical space or having the value space being some sort of JSON values.

The value space should not have SHOULD.  The value space is what it is.

The value space is defined as a set of RDF strings and then later claimed to be distinct from the value space of xsd:string.   It would be better to state at the beginning that the value space is separate from the value space of xsd:string.  In any case, JSON texts can include surrogate code points so the value space cannot be just RDF strings (assuming that surrogate code points stay out of RDF strings).

Lexicographic order needs to be very carefully defined.   What is the underlying ordering?

What happens if there are duplicate keys in an array?

The lexical-to-value mapping uses ECMASCRIPT and thus coerces JSON numbers to ECMASCRIPT floating point.  Why impose this limitation?

JSON has its own string escapes so it is possible to have double escaping in RDF serializations, as in '"\\xDEAD"'.  I think this is going to cause problems.

spec/index.html
hartig: @TallTed Perfect, that's better. I will commit it.

spec/index.html
hartig:
```suggestion
    There is no requirement for any RDF tool to
```

hartig:
Just for the record I am happy with his PR now (I mention this because, as creator of it, I cannot mark it as approved). After merging it, there can be another PR to edit the section about the `rdf:JSON` datatype.

hartig:
@domel I think all (or almost all) of your recent comments are for an outdated version of this PR. In particular, they are for changes that @gkellogg had made in this PR which were later reverted (to be proposed in a new PR after this one here is merged).

domel:
> @domel I think all (or almost all) of your recent comments are for an outdated version of this PR. In particular, they are for changes that @gkellogg had made in this PR which were later reverted (to be proposed in a new PR after this one here is merged).

Yes, the last two I removed in the last [commit](https://github.com/w3c/rdf-concepts/pull/62/commits/2d9a0bd434d531790d227eb0a87135f0c7e52847).

spec/index.html
```diff
@@ -1086,7 +1086,7 @@ <h3>The XML Schema Built-in Datatypes</h3>
       listed in the following table are the
       <dfn>RDF-compatible XSD types</dfn>. Their use is RECOMMENDED.</p>
 
-    <p>Readers might note that the xsd:hexBinary and xsd:base64Binary
+    <p>Readers might note that the `xsd:hexBinary` and `xsd:base64Binary`
       datatypes are the only safe datatypes for transferring binary
       information.</p>
```
TallTed, line 1091:
```suggestion
    <p>Readers might note that the only safe datatypes for transferring
      binary information are `xsd:hexBinary` and `xsd:base64Binary`.</p>
```

spec/index.html
```diff
@@ -1223,8 +1224,8 @@ <h3>Datatype IRIs</h3>
 
     <p class="note">Semantic extensions of RDF might choose to
       recognize other datatype IRIs
-      and require them to refer to a fixed datatype.  See the RDF
-      Semantics specification [[RDF12-SEMANTICS]] for more information on
+      and require them to refer to a fixed datatype.
```
TallTed, line 1227:
```suggestion
      and require each of them to refer to a fixed datatype.
```

spec/index.html
```diff
@@ -1237,6 +1238,101 @@ <h3>Datatype IRIs</h3>
...
+    <a>literals</a> to appear
+    in any position, i.e., as subject, predicate, object, or graph name.</p>
+
+  <p class="note" id="note-generalized-rdf">Any user of
+    generalized RDF triples, graphs, or datasets needs to be
+    aware that these notions are non-standard extensions of
+    RDF, and their use may cause interoperability problems.
+    There is no requirement for any RDF tool to
+    accept, process, or produce anything beyond standard RDF
+    triples, graphs, and datasets. </p>
```
TallTed, line 1328:
```suggestion
    accept, process, or produce anything beyond standard
    normative RDF triples, graphs, and datasets. </p>
```

spec/index.html
```diff
@@ -1237,6 +1238,101 @@ <h3>Datatype IRIs</h3>
...
+    <dfn class="export">generalized RDF graph</dfn>
+    is a set of generalized RDF triples. A
+    <dfn class="export">generalized RDF dataset</dfn>
+    comprises a distinguished generalized RDF graph, and zero
+    or more pairs each associating an IRI, a blank node, or a literal
+    to a generalized RDF graph.</p>
+
+
+  <p>Generalized RDF triples, graphs, and datasets differ
+    from normative RDF <a>triples</a>,
```
TallTed, line 1314:
```suggestion
    from standard normative RDF <a>triples</a>,
```

spec/index.html
TallTed: Fewer moving parts is generally better, yes.

Still, we now know that `lexicographical` ordering is ambiguous at best, so I think we should replace that with the ordering we believe to be most unambiguous and closest to the original intent, which I *think* now means `Unicode code point order` (or something close to that).

afs (approved):

hartig:
I will merge this one now to make way for @gkellogg to create the follow-up PR with his proposed changes to the section about the rdf:JSON datatype.

PR #67, opened 2023-10-03
[WIP] Change quoted triples to quoted graphs / graph terms

gkellogg:
This draft PR is submitted for discussion purposes.

This makes some minimal changes to replace "quoted triple" with "quoted graph". The notion of "asserted triple" is changed to that of a triples contained in graphs which are not quoted.

In this formulation, quoted graphs are terms which suggest graphs which are not named. This opens up more issues about if such graphs can be considered to be named by a blank node, and if the triples it contains must be in the default graph to be considered to be asserted, or any other named graph.

But, mechanically, changing "quoted triple" to "quoted graph" does not substantially change concepts.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/67.html" title="Last updated on Oct 19, 2023, 8:38 PM UTC (e4963fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/67/727d362...e4963fc.html" title="Last updated on Oct 19, 2023, 8:38 PM UTC (e4963fc)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -66,7 +66,7 @@
       are comprised of a default graph and zero or more named graphs.</li>
   </ul>
 
-  <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
+  <p>RDF 1.2 introduces <a>quoted graphs</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.</p>
 
   <p>RDF 1.2 Concepts introduces key concepts and terminology for RDF 1.2, discusses
@@ -126,7 +126,7 @@
 
     <p>There can be four kinds of <a>nodes</a> in an
       <a>RDF graph</a>: <a>IRIs</a>, <a>literals</a>,
-      <a>blank nodes</a>, and <a>quoted triples</a>.</p>
+      <a>blank nodes</a>, and <a>quoted graphs</a>.</p>
 
   </section>
 
@@ -167,57 +167,71 @@
       exists, without explicitly naming it.</p>
   </section>
 
-  <section id="section-quoted-triples">
-    <h3>Quoted Triples</h3>
+  <section id="section-quoted-graphs">
+    <h3>Quoted Graphs</h3>
 
-    <p>A <a>quoted triple</a> is an <a>RDF term</a> with the components of
-      an <a>RDF triple</a>, which can be used as the <a>subject</a> or <a>object</a>
+    <p>A <a>quoted graph</a> is an <a>RDF term</a> with the components of
+      an <a>RDF graph</a>, which can be used as the <a>subject</a> or <a>object</a>
       of another triple.</p>
 
-    <p>A <a>quoted triple</a> is not necessarily asserted, allowing
+    <p>A <a>quoted graph</a> is not necessarily asserted, allowing
       statements to be made about other statements that may not be
-      asserted within an <a>RDF graph</a>.
+      asserted within another <a>RDF graph</a>.
       This allows statements to be made about relationships that may be contradictory.
-      For a <a>quoted triple</a> to be asserted,
-      it must also appear in a graph as an <a>asserted triple</a>.</p>
+      For a <a>quoted graph</a> to be asserted,
+      the triples it contains must also appear in a graph which is not <a>quoted</a>.</p>
 
-    <p>The following diagram represents a statement with an unasserted <a>quoted triple</a>
+    <p>The following diagram represents a statement with an unasserted <a>quoted graph</a>
       as the subject.</p>
 
-    <figure id="fig-quoted-triple">
+    <figure id="fig-quoted-graph">
       <a href="quoted-triple.svg">
         <!-- Source for this file is at https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8 -->
         <img src="quoted-triple.svg"
-             alt="An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject"
+             alt="An RDF graph containing a triple that references an unasserted quoted graph (with grey background) as the subject"
              style="width:70%"
-             aria-describedby="fig-quoted-triple-alt"/>
+             aria-describedby="fig-quoted-graph-alt"/>
       </a>
-      <figcaption id="fig-quoted-triple-alt">
-        An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject.
+      <figcaption id="fig-quoted-graph-alt">
+        An RDF graph containing a triple that references an unasserted quoted graph (with grey background) as the subject.
       </figcaption>
     </figure>
 
-    <p>A variation on the graph shown in <a href="#fig-quoted-triple"></a> can be described
-      where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
+    <p>A variation on the graph shown in <a href="#fig-quoted-graph"></a> can be described
+      where the triples in the <a>quoted graph</a> are also in the <a>default graph</a>.</p>
 
-    <figure id="fig-quoted-asserted-triple">
+    <figure id="fig-quoted-asserted-graph">
       <a href="quoted-asserted-triple.svg">
         <!-- Source for this file is at https://docs.google.com/drawings/d/1aoUEsb07P-nu5rvPRob7O07CRKgnf7tBw2CmRJNuaQ0 -->
         <img src="quoted-asserted-triple.svg"
              alt="An RDF graph containing two triples where one of them contains the other as the subject; hence, that other triple is both quoted and asserted"
              style="width:70%"
-             aria-describedby="fig-quoted-asserted-triple-alt"/>
+             aria-describedby="fig-quoted-asserted-graph-alt"/>
       </a>
-      <figcaption id="fig-quoted-asserted-triple-alt">
+      <figcaption id="fig-quoted-asserted-graph-alt">
         An RDF graph containing two triples where one of them contains the other as the subject;
         hence, that other triple is both quoted and asserted.
       </figcaption>
     </figure>
 
-    <p>Note that a <a>quoted triple</a> may also have a quoted triple as its subject or object.</p>
+    <p>A quoted graph may contain triples which also have quoted graphs as subjects or objects.</p>
 
-    <p class="issue" data-number="34">
-      This issue considers the potential for additional terminology related to <a>quoted triples</a>.
+    <p class="note">
+      The concept of a quoted graph is similar to that of a named graph,
+      however a quoted graph does not have a name, and is intended to be used as the
+      subject or object of other triples.
+    </p>
+
+    <p class="issue">
+      An alternative interpretation is that quoted graphs can have a blank node name,
+      which can also be used as the subject or object of other triples.
+      These blank node named graphs (or <em>blank graphs</em> may have semantics distinct
+      from graphs named using IRIs or using blank nodes which are not also
+      the subject or object of another triple).
+    </p>
+
+    <p class="issue" data-number="46">
+      This issue considers using <a>quoted graphs</a> as an alternative to quoted triples.
     </p>
   </section>
 
@@ -482,23 +496,6 @@
     but can conform to such other specifications that normatively
     reference terms defined here.</p>
 
-  <p>This specification establishes two conformance levels:</p>
-
-  <ul>
-    <li><dfn class="no-export lint-ignore">Full</dfn> conformance
-      supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
-      containing <a>quoted triples</a>.
-      Concrete syntaxes in which such graphs and datasets can be expressed include
-      [[RDF12-N-TRIPLES]], [[RDF12-N-QUADS]], [[RDF12-TURTLE]], and [[RDF12-TRIG]].</li>
-    <li><dfn class="no-export lint-ignore">Classic</dfn> conformance
-      only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>,
-      that do not contain <a>quoted triples</a>.</li>
-  </ul>
-  <p class="ednote">The conformance levels described above are tentative,
-    and still the subject of group discussion. An alternative to conformance
-    levels, "profiles", may be adopted instead, abandoned, or described in 
-    another specification.</p>
-
   <section id="rdf-strings">
     <h2>Strings in RDF</h2>
 
@@ -559,27 +556,28 @@
       |p| is called the <dfn>predicate</dfn> of the triple, and
       |o| is called the <dfn>object</dfn> of the triple.</p>
 
-    <p>A <dfn data-local-lt="quoted">quoted triple</dfn> is an <a>RDF triple</a> that is
-      used as the <a>subject</a> or <a>object</a> of another triple.</p>
+    <p>A <dfn data-local-lt="quoted">quoted graph</dfn> is an <a>RDF graph</a> that is
+      used as the <a>subject</a> or <a>object</a> of triple.</p>
 
     <p class="note">By the given definitions,
       the <a>subject</a> of an <a>RDF triple</a> may be either an <a>IRI</a>,
-      a <a>blank node</a>, or a <a>quoted triple</a>;
+      a <a>blank node</a>, or a <a>quoted graph</a>;
       the <a>predicate</a> of an <a>RDF triple</a> may only be an <a>IRI</a>;
       the <a>object</a> of an <a>RDF triple</a> may be either an <a>IRI</a>,
-      a <a>blank node</a>, a <a>literal</a> or a <a>quoted triple</a>.</p>
+      a <a>blank node</a>, a <a>literal</a> or a <a>quoted graph</a>.</p>
 
-    <p class="note">The definition of <a>quoted triple</a> is recursive.
-      That is, a <a>quoted triple</a> can itself have a
-      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
-      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
+    <p class="note">The definition of <a>quoted graph</a> is recursive.
+      That is, a <a>quoted graph</a> can itself have a
+      triples where the <a>subject</a> or <a>object</a> components
+      are <a>quoted graphs</a>.
+      However, by this definition, cycles of <a>quoted graphs</a> cannot be created.</p>
 
     <p><a>IRIs</a>, <a>literals</a>,
-      <a>blank nodes</a>, and <a>quoted triples</a> are collectively known as
+      <a>blank nodes</a>, and <a>quoted graphs</a> are collectively known as
       <span id="dfn-rdf-terms"><!-- obsolete term--></span><dfn data-lt="rdf term">RDF terms</dfn>.</p>
 
     <p><a>IRIs</a>, <a>literals</a>,
-      <a>blank nodes</a>, and <a>quoted triples</a> are distinct and distinguishable.
+      <a>blank nodes</a>, and <a>quoted graphs</a> are distinct and distinguishable.
       For example, <code>http://example.org/</code> as a string literal
       is equal to neither <code>http://example.org/</code> as an IRI,
       nor a blank node with the <a>blank node identifier</a>
@@ -589,13 +587,6 @@
       is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph.
       It is possible for a predicate IRI to also occur as a node in
       the same graph.</p>
-
-    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
-      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
-
-    <p class="note">In an <a>RDF graph</a>,
-      a <a>triple</a> may occur as either a <a>quoted triple</a>, an 
-      <a>asserted triple</a>, or both.</p>
   </section>
 
   <section id="section-IRIs">
@@ -1742,8 +1733,8 @@
   <ul>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
-    <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
+    <li>Added <a href="#section-quoted-graphs" class="sectionRef"></a>
+      and definitions for <a>quoted graph</a>.</li>
     <li>Improved the use of IRI terminology,
       and added <a href="#iri-abnf" class="sectionRef"></a>.
       This improves the language using <a>relative IRI references</a>
--- a/spec/index.html
+++ b/spec/index.html
@@ -541,11 +541,11 @@
       </li>
 
       <li>
-        If |t| and <var>t'</var> are <a>RDF triples</a>,
+        If <var>G</var> and <var>G'</var> are <a>RDF graphs</a>,
         |s| is an <a>IRI</a> or a <a>blank node</a>,
         |p| is an <a>IRI</a>, and
         |o| is an <a>IRI</a>, a <a>blank node</a>, or a <a>literal</a>,
-        then (|t|, |p|, |o|), (|s|, |p|, |t|), and (|t|, |p|, <var>t'</var>) are RDF triples.
+        then (<var>G</var>, |p|, |o|), (|s|, |p|, <var>G</var>), and (<var>G</var>, |p|, <var>G'</var>) are RDF triples.
       </li>
     </ul>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -66,7 +66,7 @@
       are comprised of a default graph and zero or more named graphs.</li>
   </ul>
 
-  <p>RDF 1.2 introduces <a>quoted graphs</a> as another kind of <a>RDF term</a>
+  <p>RDF 1.2 introduces <a>graph terms</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.</p>
 
   <p>RDF 1.2 Concepts introduces key concepts and terminology for RDF 1.2, discusses
@@ -126,7 +126,7 @@
 
     <p>There can be four kinds of <a>nodes</a> in an
       <a>RDF graph</a>: <a>IRIs</a>, <a>literals</a>,
-      <a>blank nodes</a>, and <a>quoted graphs</a>.</p>
+      <a>blank nodes</a>, and <a>graph terms</a>.</p>
 
   </section>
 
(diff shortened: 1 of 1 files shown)


gkellogg:
Note that simply using blank node named graphs wouldn't require a change to the 1.1 version of abstract syntax, although it requires some semantics (at least for this aspect).

There is some discussion about triples being asserted, still, and it's not clear what that means given named graphs. We don't say that triples in named graphs are asserted; the annotation syntax can be used within a named graph but asserts triples in the default graph. Would the same be true for the annotation syntax as interpreted on quoted graphs? Does it even make sense to talk about asserted vs. quoted triples/graphs in this context?

Having quoted graphs be different from named graphs is also problematic and doesn't help the query situation and creates a new intellectual barrier to understanding.

TallTed:
> This opens up more issues about if such graphs can be considered to be named by a blank node, and if the triples it contains must be in the default graph to be considered to be asserted, or any other named graph.

I *think* that your intended meaning (if I've understood it) comes across more clearly in this —
> This opens up more issues about whether a graph can be considered to be a named graph if named only by a blank node, and whether, to be considered to be asserted, the triples in such a graph must (also) be in the default graph and/or some graph named by not-a-blank-node.

This will have extra impact on implementations that treat the union of all named graphs as the default graph, meaning that all triples are in the default graph, so there cannot be any triples that are un-asserted by virtue of being only in graphs named by blank nodes.

gkellogg:
Recent discussions have likely overtaken this WIP, but we can use it to express whatever  emerging consensus arises.

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
     </figure>
 
-    <p>Note that a <a>quoted triple</a> may also have a quoted triple as its subject or object.</p>
+    <p>A quoted graph may contain triples which also have quoted graphs as subjects or objects.</p>
 
-    <p class="issue" data-number="34">
-      This issue considers the potential for additional terminology related to <a>quoted triples</a>.
+    <p class="note">
+      The concept of a quoted graph is similar to that of a named graph,
+      however a quoted graph does not have a name, and is intended to be used as the
```
TallTed, line 221: Can a quoted graph *not* be named (identified), with a URI or other non-blank-node?

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
-      a <a>blank node</a>, a <a>literal</a> or a <a>quoted triple</a>.</p>
+      a <a>blank node</a>, a <a>literal</a> or a <a>quoted graph</a>.</p>
 
-    <p class="note">The definition of <a>quoted triple</a> is recursive.
-      That is, a <a>quoted triple</a> can itself have a
-      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
-      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
+    <p class="note">The definition of <a>quoted graph</a> is recursive.
+      That is, a <a>quoted graph</a> can itself have a
+      triples where the <a>subject</a> or <a>object</a> components
```
TallTed, line 571:
```suggestion
      triple where the <a>subject</a> or <a>object</a> components
```

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
+      are <a>quoted graphs</a>.
+      However, by this definition, cycles of <a>quoted graphs</a> cannot be created.</p>
```
TallTed, line 573: I don't expect implementations will always be able to detect, and hence error on, such cycles, until they are revealed by a SPARQL or other query. Thus, better to make the statement a forbiddance, rather than an impossibility.
```suggestion
      However, by this definition, cycles of <a>quoted graphs</a> are not permitted.</p>
```

spec/index.html
afs: A "quoted graph" (better - "graph term") does not have a name - i.e. there is not a attribute or characteristic of the graph that is its name. 

It can be named - that does not affect the graph.

(Anything can be given a name - and it does not change the thing so named.)

spec/index.html
afs: The way the rules work means cycles are impossible. The rules start with RDF 1.1 s-p-o triples (no triple terms) and iteratively allow triples with triple terms, first 1-deep nesting, then 2-deep nesting, ...

Because there is no reference to a triple term before it is created, cycles can't be built. (c.f. "let" and "letrec" in Lisp).

To me, "not permitted" suggests there is a route, but it's banned, which is not the case.

hartig:
@gkellogg It seems you forgot to adapt the second bullet point of the definition of 'RDF triple', i.e., the recursive step of the definition. Since I cannot add a change suggestion for it directly in this PR (because this part is too far from any of the changes in this PR), I am writing here how this part needs to be changed:
```html
<li>
  If |G| and <var>G'</var> are <a>RDF graphs</a>,
  |s| is an <a>IRI</a> or a <a>blank node</a>,
  |p| is an <a>IRI</a>, and
  |o| is an <a>IRI</a>, a <a>blank node</a>, or a <a>literal</a>,
  then (|G|, |p|, |o|), (|s|, |p|, |G|), and (|G|, |p|, <var>G'</var>) are RDF triples.
</li>
```
Let me know if you are okay with me pushing a commit with this change to this PR.

Also, regarding:

> [@gkellogg] Recent discussions have likely overtaken this WIP, but we can use it to express whatever emerging consensus arises.

Are you okay if, as a first step, I push a commit to this PR to change the terminology from using "quoted graph" to using "graph term"?

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
       the <a>predicate</a> of an <a>RDF triple</a> may only be an <a>IRI</a>;
       the <a>object</a> of an <a>RDF triple</a> may be either an <a>IRI</a>,
-      a <a>blank node</a>, a <a>literal</a> or a <a>quoted triple</a>.</p>
+      a <a>blank node</a>, a <a>literal</a> or a <a>quoted graph</a>.</p>
 
-    <p class="note">The definition of <a>quoted triple</a> is recursive.
-      That is, a <a>quoted triple</a> can itself have a
-      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
-      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
+    <p class="note">The definition of <a>quoted graph</a> is recursive.
```
hartig, line 569: Assuming you accept [my proposed change to the definition of the notion of an RDF triple](https://github.com/w3c/rdf-concepts/pull/67#issuecomment-1762820612), it is indeed the case that the definition of quoted graphs (graph terms) is recursive. In fact, the definition of the notion of an RDF graph becomes recursive. Notice, however, that the recursive nature of this definition is an implicit effect of the (new) way the notion of an RDF triple would be defined. More specifically, it is the combination of the two definitions---the definition of an RDF graph as a set of RDF triples and the definition of an RDF triple that may contain RDF graphs---that is recursive. Perhaps it is useful to call this out explicitly.

Nonetheless, as @afs already mentions somewhere else, by the way these two notions are defined, cycles are not possible.

spec/index.html
hartig: I agree with Andy.

hartig: I agree with Andy. Cycles are just not possible in the abstract syntax.

Of course, in concrete serialization formats that use some form of locally-scoped identifiers for graph terms (such as `%{g1}` [as proposed by Andy on the mailing list](https://lists.w3.org/Archives/Public/public-rdf-star-wg/2023Oct/0017.html)), serializations may contain cycles that need to be detected and raised as an error. But that's a topic for these serialization formats, not for the discussion of the abstract syntax in RDF Concepts.

gkellogg:
Like I said, this is just a work in progress, and ideas on the use of graphs as an alternative to quoted triples. I welcome direct contributions to this PR.

spec/index.html
gkellogg: If we use some form of graph identifier to denote the graph, then if used as an object, there is some sense that one graph can “contain”another by reference. This could lead to a graph referencing itself, or through some indirection. If the notion is that graphs are strictly terms, and graph identifiers are only used to express this in a concrete syntax, then cycles are not possible. I’m not clear that we’ve settled on this, but I don’t think there’s a use case for graphs to reference themselves.

afs: The syntax USING-FOR can't create cycles if we say that the representative token `%{g}` can't be used in the definition `FOR {..}`.

N-Quads is harder to get this prescriptive effect.

hartig: @afs If I understand your proposal for the syntax correctly, the `FOR {..}` part may contain such tokens (assuming graph terms may be nested). For instance:
```
USING %{g1} FOR { :s :p :o }
USING %{g2} FOR { %{g1} :p :o }
```
Right?
In this case, there can be cycles in this serialization format:
```
USING %{g1} FOR { %{g2} :p :o }
USING %{g2} FOR { %{g1} :p :o }
```
Yet, these cycles can be prevented if such a token is not allowed to be used inside any `FOR {..}` part before it is declared (i.e., with this requirement, the mention of `%{g2}` in the first line of my example above would be disallowed).

afs: I was assuming we didn't allow use-before-definition - that stops cycles. It is a significant change to TriG/Turtle.

However, there is another kind of cycle.

For any proposal with occurrences that are RDF resources, whether blank graph style or graph term based, it's unavoidable to have occurrence-cycles. 

```
  ## Blank graphs.
  _:a { :s :p _:a }
```
or
```
  ## Declared graph terms
  USING %{g1} FOR { :s :p _:a }
  _:a rdf:occurenceOf %{g1}
```

hartig:
I have pushed the commit with the change to the second bullet point in the definition of 'RDF triple' (see commit https://github.com/w3c/rdf-concepts/pull/67/commits/d8e0faf8141a2b44cb329982c87ec3525c07d0ed).

spec/index.html
niklasl: Graphs (as immutable triple sets) forming descriptions of their occurrences is fine (e.g. records or minds are occasionally quite self-descriptive). It's just referencing (opaquely, only IRIs or blankIDs), so there is no structural/syntactic cycle. Which is as you say is impossible; a set cannot contain *itself*, only references denoting it.

And I take it that these `%{g}` markers aren't references, more like preprocessor substitutions (i.e. like inline definitions / `#define`,  not addresses)? (I took them as only used to flatten out graph terms into a quad structure, and not necessarily to be used in TriG? In SPARQL I can see more utility of that though.)

hartig:
I have renamed "quoted graph" to "graph term" (see commit https://github.com/w3c/rdf-concepts/pull/67/commits/b9820ad1b988be8aa3f851c2dc6baa9265760843).

spec/index.html
afs: Yes, `%{g}` are syntax only. Syntactic substitutions. Turtle doesn't allow the same syntactic unit to be shared in the object position. There is no equivalent of same-subject syntax in `:s :p 1; :p 2`. Writing or reading even 10 triple graph terms repeatedly is laborious.

Do you have a concrete example of an occurrence semantically referencing itself?

gkellogg: Presumably `%{g}` is an aspect only of N-Quads and not available for use in TriG (I'm not in love with that syntax, but it will do for now). We could consider that graph terms can be used only in N-Quads and TriG, although we could consider extending Turtle to support graph terms.

I don't see a way to allow it in N-Triples without a fully expanded representation (which is possible), and it's not worth adding anything to RDF/XML. But, I think adding support to N-Triples would mean that it's no longer a strict subset of N-Quads.

We do have a nomenclature problem with Graph vs. Dataset. We define a Dataset as a default graph plus zero or more named graphs. By this definition datasets aren't required for graph terms, but the need for N-Quads to support `%{g}` conflates the two.

afs: I have tried not to say one name or the other, for the WG to decide.
Other proposals have used TriG because they exploit the TriG Syntax and not give a name to the token-type relationship. They then give rules for blank node use (e.g. not in `GRAPH ?g` or the union graph). There will interactions of syntax and semantics via entailment.

[2023Oct/0017](https://lists.w3.org/Archives/Public/public-rdf-star-wg/2023Oct/0017.html) started as an attempt to separate syntax from semantics.

The `%{g}` syntax doesn't have to be used in T* syntaxes (Turtle, TriG). They have same-subject syntax but no same-object so same-object requires writing out again and reading carefully.

One case is several occurrences of the same graph. Using rdf:type to give the nature of the occurrence:
```
_:a1 rdf:occurenceOf { :s :p :o } .
_:a1 rdf:type ... 
_:a2 rdf:occurenceOf { :s :p :o } .
_:a2 rdf:type ... 
```

If `{ :s :p :o }` has more triples, even as low as 10s, it's visually burdensome to write or read full graph terms for 5 occurrences of the same graph term (same by equality of sets of triples).

```
USING %{g} FOR { ... 10s of triples ... }
_:a1 rdf:occurenceOf %{g} 
_:a1 rdf:type ... 
_:a2 rdf:occurenceOf %{g} 
_:a2 rdf:type ... 
```
It's a shorthand. Blank node graphs proposals have the same issue. Some of the examples carefully put the graph in the subject position.

niklasl: > Yes, `%{g}` are syntax only. Syntactic substitutions. Turtle doesn't allow the same syntactic unit to be shared in the object position. There is no equivalent of same-subject syntax in `:s :p 1; :p 2`. Writing or reading even 10 triple graph terms repeatedly is laborious.

Yes, I agree. This is also related to why I proposed 1) a [new shorthand for unasserted triples/graphs](https://gist.github.com/niklasl/4f52c32ef2d888c172c8584e36c24610#proposal-new-quoted-form), re-using the  annotation syntax, but preceding objects with a "quotation dash" (or an SQL comment if you will), and also 2) adjusting the annotation syntax to be able to [describe multiple occurrences](https://gist.github.com/niklasl/4f52c32ef2d888c172c8584e36c24610#proposal-rdf-star-annotation-occurrences) of the same pattern, along with using identifiers of the occurrence (i.e. named graph). I haven't pushed more on that as we've been in the fundaments; but once we're "out of the weeds" I think it is important to consider. I agree that we must cater for the ergonomics in the concrete syntaxes. (I've been leaning towards supporting these advanced concepts only in TriG/N-Quads; but that depends on other factors.)

I [toyed with that syntax](https://gist.github.com/niklasl/2d02902b81e215b1795981df31927e9b) on all our collected use cases recently. I know this mightn't fly, but I particularly found the data around line [40](https://gist.github.com/niklasl/2d02902b81e215b1795981df31927e9b#file-ucr-examples-blankgraphs-alt-sugar-ttl-L40) and [67](https://gist.github.com/niklasl/2d02902b81e215b1795981df31927e9b#file-ucr-examples-blankgraphs-alt-sugar-ttl-L67) interesting in this variant form. I'm using an adjusted TriG parser supporting this notation, and have done some small tests of turning that into blank graphs, and recently into "fake graph terms" using those "specially named graphs" I [mentioned recently on the list](https://lists.w3.org/Archives/Public/public-rdf-star-wg/2023Oct/0018.html).

(The control options for what goes in the [union graph](https://patterns.dataincubator.org/book/union-graph.html) may leave this wanting though, since there is no real standard and lots of [implementation variation](https://timeaturdean.com/cross-rdf-graph/). Alas I haven't used Fuseki much in the near past, but I recall it having useful [configuration options](https://jena.apache.org/documentation/fuseki2/fuseki-configuration.html) here. I had that in the back of my mind when I thought of [explicit dataset composition](https://gist.github.com/niklasl/c22994e664663b6730613ecc1321c418#explicit-dataset-composition).)

 
> Do you have a concrete example of an occurrence semantically referencing itself?

I've seen (DCAT) Datasets being "part of themselves". And if "referencing" also means used in any position of a triple in the graph it is an occurrence of, I'd say the majority of all web pages are actually examples, since they at least use some `<link rel=stylesheet ... >`, and if you use the page URL as a name for the graph: `<page> { <page> xht:stylesheet <style.css> }`, I think that qualifies?  (Also, specifically in our union catalogue, each thing (book, person, topic, etc.) is described by a record which also describes itself ([example](https://libris.kb.se/xv8b95hg5khnh36/data.ttl?embellished=false)), and each record URI is used as a name for its graph in our quad store ([endpoint](https://libris.kb.se/sparql)).)

gkellogg: > The `%{g}` syntax doesn't have to be used in T* syntaxes (Turtle, TriG). They have same-subject syntax but no same-object so same-object requires writing out again and reading carefully.

Perhaps another thing to consider importing from N3 is the `is ... of` construct, which affords the `:o is :p of :s` structure, which would allow the same-object case without resorting to `%{g}`. JSON-LD has `@reverse` which can be a way of doing the same thing. But that's looking like a syntax-driven solution. If we introduce `%{g}` in one syntax, it may be hard to not include it in others.

afs: I don't see a cycle in [example](https://libris.kb.se/xv8b95hg5khnh36/data.ttl?embellished=false). The record xv8b95hg5khnh36 does not reference itself.

> if you use the page URL as a name for the graph

Isn't that an indirect relationship? 
It would help me to see the N-quads-ish realisation of that convenience syntax. I'm not clear what's the type and what's the occurrence.

afs: > Perhaps another thing to consider importing from N3 is the `is ... of` construct, which affords the `:o is :p of :s` structure, which would allow the same-object case without resorting to `%{g}`. JSON-LD has `@reverse` which can be a way of doing the same thing.

"is..of" is useful when the declaration style of properties is that way round (an ontology I work with currently does use the "uplink" type of properties and "is ... of" would be nice to have). "is..of" might work for hand written input; I'm not sure it covers all cases.

For output at scale, a writer would have to inspect the whole file before starting to write as is necessary for use of `[ ]` syntax to nest blank nodes.

A simple RDF 1.1 Turtle writer is to output subject, then output all property,object pairs.
A very simple RDF 1.1 Turtle writer is output each subject/predicate/object in abbreviated form.

Simple writers output `_:label` rather than `[ ]`.

> If we introduce `%{g}` in one syntax, it may be hard to not include it in others.

Yes, especially if we go the Turtle route because N-Triples is a subset of Turtle.

As a syntax form in Turtle, using `%{g}` is mostly optional.

spec/index.html
```diff
@@ -556,28 +556,28 @@ <h3>Triples</h3>
...
+      a <a>blank node</a>, a <a>literal</a> or a <a>graph term</a>.</p>
 
-    <p class="note">The definition of <a>quoted graph</a> is recursive.
-      That is, a <a>quoted graph</a> can itself have a
+    <p class="note">The definition of <a>graph term</a> is recursive.
+      That is, a <a>graph term</a> can itself contain
       triples where the <a>subject</a> or <a>object</a> components
-      are <a>quoted graphs</a>.
-      However, by this definition, cycles of <a>quoted graphs</a> cannot be created.</p>
+      are <a>graph terms</a>.
```
TallTed, line 572: You'll need to manually apply this, as it encompasses lines that were already deleted (572 & 573), and only makes sense if changing both the line preceding (571) and the line following (572) the deletion.
```suggestion
      triples where the <a>subject</a> or <a>object</a> component
      is a <a>graph term</a>.
```

spec/index.html
gkellogg: Some syntax we might consider is `'{' ( '=' PN_CHARS+ )? triplesBlock? '}'` This would allow a graph term identifier such as `{=g}`, a graph definition including an identifier such as `{=g :s :p :o}`, or a graph term without an identifier `{ :s :p :o }`. This could work in Turtle, TriG and N-Quads (although N-Quads is limited to use the form without `triplesBlock?`).

spec/index.html
TallTed: Hm. It occurs to me, that I see no reason why *both* subject and object couldn't be graph terms. There might be domain/range restrictions that limit the predicates usable between graph terms, but that seems reasonable to me. Thus, this revised suggestion —
```suggestion
      triples where either or both the <a>subject</a> and/or <a>object</a> component
      is a <a>graph term</a>.
```

afs: Yes - it's quite important. We already have illustrative syntax cases for "forward" and "backward" links on graph terms.

It is the graph terms themselves that can't be recursive (i.e. no cycles via graph terms inside graph terms).

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
+      a <a>blank node</a>, a <a>literal</a> or a <a>graph term</a>.</p>
 
-    <p class="note">The definition of <a>quoted triple</a> is recursive.
-      That is, a <a>quoted triple</a> can itself have a
-      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
-      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
+    <p class="note">The definition of <a>graph term</a> is recursive.
+      That is, a <a>graph term</a> can itself contain
+      triples where the <a>subject</a> or <a>object</a> components
+      are <a>graph terms</a>.
```
hartig, line 572: Reestablishing @TallTed's [change suggestion](https://github.com/w3c/rdf-concepts/pull/67#discussion_r1361298265): 
```suggestion
      That is, a <a>graph term</a> can itself contain
      triples where either or both the <a>subject</a> and/or <a>object</a> component
      is a <a>graph term</a>.
```

spec/index.html
hartig: @TallTed I re-created this change suggestion directly in the current version of the PR; see: https://github.com/w3c/rdf-concepts/pull/67/files#r1361896571

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
-  <section id="section-quoted-triples">
-    <h3>Quoted Triples</h3>
+  <section id="section-graph-terms">
+    <h3>Graph Terms</h3>
 
-    <p>A <a>quoted triple</a> is an <a>RDF term</a> with the components of
-      an <a>RDF triple</a>, which can be used as the <a>subject</a> or <a>object</a>
+    <p>A <a>graph term</a> is an <a>RDF term</a> with the components of
+      an <a>RDF graph</a>, which can be used as the <a>subject</a> or <a>object</a>
       of another triple.</p>
```
hartig, line 175: While this sentence made sense when it was talking about quoted triples, it does not make sense anymore now that it is talking about graph terms. I suggest to change it as follows.
```suggestion
    <p>A <a>graph term</a> is an <a>RDF graph</a>
      used as the <a>subject</a> or <a>object</a>
      of a triple.</p>
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
 
-    <p>A <a>quoted triple</a> is not necessarily asserted, allowing
+    <p>A <a>graph term</a> is not necessarily asserted, allowing
```
hartig, line 177: It is not clear to me what it would mean for a graph term to be asserted or not. I think, here it is better to talk about the triples in a graph term. Hence, the following change suggestion.
```suggestion
    <p>The triples in a <a>graph term</a> are not necessarily asserted, allowing
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
       statements to be made about other statements that may not be
-      asserted within an <a>RDF graph</a>.
+      asserted within another <a>RDF graph</a>.
       This allows statements to be made about relationships that may be contradictory.
-      For a <a>quoted triple</a> to be asserted,
-      it must also appear in a graph as an <a>asserted triple</a>.</p>
+      For a <a>graph term</a> to be asserted,
+      the triples it contains must also appear in a graph which is not a <a>graph term</a>.</p>
```
hartig, line 182: Based on the same rationale as [my previous change suggestion](https://github.com/w3c/rdf-concepts/pull/67/files#r1361930313), I suggest to change this sentence as follows.
```suggestion
      For the triples in a <a>graph term</a> to be asserted,
      they must also appear in a graph which is not a <a>graph term</a>.</p>
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
 
-    <p>The following diagram represents a statement with an unasserted <a>quoted triple</a>
+    <p>The following diagram represents a statement with an unasserted <a>graph term</a>
```
hartig, line 184: This sentence and the rest of the text about the examples/figures in this section needs to be revised (assuming that we won't have a notion of an unasserted graph term but rather unasserted triples in a graph term).

gkellogg: The substance of the diagrams should be fine, although they should be renamed accordingly, and we should have a diagram or two that includes a graph term with more than one triple.

TallTed:
Might be worth retitling, `Change "quoted triples" to "quoted graphs", then to "graph terms"` or similar.

spec/index.html
TallTed:
```suggestion
    <p>The following diagram represents a statement with <a>graph term</a> containing unasserted triples
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
       as the subject.</p>
 
-    <figure id="fig-quoted-triple">
+    <figure id="fig-graph-term">
       <a href="quoted-triple.svg">
         <!-- Source for this file is at https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8 -->
         <img src="quoted-triple.svg"
-             alt="An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject"
+             alt="An RDF graph containing a triple that references an unasserted graph term (with grey background) as the subject"
```
TallTed, line 191:
```suggestion
             alt="An RDF graph containing a triple that references a graph term containing unasserted triples (with grey background) as the subject"
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
              style="width:70%"
-             aria-describedby="fig-quoted-triple-alt"/>
+             aria-describedby="fig-graph-term-alt"/>
       </a>
-      <figcaption id="fig-quoted-triple-alt">
-        An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject.
+      <figcaption id="fig-graph-term-alt">
+        An RDF graph containing a triple that references an unasserted graph term (with grey background) as the subject.
```
TallTed, line 196:
```suggestion
        An RDF graph containing a triple that references that references a graph term containing unasserted triples (with grey background) as the subject.
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
       </figcaption>
     </figure>
 
-    <p>A variation on the graph shown in <a href="#fig-quoted-triple"></a> can be described
-      where the <a>quoted triple</a> is also <a data-lt="asserted triple">asserted</a>.</p>
+    <p>A variation on the graph shown in <a href="#fig-graph-term"></a> can be described
+      where the triples in the <a>graph term</a> are also in the <a>default graph</a>.</p>
```
TallTed, line 201:
```suggestion
      where the triples in the <a>graph term</a> (where they are unasserted) are also in the <a>default graph</a> (where they are asserted).</p>
```

hartig:
> [@TallTed] Might be worth retitling, `Change "quoted triples" to "quoted graphs", then to "graph terms"` or similar.

I changed the title of this PR to include the term "graph terms" (along with "quoted graphs")

spec/index.html
niklasl: This syntactic feature seems possible to generalize into some `DEFINE SYMBOL VALUE` feature (to also handle e.g. repetition of strings; but it would be very similar to `PREFIX`...). Of course this is [dangerous territory](https://en.wikipedia.org/wiki/Billion_laughs_attack). Do we have a sense of which use cases call for it? (It might be indicative of an  issue with our design or in the data.) We should create a distinct issue/discussion for that.

afs: > some `DEFINE SYMBOL VALUE`

The [`USING %{label} FOR `](https://lists.w3.org/Archives/Public/public-rdf-star-wg/2023Oct/0017.html) is a "macro" (c.f. `#define`).

Other use cases are "large literals" - ones relating to content `rdf:HTML` , and `rdf:langString` and also `rdf:XMLIteral` and  `rdf:JSON`.RDF is already in "inverse dangerous territory" where content is forced to be written out repeatedly.

Would it be possible to take the use aside from graph terms to a separate issue?

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
 
-    <figure id="fig-quoted-asserted-triple">
+    <figure id="fig-asserted-graph-term">
       <a href="quoted-asserted-triple.svg">
         <!-- Source for this file is at https://docs.google.com/drawings/d/1aoUEsb07P-nu5rvPRob7O07CRKgnf7tBw2CmRJNuaQ0 -->
         <img src="quoted-asserted-triple.svg"
-             alt="An RDF graph containing two triples where one of them contains the other as the subject; hence, that other triple is both quoted and asserted"
+             alt="An RDF graph containing two triples where one of them contains the other as the subject; hence, that other triple is both asserted and the content of a graph term"
```
TallTed, line 207:
```suggestion
             alt="An RDF graph containing two triples, where one triple contains the second as its subject; hence, the second triple is both asserted _and_ contained within a graph term"
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
              style="width:70%"
-             aria-describedby="fig-quoted-asserted-triple-alt"/>
+             aria-describedby="fig-asserted-graph-term-alt"/>
       </a>
-      <figcaption id="fig-quoted-asserted-triple-alt">
+      <figcaption id="fig-asserted-graph-term-alt">
         An RDF graph containing two triples where one of them contains the other as the subject;
```
TallTed, line 212:
```suggestion
        An RDF graph containing two triples, where one triple contains the second as its subject;
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
-        hence, that other triple is both quoted and asserted.
+        hence, that other triple is both asserted and the content of a graph term.
```
TallTed, line 213:
```suggestion
        hence, the second triple is both asserted _and_ contained within a graph term.
```

spec/index.html
```diff
@@ -167,57 +167,71 @@ <h3>Resources and Statements</h3>
...
     </figure>
 
-    <p>Note that a <a>quoted triple</a> may also have a quoted triple as its subject or object.</p>
+    <p>A graph term may contain triples which also have graph terms as subjects or objects.</p>
 
-    <p class="issue" data-number="34">
-      This issue considers the potential for additional terminology related to <a>quoted triples</a>.
+    <p class="note">
+      The concept of a graph term is similar to that of a named graph,
+      however a graph term does not have a name, and is intended to be used as the
```
TallTed, line 221:
```suggestion
      however a graph term does not have a name, and is intended to be used only as the
```

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
+      used as the <a>subject</a> or <a>object</a> of a <a>triple</a>.</p>
 
     <p class="note">By the given definitions,
       the <a>subject</a> of an <a>RDF triple</a> may be either an <a>IRI</a>,
-      a <a>blank node</a>, or a <a>quoted triple</a>;
+      a <a>blank node</a>, or a <a>graph term</a>;
       the <a>predicate</a> of an <a>RDF triple</a> may only be an <a>IRI</a>;
       the <a>object</a> of an <a>RDF triple</a> may be either an <a>IRI</a>,
-      a <a>blank node</a>, a <a>literal</a> or a <a>quoted triple</a>.</p>
+      a <a>blank node</a>, a <a>literal</a> or a <a>graph term</a>.</p>
```
TallTed, line 567:
```suggestion
      a <a>blank node</a>, a <a>literal</a>, or a <a>graph term</a>.</p>
```

spec/index.html
```diff
@@ -559,27 +556,28 @@ <h3>Triples</h3>
...
 
-    <p class="note">The definition of <a>quoted triple</a> is recursive.
-      That is, a <a>quoted triple</a> can itself have a
-      <a>subject</a> or <a>object</a> component which is another <a>quoted triple</a>.
-      However, by this definition, cycles of <a>quoted triples</a> cannot be created.</p>
+    <p class="note">The definition of <a>graph term</a> is recursive.
+      That is, a <a>graph term</a> can itself contain
+      triples where either or both the <a>subject</a> and/or <a>object</a> component
+      is a <a>graph term</a>.
+      However, by this definition, cycles of <a>graph terms</a> cannot be created.</p>
```
TallTed, line 573:
```suggestion
      However, by this definition, cycles of <a>graph terms</a> cannot be created;
      that is, a <a>graph term</a> cannot contain an identical <a>graph term</a> — 
      a copy of itself — at any depth.</p>
```

TallTed:
A bunch of suggested tweaks. As they're mostly discontiguous, I've made each standalone (instead of collecting them into a review). I think you can still apply them as a batch, to the extent that's relevant.

spec/index.html
afs: > Some syntax we might consider is `'{' ( '=' PN_CHARS+ )? triplesBlock? '}'`

Could you write out a complete example with definition and use?
(`{=g}` is a both definition and graph term identifier.)

I seem to remember that N3 did at one time have `:-` (or maybe it was `:=`) for a similar purpose on formula. That got taken out of cwm a very long time ago (well before W3C submission).

General point: I think we have to be careful which characters to use because viewed long term these characters are valuable and finite!

(I used words for clarity - I'm not arguing strongly that words are necessary.)

niklasl: (Aside: at one time we've used SPARQL construct with `VALUES` and an empty graph input as a way to do an "RDF macro" for repetitive chunks of data ([example](https://raw.githubusercontent.com/libris/definitions/develop/source/marc/construct-enums.rq)). A bit like CSVW but for more complex structures. i don't at all think Turtle should adopt something remotely as complex, but  `VALUE ?label "large term"` came to mind...)

spec/index.html
gkellogg: That seems to use triples as resources, whereas one triple is both in the default graph and in a referenced graph term. Perhaps, something like the following:

```suggestion
             alt="An RDF graph containing two triples, where one triple has a subject which is a graph term containing the second triple; hence, the second triple is both asserted _and_ contained within a graph term"
```

spec/index.html
gkellogg: As above:

```suggestion
        An RDF graph containing two triples, where one triple has a subject referencing a graph term containing the second;
```

spec/index.html
gkellogg: > > Some syntax we might consider is `'{' ( '=' PN_CHARS+ )? triplesBlock? '}'`
> 
> Could you write out a complete example with definition and use? (`{=g}` is a both definition and graph term identifier.)

Sure, consider something like the following:

```turtle
PREFIX : <http://example.com/#>
PREFIX rdfx: <http://example.com/rdfx#>

:a :referencesLabledGraph {=g1 :s1 :p :o1} .
:b :referencesLabledGraph {=g1 :s2 :p :o2} .
{=g1} a rdfx:GraphTerm .
:c :referencesUnlabeledGraph { :s3 :p :o3} .
```

The example defines a default graph containing four triples. A graph term (identified by _g1_) is the object of the first two triples and the subject of the third triple.

The first to statements define the same graph term, which collectively add two triples to it.

The third statement makes that graph term the subject of an `rdf:type`.

The fourth statement has another (unidentified) graph term as its object containing a single triple.

> I seem to remember that N3 did at one time have `:-` (or maybe it was `:=`) for a similar purpose on formula. That got taken out of cwm a very long time ago (well before W3C submission).

There was discussion of updating `blankNodePropertyList` to be able to start with `= subject`, such as `[ = :s :p :o] .` which would allow that `[ ... ]` form to be used more recursively. (May have been `:=` due to the problem the `=` is a verb, and so would create parsing issues).

> General point: I think we have to be careful which characters to use because viewed long term these characters are valuable and finite!

Sure, perhaps something other than `=` should be used here; it's not a problem for Turtle/TriG, but would be for N3.

spec/index.html
TallTed: Works for me.

spec/index.html
TallTed: Works for me.

gkellogg:
This proposal has been taken by option 3 "triple terms".

PR #78, opened 2024-02-16
Proposal to adapt the abstract syntax to option 3

hartig:
Given the consensus that we reached in yesterday's call, here is a proposal on how to change the relevant definitions in RDF Concepts in order to capture [Option 3](https://github.com/w3c/rdf-star-wg/blob/main/docs/seeking-consensus-2024-01.html).

Some remarks about this PR:
1. I am currently using the term "triple term" in this proposal while the term "triple descriptor" has been floating around as an alternative name for the same concept. Personally, I like "triple term" more but, of course, I am open to replace "triple term" by "triple descriptor" in this PR if there is consensus in the group towards the latter.
2. At the end of yesterday's call, there was a brief discussion whether the definition of triple terms should be recursive (i.e., whether a triple term may contain another triple term). The current definition in this PR is recursive. I defined it this way because of @pchampin's argument that, without recursion, it becomes impossible to talk about `rdf:nameOf` triples which have a triple term in their object position. Of course, if the group decides against such a recursive definition, this PR can easily be adapted to make the definition non-recursive.
3. I have separated the definitions of "RDF triple" and "triple term" from one another even if these two definitions are essentially the same (syntactically, an RDF triple and a triple term are exactly the same thing). My rationale for this separation is to make it more obvious that these are two different concepts. I have also added a brief 'Note' on this topic.
4. I have removed the definition of "asserted triple" because I think, by clearly separating the notion of an RDF triple from the notion of a triple term, it becomes obsolete to talk about asserted triples. Every triple (in a graph) is asserted; triple terms are not members of graphs.
5. This PR changes only the definitions. I explicitly decided not to touch Section 1.3, which (currently) provides an informal introduction of quoted triples. We will certainly have to change that section as well, but I think that should be a different PR. We should first make sure that we have the definitions right.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/78.html" title="Last updated on Mar 7, 2024, 6:14 PM UTC (497bbb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/78/2ff13d4...497bbb2.html" title="Last updated on Mar 7, 2024, 6:14 PM UTC (497bbb2)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1789,8 +1789,12 @@
   <ul>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
+    <!-- The following bullet point should be removed when Section 1.3 is changed
+         to introduce triple terms instead of quoted triples. -->
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
+      <s>and definitions for <a>quoted triple</a> and <a>asserted triple</a></s>.</li>
+    <li>Added the notion of a <a>triple term</a> and extended the definition of
+      <a>RDF triple</a> to permit triple terms as objects.</li>
     <li>Added the <a>base direction</a> element as part of 
       a <a>literal</a>,
       and a description of its use in <a href="#section-text-direction" class="sectionRef"></a>.</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1798,7 +1798,7 @@
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      <s>and definitions for <a>quoted triple</a> and <a>asserted triple</a></s>.</li>
+      <s>and definitions for <a>quoted triple</a> and <a>asserted triple</a></s>.
 
     <div class="issue" data-number="81">
       The notion of quoted triples is deprecated and will be removed again from this document. Currently, Section <a href="#section-quoted-triples" class="sectionRef"></a> still describes this notion. Once this section is changed to describe whatever new concept the WG settles on (currently, this may be <a>triple terms</a>---see the next bullet point), the previous bullet point can be removed and the addition of the then-rewritten section can be mentioned in the next bullet point.
--- a/spec/index.html
+++ b/spec/index.html
@@ -548,7 +548,7 @@
         or a <a>triple term</a>.</li>
     </ul>
 
-    <p>An <dfn>triple term</dfn> is a 3-tuple that is defined recursively as follows:</p>
+    <p>A <dfn>triple term</dfn> is a 3-tuple that is defined recursively as follows:</p>
 
     <ul>
       <li>
@@ -561,8 +561,8 @@
       <li>
         If |s| is an <a>IRI</a> or a <a>blank node</a>,
         |p| is an <a>IRI</a>, and
-        |t| is a <a>triple term</a>,
-        then (|s|, |p|, |t|) is a triple term.
+        |o| is a <a>triple term</a>,
+        then (|s|, |p|, |o|) is a triple term.
       </li>
     </ul>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -68,7 +68,7 @@
   </ul>
 
   <p>RDF 1.2 introduces <a>triple terms</a> as another kind of <a>RDF term</a>
-    which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
+    which can be used as the <a>object</a> of another <a>triple</a>.
     RDF 1.2 also introduces <a>directional language-tagged strings</a>,
     which contain a <a>base direction</a> element that allows the
     initial text direction to be specified for presentation by a user agent.</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -67,7 +67,7 @@
       are comprised of a default graph and zero or more named graphs.</li>
   </ul>
 
-  <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
+  <p>RDF 1.2 introduces <a>triple terms</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
     RDF 1.2 also introduces <a>directional language-tagged strings</a>,
     which contain a <a>base direction</a> element that allows the
@@ -130,7 +130,7 @@
 
     <p>There can be four kinds of <a>nodes</a> in an
       <a>RDF graph</a>: <a>IRIs</a>, <a>literals</a>,
-      <a>blank nodes</a>, and <a>quoted triples</a>.</p>
+      <a>blank nodes</a>, and <a>triple terms</a>.</p>
 
   </section>
 
@@ -492,12 +492,12 @@
   <ul>
     <li><dfn class="no-export lint-ignore">Full</dfn> conformance
       supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
-      containing <a>quoted triples</a>.
+      with <a>triples</a> that contain <a>triple terms</a>.
       Concrete syntaxes in which such graphs and datasets can be expressed include
       [[RDF12-N-TRIPLES]], [[RDF12-N-QUADS]], [[RDF12-TURTLE]], and [[RDF12-TRIG]].</li>
     <li><dfn class="no-export lint-ignore">Classic</dfn> conformance
-      only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>,
-      that do not contain <a>quoted triples</a>.</li>
+      only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>
+      with <a>triples</a> that do not contain <a>triple terms</a>.</li>
   </ul>
   <p class="ednote">The conformance levels described above are tentative,
     and still the subject of group discussion. An alternative to conformance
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
+      is a 3-tuple (|s|, |p|, |o|) where:</p>
+
+    <ul>
+      <li>|s| is an <a>IRI</a> or a <a>blank node</a>,</li>
+      <li>|p| is an <a>IRI</a>, and</li>
+      <li>|o| is an <a>IRI</a>, a <a>blank node</a>, a <a>literal</a>,
+        or a <a>triple term</a>.</li>
+    </ul>
+
+    <p>An <dfn>triple term</dfn> is a 3-tuple that is defined recursively as follows:</p>
```
hartig, line 551:
```suggestion
    <p>A <dfn>triple term</dfn> is a 3-tuple that is defined recursively as follows:</p>
```

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
 
       <li>
-        If |t| and <var>t'</var> are <a>RDF triples</a>,
-        |s| is an <a>IRI</a> or a <a>blank node</a>,
+        If |s| is an <a>IRI</a> or a <a>blank node</a>,
         |p| is an <a>IRI</a>, and
-        |o| is an <a>IRI</a>, a <a>blank node</a>, or a <a>literal</a>,
-        then (|t|, |p|, |o|), (|s|, |p|, |t|), and (|t|, |p|, <var>t'</var>) are RDF triples.
+        |t| is a <a>triple term</a>,
+        then (|s|, |p|, |t|) is a triple term.
```
hartig, line 565:
```suggestion
        |o| is a <a>triple term</a>,
        then (|s|, |p|, |o|) is a triple term.
```

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
       its <a>lexical form</a>
       is not equal to the IRI <code>http://example.org/</code>,
       nor to a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
 
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
-      is the set of <a>subjects</a> and <a>objects</a> of <a>triples</a> in the graph.
+      is the set of <a>subjects</a> and <a>objects</a> of the <a>triples</a> in the graph.
       It is possible for a predicate IRI to also occur as a node in
       the same graph.</p>
```
gkellogg, line 604:
```suggestion
  </section>
```

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
 
-    <p>An <dfn data-local-lt="asserted">asserted triple</dfn>
-      is an <a>RDF triple</a> that is an element of an <a>RDF graph</a>.</p>
-
-    <p class="note">In an <a>RDF graph</a>,
-      a <a>triple</a> may occur as either a <a>quoted triple</a>, an 
-      <a>asserted triple</a>, or both.</p>
```
gkellogg, line 604: Why loose the notion of "asserted triple", is that not just as important with the change to triple terms?

gkellogg (changes requested):
* One missing `</section>.
* Not sure what the motivation for removing "asserted triple" is.
* This will need some update to Changes since, otherwise, looks good.

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
   <section id="section-triples">
     <h3>Triples</h3>
 
-    <p>An <dfn data-local-lt="triple">RDF triple</dfn> is a 3-tuple that is defined recursively as follows:</p>
+    <p>An <dfn data-local-lt="triple">RDF triple</dfn> (usually called "triple")
+      is a 3-tuple (|s|, |p|, |o|) where:</p>
+
+    <ul>
+      <li>|s| is an <a>IRI</a> or a <a>blank node</a>,</li>
```
pfps, line 545: This document says above that a triple can have a triple term as a subject.

spec/index.html
```diff
@@ -67,7 +67,7 @@
       are comprised of a default graph and zero or more named graphs.</li>
   </ul>
 
-  <p>RDF 1.2 introduces <a>quoted triples</a> as another kind of <a>RDF term</a>
+  <p>RDF 1.2 introduces <a>triple terms</a> as another kind of <a>RDF term</a>
     which can be used as the <a>subject</a> or <a>object</a> of another <a>triple</a>.
```
hartig, line 71: As [pointed out](https://github.com/w3c/rdf-concepts/pull/78/files#r1492917262) by @pfps, the new definition of 'RDF triple' in this PR does not permit triple terms in the subject position. Hence, this sentence here needs to be adapted.
```suggestion
    which can be used as the <a>object</a> of another <a>triple</a>.
```

spec/index.html
hartig: @pfps

> This document says above that a triple can have a triple term as a subject.

I have [now changed](https://github.com/w3c/rdf-concepts/pull/78/files#r1494587309) that statement above in order to be aligned with the new definition here.

spec/index.html
hartig: @gkellogg 
I don't see what introducing this notion of "asserted triple" adds; in particular now that we want to establish a clear conceptual separation between the notion of an RDF triple (or, short, triple) and the notion of a triple term. By the way the notion of "asserted triple" was defined here, whenever you want to talk about an asserted triple in a particular graph, you may as well talk about a triple in that graph.

Of course, we may include a statement somewhere, saying that every triple that is an element of an RDF graph is considered as asserted in that graph, but I don't think we need to introduce the notion of "asserted triple" to do so.

spec/index.html
pfps: I thought that triple terms were to be allowed as both subjects and objects.  Lots of use cases are best expresses with triple terms as subjects.

pfps (changes requested):

spec/index.html
pfps: See https://github.com/w3c/rdf-concepts/issues/80

gkellogg: People are still being confused between Triple Terms and Occurrences (still needs a better name). A Triple Term only is intended to exist as the object of a triple where the predicate is `rdf:nameOf` (or similar). An occurrence in the abstract syntax is just another node (IRI or Blank Node to be determined).

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
 
       <li>
-        If |t| and <var>t'</var> are <a>RDF triples</a>,
-        |s| is an <a>IRI</a> or a <a>blank node</a>,
+        If |s| is an <a>IRI</a> or a <a>blank node</a>,
         |p| is an <a>IRI</a>, and
-        |o| is an <a>IRI</a>, a <a>blank node</a>, or a <a>literal</a>,
-        then (|t|, |p|, |o|), (|s|, |p|, |t|), and (|t|, |p|, <var>t'</var>) are RDF triples.
+        |o| is a <a>triple term</a>,
+        then (|s|, |p|, |o|) is a triple term.
```
gkellogg, line 565: Again, this describes an RDF triple, but in this case containing <var>t`</var>, <var>t`</var> is a triple term, but <var>t</var> is an RDF triple.

spec/index.html
```diff
@@ -1791,8 +1789,12 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
   <ul>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
+    <!-- The following bullet point should be removed when Section 1.3 is changed
+         to introduce triple terms instead of quoted triples. -->
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      and definitions for <a>quoted triple</a> and <a>asserted triple</a>.</li>
+      <s>and definitions for <a>quoted triple</a> and <a>asserted triple</a></s>.</li>
```
gkellogg, line 1795: We may want something here, and in the Quoted Triples section, that this notion is deprecated and will be removed or changed when the WG resolves nomenclature.

gkellogg (commented):
Still some changes needed, I think.

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
+
+    <p>A <dfn>triple term</dfn> is a 3-tuple that is defined recursively as follows:</p>
 
     <ul>
       <li>
         If |s| is an <a>IRI</a> or a <a>blank node</a>,
         |p| is an <a>IRI</a>, and
         |o| is an <a>IRI</a>, a <a>blank node</a>, or a <a>literal</a>,
-        then (|s|, |p|, |o|) is an RDF triple.
+        then (|s|, |p|, |o|) is a triple term.
```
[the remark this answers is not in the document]
hartig: I agree that having 'triple term' here, instead of 'RDF triple', deviates from the grammar in the seeking-consensus table. Yet, if we are aiming to clearly distinguish between the concept of an RDF triple (as something that can be used as an element of RDF graphs) and the concept of a triple term (as something that can (recursively) be used as a component of RDF triples), then it should be 'triple term' here.

spec/index.html
hartig: Based on the same rationale as in [my reply to your other comment](https://github.com/w3c/rdf-concepts/pull/78/files#r1495450008), I think this should be 'triple term' here.

spec/index.html
hartig: I created issue #81 as a reminder that Section 1.3 needs to be changed, and extended this PR (see commit https://github.com/w3c/rdf-concepts/pull/78/commits/195505a07a352e1d6fc4c2825dd5a408a50c7900) by adding two references to this issue, one below this bullet point here and the other one at the beginning of Section 1.3.

gkellogg (approved):

TallTed (commented):
Not sure if these comments are still valid... apparently closed a review without successfully submitting it.

spec/index.html
hartig: @TallTed your suggestion for this line is to add triple term as another option for what the subject may be. My response to this suggestion is in [my comment above](https://github.com/w3c/rdf-concepts/pull/78/files#r1494748017).

TallTed: Picking up in #80.

TallTed (approved):
OK to merge #78 and follow up via #80

pfps (approved):

pchampin:
merging, as agreed during today's call: https://www.w3.org/2024/03/07-rdf-star-minutes.html#x282

spec/index.html
```diff
@@ -538,72 +538,70 @@ <h2>RDF Graphs</h2>
...
 
-    <p>An <dfn data-local-lt="triple">RDF triple</dfn> is a 3-tuple that is defined recursively as follows:</p>
+    <p>An <dfn data-local-lt="triple">RDF triple</dfn> (usually called "triple")
+      is a 3-tuple (|s|, |p|, |o|) where:</p>
+
+    <ul>
+      <li>|s| is an <a>IRI</a> or a <a>blank node</a>,</li>
+      <li>|p| is an <a>IRI</a>, and</li>
+      <li>|o| is an <a>IRI</a>, a <a>blank node</a>, a <a>literal</a>,
+        or a <a>triple term</a>.</li>
```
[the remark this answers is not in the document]
gkellogg: This is being deferred to #80.

PR #95, opened 2024-07-05
add definition for asserted and update change log accordingly

pfps:
There are several uses of asserted in Concepts and Semantics so there should be a definition of the term.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/95.html" title="Last updated on Aug 1, 2024, 4:50 PM UTC (bd8bf2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/95/8a1700c...bd8bf2e.html" title="Last updated on Aug 1, 2024, 4:50 PM UTC (bd8bf2e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1823,7 +1823,8 @@
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      <s>and definitions for <a>quoted triple</a></s> and definition for <a>asserted triple</a></s>.
+      <s>and definitions for <a>quoted triple</a></s>
+      and definition for <a>asserted triple</a>.
 
     <div class="issue" data-number="81">
       The notion of quoted triples is deprecated and will be removed again from this document. Currently, Section <a href="#section-quoted-triples" class="sectionRef"></a> still describes this notion. Once this section is changed to describe whatever new concept the WG settles on (currently, this may be <a>triple terms</a>---see the next bullet point), the previous bullet point can be removed and the addition of the then-rewritten section can be mentioned in the next bullet point.
--- a/spec/index.html
+++ b/spec/index.html
@@ -539,7 +539,7 @@
 
   <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
-  <p>An RDF triple is said to be <dfn data-lt="asserted|asserted triple">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
+  <p>An RDF triple is said to be <dfn id="dfn-asserted-triple" data-lt="asserted|asserted triple">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
 
   <section id="section-triples">
     <h3>Triples</h3>
--- a/spec/index.html
+++ b/spec/index.html
@@ -539,9 +539,7 @@
 
   <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
-  <p class=issue>
-    Do we still need to define <dfn>asserted triple</dfn>?.
-  </p>
+  <p>An RDF triple is said to be <dfn="asserted">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
 
   <section id="section-triples">
     <h3>Triples</h3>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -539,9 +539,7 @@ <h2>RDF Graphs</h2>
 
   <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
-  <p class=issue>
-    Do we still need to define <dfn>asserted triple</dfn>?.
-  </p>
+  <p>An RDF triple is said to be <dfn="asserted">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
```
gkellogg, line 542: You probably want to use the following:

```suggestion
  <p>An RDF triple is said to be <dfn data-lt="asserted|asserted triple">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
```

spec/index.html
```diff
@@ -1825,7 +1823,7 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      <s>and definitions for <a>quoted triple</a> and <a>asserted triple</a></s>.
+      <s>and definitions for <a>quoted triple</a></s> and definition for <a>asserted triple</a></s>.
```
gkellogg, line 1826: IIUC, the term "quoted triple" is now obsolete. It can simply be removed when it's complete, but the whole section is affected. It probably becomes "triple term", but that's still a bit up in the air.

gkellogg (changes requested):
While "asserted triple" is its own concept which needs definition, so is "triple term" and "quoted triple" becomes obsolete. WG hasn't proceeded quite enough to make the "quoted triple" change quite yet, though. Otherwise, adding a term for "asserted triple" seems good. We might want to make sure that the term defined is "dfn-asserted-triple" rather than "dfn-asserted" and that may need the din to be tweaked further.

spec/index.html
TallTed: There's now an orphan `</s>`. I think this suggestion makes the right correction, but I'm not certain.
```suggestion
      <s>and definitions for <a>quoted triple</a></s>
      and definition for <a>asserted triple</a>.
```

gkellogg: That strike-through text will eventually go away when we remove quoted triple. But, this is an appropriate change for now.

gkellogg (approved):

spec/index.html
```diff
@@ -1823,7 +1823,8 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
       for informative definition of a <a>quad</a>.</li>
     <li>Added <a href="#section-quoted-triples" class="sectionRef"></a>
-      <s>and definitions for <a>quoted triple</a></s> and definition for <a>asserted triple</a></s>.
+      <s>and definitions for <a>quoted triple</a></s>
+      and definition for <a>asserted triple</a>.
```
TallTed, line 1827:
```suggestion
      and <a>asserted triple</a>.
```

hartig (approved):

spec/index.html
```diff
@@ -539,9 +539,7 @@ <h2>RDF Graphs</h2>
 
   <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
-  <p class=issue>
-    Do we still need to define <dfn>asserted triple</dfn>?.
-  </p>
+  <p>An RDF triple is said to be <dfn data-lt="asserted|asserted triple">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
```
gkellogg, line 542: This creates the definition anchor `dfn-asserted`, which seems overly broad. We can tweak the definition so that it creates `dfn-asserted-triple` which may be more suitable.

```suggestion
  <p>An RDF triple is said to be <dfn id="dfn-asserted-triple" data-lt="asserted|asserted triple">asserted</dfn> in an RDF graph if it is an element of the RDF graph.</p>
```

gkellogg (changes requested):

gkellogg (approved):

PR #96, opened 2024-07-15
Change the section on Quoted Triples to Triple Terms.

### LINKED ISSUE #81: Change section that (currently) introduces 'quoted triples'
The current draft of RDF 1.2 Concepts contains a section that describes what quoted triples are ([Section 1.3](https://w3c.github.io/rdf-concepts/spec/#section-quoted-triples)). As the WG has moved away from the notion of quoted triples, this section needs to be changed such that it does not mention quoted triples anymore but, instead, describes whatever new concept the WG settles on.

gkellogg:
* Updates the diagrams to use a reifier.
* These represent changes based on the [minimal baseline](https://github.com/w3c/rdf-star-wg/wiki/RDF-star-%22minimal-baseline%22#rdf-star-profile-transparent---aka-minimal-baseline-draft-20240628)

Fixes #81.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/96.html" title="Last updated on Aug 1, 2024, 8:28 PM UTC (245a540)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/96/df199fd...245a540.html" title="Last updated on Aug 1, 2024, 8:28 PM UTC (245a540)">Diff</a>


domel:
Just a little comment on the SVG images. I see that everything is converted to paths. It seems a bit redundant. All the elements are simple primitives. The text, which is a path, also seems a bit exotic.

niklasl:
I simplified the old ones a while ago (just [put them in a gist](https://gist.github.com/niklasl/4a1ff0bde1571da86e6853bc37d12409)). I can make such versions of these new ones if you want to?

In any case I'd like to tweak them so that the arrow in the new ones go to the reified edge (in the current transparent version it looks like the reifier just links to the subject).

gkellogg:
> I simplified the old ones a while ago (just [put them in a gist](https://gist.github.com/niklasl/4a1ff0bde1571da86e6853bc37d12409)). I can make such versions of these new ones if you want to?
> 
> In any case I'd like to tweak them so that the arrow in the new ones go to the reified edge (in the current transparent version it looks like the reifier just links to the subject).

I show that `_:reif01` is the subject for both the triple term and `:bob`, which essentially represents what the previous version with "Quoted Triples" does.

The examples in your gist don't use a reifier, which I thought was important for these examples. But, perhaps an example that represents something like the syntactic sugar `<< _:reif01 | <alice> :name "Alice" >> :accordingTo :bob` might illustrate the intended usage better. Perhaps, but adding `_:reif01` as a label to the grey box.

niklasl:
Yes, my gist example is from last year, they *do not* reflect what we need. I just put them there to show an example of the primitive shapes asked for (see the sources).

gkellogg:
I'd be happy to use simplified versions of the diagrams. Looking at the source for each, they would seem easy enough to maintain by hand. Feel free to push a commit with newer versions of these diagrams, although we should probably continue to reference (in the comment) the Google Doc version as an alternative representation.

niklasl:
I don't seem to have permission to push to this repo. Meanwhile I put [proposed new versions  in another gist](https://gist.github.com/niklasl/da6b1cbcb73637d7cdaefb088f4d616d). Feel free to grab those copies if the permissions are hard to sort out.

spec/index.html
```diff
@@ -172,61 +172,59 @@ <h3>Resources and Statements</h3>
...
+    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>.
+      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
+      Assertions on the <a>triple term</a> are made using the <a>reifier</a>.</p>
 
-    <figure id="fig-quoted-triple">
-      <a href="quoted-triple.svg">
-        <!-- Source for this file is at https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8 -->
-        <img src="quoted-triple.svg"
-             alt="An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject"
+    <p>The following diagram represents a statement an a reification of an an unasserted <a>triple term</a>.</p>
```
niklasl, line 193:
```suggestion
    <p>The following diagram represents a statement and a reification of an unasserted <a>triple term</a>.</p>
```
Not sure about the formulation though, since it mixes statement (interpretation?) and triple (graph?). The triple term is used in another triple to encode a reification?

(The specs may need more clarity in general about whether a triple "encodes a statement", and/or asserting a triple (which "says that some relationship holds") is a statement (which only "corresponds" to a triple).)

gkellogg:
> I don't seem to have permission to push to this repo. Meanwhile I put [proposed new versions in another gist](https://gist.github.com/niklasl/da6b1cbcb73637d7cdaefb088f4d616d). Feel free to grab those copies if the permissions are hard to sort out.

Thanks, those are great. I'll push them up. Probably still needs a paragraph or two on the rationale for using reifiers.

spec/index.html
gkellogg: There's a reason this is still a draft, and I'm hoping for other contributors to help improve the language for clarity and accuracy.

In this case, it is probably better to stick with "triples" and not "statements".

domel (approved):

spec/index.html
```diff
@@ -172,61 +172,67 @@ <h3>Resources and Statements</h3>
...
-
-    <figure id="fig-quoted-triple">
-      <a href="quoted-triple.svg">
-        <!-- Source for this file is at https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8 -->
-        <img src="quoted-triple.svg"
-             alt="An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject"
+    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>.
+      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
+      Assertions on the <a>triple term</a> are made using the <a>reifier</a>.</p>
+
```
gkellogg, line 192: This could be the place to define a term such as "reified triple term" to describe the syntactic structure that optionally includes a "reifier" and a "triple term".

spec/index.html
```diff
@@ -172,61 +172,67 @@ <h3>Resources and Statements</h3>
...
+    <p class="note">Using the <a>reifier</a> as an indirect way of referencing a <a>triple term</a>
+      allows annotations to be separated from each other,
+      which might be necessary if the same <a>triple term</a> was derived from different sources.
+      Annotations are always made on the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
```
hartig, line 196: Since you are using the word "assertions" before this Note, I suggest to continue using it within the Note (instead of switching to "annotations" here). Additionally, the word "annotations" may be misleading as we also have the Turtle shorthand notation that we call "annotation syntax" and that does more than only making assertions on the triple term.
```suggestion
      Assertions are always made using the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
```

spec/index.html
```diff
@@ -172,61 +172,67 @@ <h3>Resources and Statements</h3>
...
+      of different triples.
+      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]],
+      may have shortcuts for describing a <a>triple term</a> with its <a>reifier</a>.</p>
```
hartig, line 199:
```suggestion
      may have shortcuts for capturing a <a>triple term</a> with its <a>reifier</a>.</p>
```

spec/index.html
```diff
@@ -593,6 +599,13 @@ <h3>Triples</h3>
       <a>object</a> component which is another <a>triple term</a>.
       However, by this definition, cycles of <a>triple terms</a> cannot be created.</p>
 
+    <p class="note"><a>Triples</a> with a <a>triple term</a> object SHOULD
+      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
+      as the <a>predicate</a>.
```
hartig, line 604:
```suggestion
    <p class="note">Every <a>triple</a> with a <a>triple term</a> as its object SHOULD
      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
      as its <a>predicate</a>.
```

spec/index.html
```diff
@@ -593,6 +599,13 @@ <h3>Triples</h3>
...
+      <a>Triples</a> where the <a>object</a> is not a <a>triple term</a> SHOULD NOT
+      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
+      as the <a>predicate</a>.</p>
```
hartig, line 607:
```suggestion
      Every <a>triple</a> whose <a>object</a> is not a <a>triple term</a> SHOULD NOT
      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
      as its <a>predicate</a>.</p>
```

spec/index.html
```diff
@@ -1824,15 +1838,10 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
...
-    <div class="issue" data-number="81">
-      The notion of quoted triples is deprecated and will be removed again from this document. Currently, Section <a href="#section-quoted-triples" class="sectionRef"></a> still describes this notion. Once this section is changed to describe whatever new concept the WG settles on (currently, this may be <a>triple terms</a>---see the next bullet point), the previous bullet point can be removed and the addition of the then-rewritten section can be mentioned in the next bullet point.
-    </div>
-
-    <li>Added the notion of a <a>triple term</a> and extended the definition of
-      <a>RDF triple</a> to permit triple terms as objects.</li>
+    <li>Added <a href="#section-triple-terms" class="sectionRef"></a>
+      and definitions for <a>triple term</a> and <a>asserted triple</a>
+      and extended the definition of
+      <a>RDF triple</a> to permit triple terms as objects..
```
hartig, line 1844:
```suggestion
      <a>RDF triple</a> to permit triple terms as objects.</li>
```

hartig (commented):
Here are a few edit suggestions.

spec/index.html
```diff
@@ -172,61 +172,67 @@ <h3>Resources and Statements</h3>
...
-    <p>The following diagram represents a statement with an unasserted <a>quoted triple</a>
-      as the subject.</p>
-
-    <figure id="fig-quoted-triple">
-      <a href="quoted-triple.svg">
-        <!-- Source for this file is at https://docs.google.com/drawings/d/1I_QxbUgnQXumXzb8c0WNJHIQ-mtRs2S80dDG6i9aOD8 -->
-        <img src="quoted-triple.svg"
-             alt="An RDF graph containing a triple that references an unasserted quoted triple (with grey background) as the subject"
+    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>.
+      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
```
hartig, line 190: Given that this section is non-normative, I wonder whether it makes sense to put a definition here (`<dfn>reifier</dfn>`) ?

gkellogg: As a reifier is really just a concept, and not a component of the abstract syntax itself, it seems to me that a non-normative definition is fine. I think we also need a concept for "reified triple term", or whatever we come up with. In Turtle, both "reifier" and "reified triple term" will be normatively defined as part of the grammar, and it may be elsewhere, so having the source definition in RDF Concepts makes sense.

But, I'm open to suggestions to handle this better.

hartig: Makes sense.

PR #105, opened 2024-10-22
GH-100: Literal equality and language tags

### LINKED ISSUE #100: Literal equality and language tags
The last paragraph of Section [§3.3 Literals](https://www.w3.org/TR/rdf12-concepts/#section-Graph-Literal) says

> Literal term equality: Two literals are term-equal (the same [RDF literal](https://www.w3.org/TR/rdf12-concepts/#dfn-literal)) if and only if the two [lexical forms](https://www.w3.org/TR/rdf12-concepts/#dfn-lexical-form), the two [datatype IRIs](https://www.w3.org/TR/rdf12-concepts/#dfn-datatype-iri), the two [language tags](https://www.w3.org/TR/rdf12-concepts/#dfn-language-tag) (if any), and the two [base directions](https://www.w3.org/TR/rdf12-concepts/#dfn-base-direction) (if any) compare equal, using [case sensitive matching](https://www.w3.org/TR/i18n-glossary/#dfn-case-sensitive) 

But the third bullet point at the beginning of the section says:

> The language tag _MUST_ be well-formed according to [section 2.2.9](https://www.rfc-editor.org/rfc/rfc5646#section-2.2.9) of [[BCP47](https://www.w3.org/TR/rdf12-concepts/#bib-bcp47 "Tags for Identifying Languages")], and _MUST_ be treated consistently, that is, in a case insensitive manner.

The two statements seem to contradict one another on the language tags: they are to be compared in a case-insensitive manner...

afs:
This closes #100.

This suggestion turns the list into bullet points.

Notes:
It is written as "case sensitive" in i18n-glossary.
It is "ASCII case-insenitive" in i18n-glossary.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/105.html" title="Last updated on Oct 24, 2024, 7:34 AM UTC (9c5ef40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/105/eefa2a6...9c5ef40.html" title="Last updated on Oct 24, 2024, 7:34 AM UTC (9c5ef40)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -747,9 +747,9 @@
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
       is present and the fourth element is not present.
-       Lexical representations of language tags
-       MAY be case normalized,
-       (for example, by converting to lower case).
+      Lexical representations of language tags MAY be case normalized,
+      (for example, by canonicalizing as defined by 
+      <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
       </p>
 
     <p>A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
@@ -804,12 +804,19 @@
 
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
       Two literals are term-equal (the same <a>RDF literal</a>)
-      if and only if the two <a>lexical forms</a>,
-      the two <a>datatype IRIs</a>,
-      the two <a>language tags</a> (if any), and
-      the two <a>base directions</a> (if any) compare equal,
-      using <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
-      (see description of string comparison in <a href="#rdf-strings" class="sectionRef"></a>).
+      if and only if:
+      <ul>
+        <li>the two <a>lexical forms</a> compare equal</li>
+        <li>the two <a>datatype IRIs</a> compare equal</li>
+        <li>the two <a>language tags</a> (if any) compare equal</li>
+        <li>the two <a>base directions</a> (if any) compare equal</li>
+      </ul>
+      Comparison is performed using 
+      <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
+      (see description of string comparison in 
+      <a href="#rdf-strings" class="sectionRef"></a>)
+      except for language tags, where the comparison is performed using 
+      <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">ASCII case-insensitive matching</a>.
       Thus, two literals can have the same value
       without being the same <a>RDF term</a>.
       For example:</p>


spec/index.html
```diff
@@ -804,12 +804,19 @@ <h2>Literals</h2>
...
-      the two <a>base directions</a> (if any) compare equal,
-      using <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
-      (see description of string comparison in <a href="#rdf-strings" class="sectionRef"></a>).
+      if and only if:
+      <ol>
+        <li>the two <a>lexical forms</a> compare equal</li>
+        <li>the two <a>datatype IRIs</a> compare equal</li>
+        <li>the two <a>language tags</a> (if any) compare equal</li>
+        <li>the two <a>base directions</a> (if any) compare equal</li>
+      </ol>
```
TallTed, line 813: There's no reason the various comparisons need to be done in the order of this presentation, so an unordered list is more appropriate.
```suggestion
      <ul>
        <li>the two <a>lexical forms</a> compare equal</li>
        <li>the two <a>datatype IRIs</a> compare equal</li>
        <li>the two <a>language tags</a> (if any) compare equal</li>
        <li>the two <a>base directions</a> (if any) compare equal</li>
      </ul>
```

hartig (approved):

gkellogg (approved):
We might also consider updating the following sentence:

> A literal is a language-tagged string if the third element is present and the fourth element is not present. Lexical representations of language tags MAY be case normalized, (for example, by converting to lower case).

and instead say "(for example, by canonicalizing as defined by BCP47 [§4.5](https://www.rfc-editor.org/rfc/rfc5646.html#page-66))". We shouldn't be encouraging normalization to lower case.

afs:
> We might also consider updating the following sentence:

Change made.

TallTed:
Looks like [this ordered list](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/105.html#dfn-literal-value) is what [you refer to](https://github.com/w3c/rdf-concepts/pull/105#discussion_r1811138618)... as it appears to be the only ordered list in [the document previewed from here](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/105.html).

I suggest that also be changed to unordered (likewise the nested sub-list), as the order by which the items are evaluated doesn't appear to me to matter; they're all mutually exclusive.

afs:
> I suggest that also be changed to unordered (likewise the nested sub-list), as the order by which the items are evaluated doesn't appear to me to matter; they're all mutually exclusive.

@TallTed would you like to put in a PR for that? Thanks.

This PR is in response to a specific comment #100. I find it easier to make progress if PR's don't grow. It can become problematic with longer living and wider PRs.

TallTed (approved):

PR #109, opened 2024-10-24
Add a further explanation of `ltr` and `rtl` (based on ePub).

gkellogg:
Suggested by w3c/rdf-turtle#75.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/109.html" title="Last updated on Oct 29, 2024, 8:15 PM UTC (038ffc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/109/f0ce6fc...038ffc5.html" title="Last updated on Oct 29, 2024, 8:15 PM UTC (038ffc5)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -761,8 +761,8 @@
 
     <p>The meanings of the <a>base direction</a> values are:</p>
     <ul>
-      <li>`ltr`: indicates that the initial text direction is set to left-to-right text.</li>
-      <li>`rtl`: indicates that the initial text direction is set to right-to-left text.</li>
+      <li>`ltr`: indicates that the initial text direction is set to left-to-right.</li>
+      <li>`rtl`: indicates that the initial text direction is set to right-to-left.</li>
     </ul>
 
     <p>Please note that concrete syntaxes MAY support
--- a/spec/index.html
+++ b/spec/index.html
@@ -812,14 +812,15 @@
 
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
       Two literals are term-equal (the same <a>RDF literal</a>)
-      if and only if:
-      <ul>
-        <li>the two <a>lexical forms</a> compare equal</li>
-        <li>the two <a>datatype IRIs</a> compare equal</li>
-        <li>the two <a>language tags</a> (if any) compare equal</li>
-        <li>the two <a>base directions</a> (if any) compare equal</li>
-      </ul>
-      Comparison is performed using 
+      if and only if:</p>
+
+    <ul>
+      <li>the two <a>lexical forms</a> compare equal</li>
+      <li>the two <a>datatype IRIs</a> compare equal</li>
+      <li>the two <a>language tags</a> (if any) compare equal</li>
+      <li>the two <a>base directions</a> (if any) compare equal</li>
+    </ul>
+    <p>Comparison is performed using 
       <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
       (see description of string comparison in 
       <a href="#rdf-strings" class="sectionRef"></a>)
@@ -862,7 +863,7 @@
          can detect text direction from the initial character.
          For example "<bdi lang="ar">اَلأَعْشَى</bdi> - 1st place" displays the
          aarabic characters in a right-to-left fashion, while in memory
-         they would be stored as "<bdo lang="ar">اَلأَعْشَى</bdo> - 1st place"</p>
+         they would be stored as "<bdo dir="ltr" lang="ar">اَلأَعْشَى</bdo> - 1st place"</p>
     </section>
   </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -859,10 +859,10 @@
        </p>
        <div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
 
-       <p>In the absense of an explicit initial text direction, the [[?UAX9]] algorithm
+       <p>In the absence of an explicit initial text direction, the [[?UAX9]] algorithm
          can detect text direction from the initial character.
          For example "<bdi lang="ar">اَلأَعْشَى</bdi> - 1st place" displays the
-         aarabic characters in a right-to-left fashion, while in memory
+         Arabic characters in a right-to-left fashion, while in memory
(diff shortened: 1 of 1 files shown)


hartig (approved):

iherman (approved):

spec/index.html
```diff
@@ -757,6 +757,12 @@ <h2>Literals</h2>
       The third element, the language tag, is treated identically as in a <a>language-tagged string</a>,
       and the fourth element, <a>base direction</a>, MUST be either `ltr` or `rtl`, which MUST be in lower case.</p>
 
+    <p>The meanings of the <a>base direction</a> values are:</p>
+    <ul>
+      <li>`ltr`: indicates that the textual value is explicitly directionally set to left-to-right text.</li>
+      <li>`rtl`: indicates that the textual value is explicitly directionally set to right-to-left text.</li>
```
afs, line 763:
```suggestion
      <li>`ltr`: indicates that the initial text direction is set to left-to-right text.</li>
      <li>`ltr`: indicates that the initial text direction is set to right-to-left text.</li>
```

The lexical form may contain code points such as U+200E and U+200F that affect the direction while rendering. I'd avoid "textual value" because in RDF "value" relates to datatype; here RDF transports the information of initial text direction but the use of the information is when rendering: [non-normative section 3.3.1](https://www.w3.org/TR/rdf12-concepts/#section-text-direction).

gkellogg: Other than the second line being for `rtl` rather than `ltr`, I think this is a good change. I'm going to just fix that inline and accept the suggestion.

gkellogg:
Dear I18N people, I want to be sure that the recently added example for natively displaying Arabic RTL seems appropriate and correct.

spec/index.html
```diff
@@ -849,6 +856,12 @@ <h3>Initial Text Direction</h3>
          HTML's `dir` attribute) it can then be correctly presented as:
        </p>
        <div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
+
+       <p>In the absense of an explicit initial text direction, the [[?UAX9]] algorithm
```
gkellogg, line 860:
```suggestion
       <p>In the absence of an explicit initial text direction, the [[?UAX9]] algorithm
```

spec/index.html
```diff
@@ -849,6 +856,12 @@ <h3>Initial Text Direction</h3>
...
+         can detect text direction from the initial character.
+         For example "<bdi lang="ar">اَلأَعْشَى</bdi> - 1st place" displays the
+         aarabic characters in a right-to-left fashion, while in memory
```
gkellogg, line 863:
```suggestion
         Arabic characters in a right-to-left fashion, while in memory
```

spec/index.html
```diff
@@ -849,6 +856,12 @@ <h3>Initial Text Direction</h3>
          HTML's `dir` attribute) it can then be correctly presented as:
        </p>
        <div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
+
+       <p>In the absence of an explicit initial text direction, the [[?UAX9]] algorithm
+         can detect text direction from the initial character.
```
aphillips, line 861: UAX9 is the right underlying reference, but won't be clear in this context. Might I suggest linking to I18N-glossary's term [Unicode Bidirectional Algorithm](https://w3c.github.io/i18n-glossary/#dfn-unicode-bidi-algorithm) here instead? (Note that the term is exported, so you can just `[=link it=]` in respec.)

---

This definition isn't quite correct. I'm leery of specs such as RDF importing lots of material about bidi because of the complexity of the topic. I think it might be better overall to point back to I18N's materials on this in [Specdev](https://www.w3.org/TR/international-specs/#sec_dir_background), [String-Meta](https://www.w3.org/TR/string-meta/) and the [use cases article](https://www.w3.org/International/articles/lang-bidi-use-cases/).

Correcting this I might suggest:

```suggestion
       <p>In the absence of an explicit initial text direction,
       the [=Unicode Bidirectional Algorithm=] detects the text direction from the content.
       This depends on the first strongly directional character in the text
       or on the context.
       Users need to provide [=bidi isolation=] whenever text is inserted into a larger document
       in order to avoid [=spillover effects=].
```

spec/index.html
```diff
@@ -757,6 +757,12 @@ <h2>Literals</h2>
       The third element, the language tag, is treated identically as in a <a>language-tagged string</a>,
       and the fourth element, <a>base direction</a>, MUST be either `ltr` or `rtl`, which MUST be in lower case.</p>
 
+    <p>The meanings of the <a>base direction</a> values are:</p>
+    <ul>
+      <li>`ltr`: indicates that the initial text direction is set to left-to-right text.</li>
+      <li>`rtl`: indicates that the initial text direction is set to right-to-left text.</li>
```
aphillips, line 763: "... left-to-right text" is a bit redundant. Perhaps:

```suggestion
      <li>`ltr`: indicates that the initial text direction is set to left-to-right.</li>
      <li>`rtl`: indicates that the initial text direction is set to right-to-left.</li>
```

aphillips (changes requested):
A good start, but would suggest leaning on the I18N materials a bit harder.

afs (approved):

spec/index.html
```diff
@@ -849,6 +856,12 @@ <h3>Initial Text Direction</h3>
          HTML's `dir` attribute) it can then be correctly presented as:
        </p>
        <div lang="he" dir="rtl">פעילות הבינאום, W3C</div>
+
+       <p>In the absence of an explicit initial text direction, the [[?UAX9]] algorithm
+         can detect text direction from the initial character.
+         For example "<bdi lang="ar">اَلأَعْشَى</bdi> - 1st place" displays the
+         Arabic characters in a right-to-left fashion, while in memory
+         they would be stored as "<bdo dir="ltr" lang="ar">اَلأَعْشَى</bdo> - 1st place"</p>
```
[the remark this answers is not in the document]
gkellogg: I updated to use &quot;<bdi lang="he">ספרים בינלאומיים!</bdi>&quot;, which I hope is correct and reasonable.

aphillips (approved):

PR #114, opened 2024-11-06
rdf:JSON lexical-to-value updates

### LINKED ISSUE #87: rdf:JSON value space too liberal
The value space contains elements like the list that is its only member. There is nothing wrong with this, but I don't believe that such values are part of JSON and they certainly cause problems in serialization. If these values are not wanted a datatypes expert should be consulted to provide the best wording for excluding them.

My wording change would be to add "the smallest set containing" to the definition.

### LINKED ISSUE #88: base JSON value objects on unordered maps
The value space still contains orderd maps (with insignicant order).  It would be much better to start out with a map datatype that is unordered instead of going through all the contortions to try to eliminate all vestiges of the order.

https://en.wikipedia.org/wiki/Associative_array  appears to be a reasonable reference for an unordered finite map.

### LINKED ISSUE #89: underspecified lexical-to-value map for rdf:JSON
The lexical-to-value mapping does not specify the result of the map for Objects, Arrays, Strings, or Literal Names.  For example, the literal name true could be mapped to any of true, false, or nil.  For another example, there is no description of how JSON character escapes are handled.  The JSON string "\u1234" could be mapped to the value "A" and still be compliant with the lexical-to-value mapping.
email to Steve

### LINKED ISSUE #90: remove rdf:JSON if it doesn't meet the requirements of an RDF datatype
If rdf:JSON does not meet the requirements of an RDF datatype then it should be removed from RDF concepts.

### LINKED ISSUE #91: rdf:JSON equality of +0 and -0
Is +0 equal to -0 or not?  Equality of IEEE floating point double would indicate so but equality for rdf;JSON says they have to be the same which means that they are not equal.   If the latter is desired there should be a note along the lines of

> 0 and -0 compare equal in IEEE floating point double.  RDF, however, uses a notion of true equality for datatype values.  As 0 and -0 are different values in the value space of rdf:JSON they are not equal as far as rdf:JSON is concerned.

gkellogg:
Limit the value space for rdf:JSON to be the "smallest set containing ...". 

Updates rdf:JSON L2V mapping.

Fixes #87.Fixes #88. Fixes #89. Fixes #90. Fixes #91.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/114.html" title="Last updated on Nov 13, 2024, 10:31 PM UTC (1a4984b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/114/e24289d...1a4984b.html" title="Last updated on Nov 13, 2024, 10:31 PM UTC (1a4984b)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1617,10 +1617,13 @@
             <a data-cite="INFRA#map-entry">Map entries</a> are treated as being unordered.
           </li>
           <li>A <a data-cite="RFC8259#section-5">JSON Array</a> is mapped to a <a data-cite="INFRA#list">list</a>
-            such that this list contains as many elements as the JSON Array
-            and, for every position |i| in the array, the element at the |i|-th
-            position in the list is the value that results from applying this
-            mapping to the |i|-th element of the array.</li>
+            such that this <a data-cite="INFRA#list">list</a> contains as many
+            elements as the <a data-cite="RFC8259#section-5">JSON Array</a> and, for
+            every position |i| in the <a data-cite="RFC8259#section-5">array</a>,
+            the element at the |i|-th position in the
+            <a data-cite="INFRA#list">list</a> is the value that results from
+            applying this mapping to the |i|-th element of the
+            <a data-cite="RFC8259#section-5">array</a>.</li>
           <li>A <a data-cite="RFC8259#section-6">JSON Number</a> is mapped to
             an <a data-cite="XMLSCHEMA11-2#double"><strong>xsd:double</strong></a>
             using a method consistent with <a data-cite="XMLSCHEMA11-2#f-doubleLexmap">doubleLexicalMap</a>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -1609,19 +1609,19 @@
         literal value (<a data-cite="INFRA#boolean">`true`, `false`</a>, and <a data-cite="INFRA#nulls">null</a>).
 
         <ul>
-          <li>A <a href="RFC8259#section-4">JSON Object</a> is mapped to a <a data-cite="INFRA#ordered-map">map</a>
+          <li>A <a data-cite="RFC8259#section-4">JSON Object</a> is mapped to a <a data-cite="INFRA#ordered-map">map</a>
             by transforming each object member into a <a data-cite="INFRA#map-entry">map entry</a>
             with the <a data-cite="INFRA#map-key">key</a> taken from the member name and
             <a data-cite="INFRA#map-value">value</a> taken by performing this mapping
             to the member value.
             <a data-cite="INFRA#map-entry">Map entries</a> are treated as being unordered.
           </li>
-          <li>A <a href="RFC8259#section-5">JSON Array</a> is mapped to a <a data-cite="INFRA#list">list</a>
+          <li>A <a data-cite="RFC8259#section-5">JSON Array</a> is mapped to a <a data-cite="INFRA#list">list</a>
             such that this list contains as many elements as the JSON Array
             and, for every position |i| in the array, the element at the |i|-th
             position in the list is the value that results from applying this
             mapping to the |i|-th element of the array.</li>
-          <li>A <a href="RFC8259#section-6">JSON Number</a> is mapped to
+          <li>A <a data-cite="RFC8259#section-6">JSON Number</a> is mapped to
             an <a data-cite="XMLSCHEMA11-2#double"><strong>xsd:double</strong></a>
             using a method consistent with <a data-cite="XMLSCHEMA11-2#f-doubleLexmap">doubleLexicalMap</a>.
             <div class="note">Some numbers cannot be represented
@@ -1629,10 +1629,10 @@
               and may map to `+INF` or `-INF`.
               Such values cannot be represented as JSON Numbers, limiting the ability to serialize such values back to JSON.</div>
           </li>
-          <li>A <a href="RFC8259#section-7">JSON String</a> is mapped to
+          <li>A <a data-cite="RFC8259#section-7">JSON String</a> is mapped to
             a <a>string</a> after converting any escape sequences to the associated
             <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>.</li>
-          <li>A <a href="RFC8259#section-3">JSON literal name</a> maps
+          <li>A <a data-cite="RFC8259#section-3">JSON literal name</a> maps
             `true` to <a data-cite="INFRA#boolean">`true`</a>,
             `false` to <a data-cite="INFRA#boolean">`false`</a>, and
             `null` to <a data-cite="INFRA#nulls">`null`</a>.</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1617,8 +1617,10 @@
             <a data-cite="INFRA#map-entry">Map entries</a> are treated as being unordered.
           </li>
           <li>A <a href="RFC8259#section-5">JSON Array</a> is mapped to a <a data-cite="INFRA#list">list</a>
-            by performing this mapping on each array value,
-            and adding that mapped value at the same index within the <a data-cite="INFRA#list">list</a>.</li>
+            such that this list contains as many elements as the JSON Array
+            and, for every position |i| in the array, the element at the |i|-th
+            position in the list is the value that results from applying this
+            mapping to the |i|-th element of the array.</li>
           <li>A <a href="RFC8259#section-6">JSON Number</a> is mapped to
             an <a data-cite="XMLSCHEMA11-2#double"><strong>xsd:double</strong></a>
             using a method consistent with <a data-cite="XMLSCHEMA11-2#f-doubleLexmap">doubleLexicalMap</a>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -1636,9 +1636,10 @@
             a <a>string</a> after converting any escape sequences to the associated
             <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>.</li>
           <li>A <a data-cite="RFC8259#section-3">JSON literal name</a> maps
-            `true` to <a data-cite="INFRA#boolean">`true`</a>,
+            JSON `true`, `false`, and `null` values
+            to [[INFRA]] <a data-cite="INFRA#boolean">`true`</a>,
             `false` to <a data-cite="INFRA#boolean">`false`</a>, and
-            `null` to <a data-cite="INFRA#nulls">`null`</a>.</li>
+            `null` to <a data-cite="INFRA#nulls">`null`</a> values, respectively.</li>
         </ul>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1561,7 +1561,7 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
         </div></dd>
 
       <dt id="JSON-value-space">The <a>value space</a></dt>
-      <dd>is the set of
+      <dd>is the set the smallest set containing
```
gkellogg, line 1564:
```suggestion
      <dd>is the smallest set containing
```

hartig (approved):

TallTed:
Please retitle from `rdf:JSON L2V` to `rdf:JSON lexical-to-value` or similar.

pfps:
Looks good.  It may even actually satisfy the requirements of an RDF datatype. (I need to see whether there are problems with unordered ordered maps and how identity is defined.  It looks so far as if there is a problem here.)

I note that the wording is clear that +0.0 and -0.0 are separate elements of the value space.

In any case, this is much better and should be merged.

spec/index.html
```diff
@@ -1623,9 +1630,12 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
...
           <li>A <a href="RFC8259#section-7">JSON String</a> is mapped to
-            a <a>string</a>.</li>
-          <li>A <a href="RFC8259#section-3">JSON literal name</a> is mapped to
-            <a data-cite="INFRA#boolean">`true`, `false`</a>, or <a data-cite="INFRA#nulls">null</a>.</li>
+            a <a>string</a> after converting any escape sequences to the associated
+            <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>.</li>
+          <li>A <a href="RFC8259#section-3">JSON literal name</a> maps
+            `true` to <a data-cite="INFRA#boolean">`true`</a>,
+            `false` to <a data-cite="INFRA#boolean">`false`</a>, and
+            `null` to <a data-cite="INFRA#nulls">`null`</a>.</li>
```
TallTed, line 1638: I wonder whether there should be some difference in the visible target text? Something like this, perhaps?
```suggestion
            `true` to <a data-cite="INFRA#boolean">Infra boolean `true`</a>,
            `false` to <a data-cite="INFRA#boolean">Infra boolean `false`</a>,
            and `null` to <a data-cite="INFRA#nulls">Infra `null`</a>.</li>
```

spec/index.html
```diff
@@ -1611,9 +1615,12 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
...
             to the member value.
             <a data-cite="INFRA#map-entry">Map entries</a> are treated as being unordered.
-            </li>
+          </li>
           <li>A <a href="RFC8259#section-5">JSON Array</a> is mapped to a <a data-cite="INFRA#list">list</a>
-            by performing this mapping on each array value.</li>
+            such that this list contains as many elements as the JSON Array
+            and, for every position |i| in the array, the element at the |i|-th
+            position in the list is the value that results from applying this
+            mapping to the |i|-th element of the array.</li>
```
TallTed, line 1623:
```suggestion
            such that this <a data-cite="INFRA#list">list</a> contains as many
            elements as the <a href="RFC8259#section-5">JSON Array</a> and, for
            every position |i| in the <a href="RFC8259#section-5">array</a>,
            the element at the |i|-th position in the
            <a data-cite="INFRA#list">list</a> is the value that results from
            applying this mapping to the |i|-th element of the
            <a href="RFC8259#section-5">array</a>.</li>
```

spec/index.html
gkellogg: Maybe something like the following:

A JSON literal name maps the values `true`, `false, and `null` to <a data-cite="INFRA#boolean">`true`</a>, <a data-cite="INFRA#boolean">`false`</a>, and <a data-cite="INFRA#nulls">`null`</a> from [[INFRA]], respectively.

TallTed (approved):

pfps:
A problem with the proposed wording is that

maps (mapping strings to values in the value space where the order of map entries is not significant),

is not very well defined.  And worse, equality is defined with 

They are both maps with equal entries – meaning that both a and b have the same size, and for each entry ea in a there is an entry eb in b such that the key in ea equals the key in eb, and the value in ea equals the value in eb.

which looks as if elements of the value space are ordered maps and equality is defined over this notion of identity.  This is not suitable for an RDF value space.  


Instead it would be much better to start with unordered maps, defined as

finite maps from strings to values in the value space

without introducing any notion of order.  If a referent for "map" is needed point to some mathematical definition of maps.

Then make the NOTE be something like

NOTE
The lexical-to-value mapping the JSON lexical values { "a": 1, "b": 2 } and { "b": 2, "a": 1 } is the same map, as the order in the lexical values is not significant.

spec/index.html
```diff
@@ -1605,27 +1609,36 @@ <h3>The <code>rdf:JSON</code> Datatype</h3>
...
+            such that this <a data-cite="INFRA#list">list</a> contains as many
+            elements as the <a data-cite="RFC8259#section-5">JSON Array</a> and, for
+            every position |i| in the <a data-cite="RFC8259#section-5">array</a>,
+            the element at the |i|-th position in the
+            <a data-cite="INFRA#list">list</a> is the value that results from
+            applying this mapping to the |i|-th element of the
+            <a data-cite="RFC8259#section-5">array</a>.</li>
+          <li>A <a data-cite="RFC8259#section-6">JSON Number</a> is mapped to
             an <a data-cite="XMLSCHEMA11-2#double"><strong>xsd:double</strong></a>
             using a method consistent with <a data-cite="XMLSCHEMA11-2#f-doubleLexmap">doubleLexicalMap</a>.
```
pfps, line 1629: I would add "strictly" here and a note saying that the rounding algorithm there MUST be used.  The reason for these additions is to trump the wording at the end of the box saying.  "Note: This specification permits the substitution of any other rounding algorithm which conforms to the requirements of [[IEEE 754-2008]](https://www.w3.org/TR/xmlschema11-2/#ieee754-2008)."

gkellogg:
> maps (mapping strings to values in the value space where the order of map entries is not significant),

We could just drop the "order of map entries is not significant" as it has no operational effect. In INFRA, a _map_ is defined as "a finite ordered sequence of [tuples](https://infra.spec.whatwg.org/#tuple), each consisting of a key and a value, with no key appearing twice. Each such tuple is called an entry." In RFC8259, a JSON object is defined as "zero or more name/value pairs (or members)". We could instead say:

  [maps](https://infra.spec.whatwg.org/#ordered-map) (mapping JSON Object members to [map entries](https://infra.spec.whatwg.org/#map-entry) in a [map](https://infra.spec.whatwg.org/#ordered-map)).

> They are both maps with equal entries – meaning that both a and b have the same size, and for each entry ea in a there is an entry eb in b such that the key in ea equals the key in eb, and the value in ea equals the value in eb.

> which looks as if elements of the value space are ordered maps and equality is defined over this notion of identity. This is not suitable for an RDF value space.

I don't read it that way, but we could adopt some of the wording for lists:

> They are both [maps](https://infra.spec.whatwg.org/#ordered-map) with equal [entries](https://infra.spec.whatwg.org/#map-entry) – meaning that for each entry <var>e<sub>a</sub><var> in <var>a</var> there exists an entry <var>e<sub>b</sub></var> in <var>b</var> such that the [key](https://infra.spec.whatwg.org/#map-key) in <var>e<sub>a</sub><var> equals the [key](https://infra.spec.whatwg.org/#map-key) in <var>e<sub>b</sub><var>, and the [value](https://infra.spec.whatwg.org/#map-value) in <var>e<sub>a</sub><var> equals the [value](https://infra.spec.whatwg.org/#map-value) in <var>e<sub>b</sub><var>, and both <var>a</var> and <var>b</var> have the same [size](https://infra.spec.whatwg.org/#map-size).

While I don't think it's useful to define another kind of map, we already have such a note in place.

pfps:
The point is that the value space has to be defined carefully and correctly, so that elements of the value space are just maps, not ordered maps.  The best way to do that is to *not* start with ordered maps.

spec/index.html
TallTed: Does this work?
```suggestion
            JSON `true`, `false`, and `null` values
            to [[INFRA]] `true`, `false`, and `null` values,
            respectively.</li>
```

gkellogg: That's what I was thinking, other than having the `true`, `false`, and `null` values from INFRA cited as is the current practice.

gkellogg:
> The point is that the value space has to be defined carefully and correctly, so that elements of the value space are just maps, not ordered maps. The best way to do that is to _not_ start with ordered maps.

If the mapping from JSON to INFRA was unique to RDF then there might be an argument to do something different, but INFRA is largely there, in large part, to create a data model for JSON, and is used widely in W3C specs. Both EPub and DID Core use it specifically for representing JSON-LD. If we were to go on our own here, we would be going counter to established patterns at W3C, which could cause problems during Horizontal Review.

The [DID Core Data Model](https://www.w3.org/TR/did-core/#data-model) specifically calls out the fact that INFRA orders things, even though order is not significant:

> As a result of the [data model](https://www.w3.org/TR/did-core/#data-model) being defined using terminology from [[INFRA](https://www.w3.org/TR/did-core/#bib-infra)], property values which can contain more than one item, such as [lists](https://infra.spec.whatwg.org/#list), [maps](https://infra.spec.whatwg.org/#ordered-map) and [sets](https://infra.spec.whatwg.org/#ordered-set), are explicitly ordered. All list-like value structures in [[INFRA](https://www.w3.org/TR/did-core/#bib-infra)] are ordered, whether or not that order is significant. For the purposes of this specification, unless otherwise stated, [map](https://infra.spec.whatwg.org/#ordered-map) and [set](https://infra.spec.whatwg.org/#ordered-set) ordering is not important and implementations are not expected to produce or consume deterministically ordered values.

In any case, the fact that the definition of a map defines order none of our use is dependent on that. Perhaps adding some language to the value space note on maps based on what DID described would help further clarify this.

spec/index.html
gkellogg:
```suggestion
            JSON `true`, `false`, and `null` values
            to [[INFRA]] <a data-cite="INFRA#boolean">`true`</a>,
            `false` to <a data-cite="INFRA#boolean">`false`</a>, and
            `null` to <a data-cite="INFRA#nulls">`null`</a> values, respectively.</li>
```

spec/index.html
gkellogg:
```suggestion
            using a method consistent with <a data-cite="XMLSCHEMA11-2#f-doubleLexmap">doubleLexicalMap</a>,
            which MUST strictly conform to the rounding method described in
            <a data-cite="XMLSCHEMA11-2#f-floatPtRound">floatPtRound</a>.
```

PR #115, opened 2024-11-13
add section about 'unstar' mapping

pchampin:
as per https://github.com/w3c/rdf-star-wg/issues/129

[Preview with the examples working](https://raw.githack.com/w3c/rdf-concepts/refs/heads/action-129/spec/index.html)
(as opposed to the automatic preview below)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html" title="Last updated on Feb 20, 2025, 6:38 PM UTC (76cd23c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/115/536b67e...76cd23c.html" title="Last updated on Feb 20, 2025, 6:38 PM UTC (76cd23c)">Diff</a>


afs:
Is this "as well as" the graph algorithm or "instead of"?

https://github.com/w3c/rdf-star-wg/issues/114#issuecomment-2377595982

niklasl:
This approach requires graphs containing triple terms to be represented as datasets. That excludes cases where you need to put "unstarred" RDF 1.2 graphs into an RDF 1.1-based quad store and manage them within specific named graphs. Implementations supporting the [default graph union](https://www.w3.org/TR/sparql11-service-description/#sd-uniondefaultgraph) mechanism would also treat the "triple term graphs" as asserted in that union graph.

niklasl:
It has been mentioned (or opined) that the "star" name would eventually go away (it would be just RDF 1.2 with triple terms). If so, perhaps "unstar" is an unfortunate name for future reference?

rat10:
There seems to be an issue with the [examples section](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#section-unstar-example): all examples say "Cannot GET /uploads/dcqFS6/spec/ex-unstar-output.trig".

gkellogg:
> There seems to be an issue with the [examples section](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#section-unstar-example): all examples say "Cannot GET /uploads/dcqFS6/spec/ex-unstar-output.trig".

It's the general PR-preview issue of not being able to retrieve neighboring resources. They are fleshed out if you look at the [GitHack](https://raw.githack.com/w3c/rdf-concepts/8e07ca3dd592b9b32049ace3e3e0995bf9d1a0b1/spec/index.html#section-unstar-example) version.

pchampin:
> Is this proposed "as well as" the graph-to-graph algorithm or "instead of"?

I personally don't think that we should have multiple such mappings, and I am more and more convinced that the graph-to-graph approach makes more sense.

The reasons I stuck to my initial graph-to-dataset approach in this PR are that
* I am still not clear about the details of the graph-to-graph approach, and
* I wanted to write down the design goals of the mapping, so that we can discuss them (and inform the responses to the previous point).

rat10:
What is the "graph-to-graph" algorithm? A mapping based on the RDF standard reification vocabulary?

niklasl:
> What is the "graph-to-graph" algorithm? A mapping based on the RDF standard reification vocabulary?

Or something isomorphic to it but using dedicated terms. There's an example of that in [this recent wiki page](https://github.com/w3c/rdf-star-wg/wiki/Converting-RDF-1.2-data-to-RDF-1.1) (with links to https://github.com/w3c/rdf-semantics/issues/49 and https://github.com/w3c/rdf-star-wg/issues/114.).

afs:
PR goal:
(content from the PR)

* Information preserving
  It must be possible to reconstruct the input dataset from the output dataset.
  Note that, on the other hand, the algorithm is not designed to be semantics preserving:
  the graphs in the produced dataset are not semantically <a>equivalent</a> to their corresponding graph in the input dataset.
* Idempotent
  Transforming a dataset that is already complying with RDF Classic (i.e. containing no <a>triple term</a>) must result in the same dataset.
* Universal
  It should be possible to transform any RDF Full dataset using this method.

pchampin:
I just updated the PR; the new algorithm is "graph-to-graph", repurposing the reification vocabulary.
Note that I coined `rdf:UnstarredTripleTerm` to type the generated blank nodes.

Note that I deliberately chose a very specific name for this, to distinguish it from the type we will probably introduce as the class of all triple terms, for example, to describe the range of rdf:reifies (say, `rdf:Triple`). Indeed, I believe that there will be valid use-cases to use that type (`rdf:Triple`) in Fuill RDF graphs, while `rdf:UnstarredTripleTerm` should really be considered as "reserved" and not to be used outside of the "unstar" algorithm.

afs:
This a suggestion relate to presentation only.

RDF-concepts defines RDF and is also a readable document.

The algorithms of unstar/restar are good for defining the translation but algorithms do not communicate the broad intent so well. 

Maybe: have all the written description, discussion and examples, then have the algorithms.

* Put the overview "The general principle" at the beginning of section 8
* Pull the examples into section 8.
* Either push all algorithms to later sections within section 8 or put the algorithms as normative appendixes.

rat10:
> > There seems to be an issue with the [examples section](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#section-unstar-example): all examples say "Cannot GET /uploads/dcqFS6/spec/ex-unstar-output.trig".
> 
> It's the general PR-preview issue of not being able to retrieve neighboring resources. They are fleshed out if you look at the [GitHack](https://raw.githack.com/w3c/rdf-concepts/8e07ca3dd592b9b32049ace3e3e0995bf9d1a0b1/spec/index.html#section-unstar-example) version.

@gkellogg  How did you create that link? I'm trying to read the new PR but again am unable to read the included examples.

gkellogg:
> > > There seems to be an issue with the [examples section](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#section-unstar-example): all examples say "Cannot GET /uploads/dcqFS6/spec/ex-unstar-output.trig".
> > 
> > 
> > It's the general PR-preview issue of not being able to retrieve neighboring resources. They are fleshed out if you look at the [GitHack](https://raw.githack.com/w3c/rdf-concepts/8e07ca3dd592b9b32049ace3e3e0995bf9d1a0b1/spec/index.html#section-unstar-example) version.
> 
> @gkellogg How did you create that link? I'm trying to read the new PR but again am unable to read the included examples.

Select spec/index.html in the branch you want to see and enter it in http://raw.githack.com/. It gives you a link to the rendered version.

niklasl:
I think this approach is good.

The use of a dedicated type for unstarred triple terms seems prudent. If used, an `rdf:UnstarredTripleTerm rdfs:subClassOf rdf:TripleTerm` axiom should be defined (possibly where https://github.com/w3c/rdf-semantics/issues/49 is defined, if it will be). This since the unstarred form is useful as input to e.g. OWL reasoners without full RDF 1.2 support, but users of that should only rely on the `rdf:TripleTerm` type, not this special subclass. This is because when such tooling is updated to RDF 1.2, the type of triple terms will just be `rdf:TripleTerm`, and any OWL-based axioms should still work.

Whether the constituent triple term predicates should be reused from the reification vocabulary or not has been debated some (see e.g. https://github.com/w3c/rdf-semantics/issues/49#issuecomment-2422786852). It depends on whether or not it makes sense to make `rdf:TripleTerm` a subclass of `rdf:Statement`.

I think the name `rdf:TripleTerm` has been used most recently, but it has perhaps not been finalized  (I am somewhat in favor of `rdf:Triple` if there is room for debate).

But these details, about naming and which constituent triple term predicates to use, can probably be dealt with separately, to avoid blocking this PR.

The question about whether to use name "unstar" at all remains (as "RDF-star" is not mentioned as such in RDF 1.2; only in reference to the RDF-star WG).

rat10:
> I just updated the PR; the new algorithm is "graph-to-graph", repurposing the reification vocabulary. Note that I coined `rdf:UnstarredTripleTerm` to type the generated blank nodes.
> 
> Note that I deliberately chose a very specific name for this, to distinguish it from the type we will probably introduce as the class of all triple terms, for example, to describe the range of rdf:reifies (say, `rdf:Triple`). Indeed, I believe that there will be valid use-cases to use that type (`rdf:Triple`) in Fuill RDF graphs, while `rdf:UnstarredTripleTerm` should really be considered as "reserved" and not to be used outside of the "unstar" algorithm.

To me it seems like the discussions in the Semantics TF and in Github issues, e.g. [rdf-semantics issue #49 ](https://github.com/w3c/rdf-semantics/issues/49) and [rdf-star-wg issue #130](https://github.com/w3c/rdf-star-wg/issues/130), moved away from re-using the RDF reification vocabulary. The reason is a bit intricate: the RDF standard reification describes an occurrence/instance/reification of a triple. The triple term describes a triple and only the reification, indicated by `rdf:reifies`, creates a reference to the instance/occurrence/reification. That means that in the following example `_:r` and `_:s` are semantically equivalent, but the triple term and the reification quad are not.
```
_:r rdf:reifies <<( :s :p :o )>> .
_:s a rdf:Statement ;
    rdf:subject :s ;
    rdf:predicate :p ;
    rdf:object :o .
```
Given this difference in meaning I think it's more prudent to not reuse the properties from the reification vocabulary but to mint new ones, like e.g. `rdf:tripleTermSubject`, etc.

W.r.t. other wordings:
- how about `rdf:unTripleTerm` and `rdf:reTripleTerm`?
- I favor `rdf:TripleTerm` over `rdf:Triple` to refer to a triple term, since the latter could be easily misunderstood to refer to regular, asserted RDF triples.

pchampin:
Regarding the name of the algorithm,

I guess we could go for "classicize" (as it converts to RDF "classic").
"flatten" would also seem appropriate, but could create confusion with the [algorithm of the same name](https://www.w3.org/TR/json-ld-api/#flattening-algorithms) in JSON-LD? The context (no pun intended) is quite different so I'm not convinced this would be a real issue.

rat10:
> Regarding the vocabulary,
> 
> I think that duplicating the properties `rdf:subject`, `rdf:predicate`, `rdf:object` could also create a lot of confusion, so I would refrain from doing that unless repurposing them really breaks something badly. But since `rdf:Statement` is so loosely defined, I don't think that would be the case. I am happy to consider that, in retrospect, `rdf:Statement` can include both the "platonic triples" denoted by triple terms and the "occurrences" denoted by reifiers.

The WG seems to tend towards allowing the usage of triple terms in a more general way than just as a source of reifiers. In that context it is probably prudent to differentiate the two concepts, and not mix and mingle them. Also, I don't see how `rdf:Statement`is loosely defined. The naming may be a bit vague, but the definition IMO is not.

rat10:
> > Regarding the vocabulary,
> > I think that duplicating the properties `rdf:subject`, `rdf:predicate`, `rdf:object` could also create a lot of confusion, so I would refrain from doing that unless repurposing them really breaks something badly. But since `rdf:Statement` is so loosely defined, I don't think that would be the case. I am happy to consider that, in retrospect, `rdf:Statement` can include both the "platonic triples" denoted by triple terms and the "occurrences" denoted by reifiers.
> 
> The WG seems to tend towards allowing the usage of triple terms in a more general way than just as a source of reifiers. In that context it is probably prudent to differentiate the two concepts, and not mix and mingle them. Also, I don't see how `rdf:Statement` is loosely defined. The naming may be a bit vague, but the definition IMO is not.

On second (or rather n-th) thought I'd like to add something. 

Following recent discussions in the Semantics TF (as captured [here](https://github.com/w3c/rdf-star-wg/wiki/RDF-star-%22liberal-baseline%22) and discussed [there](https://github.com/w3c/rdf-star-wg/issues/138#issuecomment-2531294895)) we might settle for calling `rdf:Statement` the act of stating a triple (without of course saying if that stating actually happened, and where), and calling `rdf:Proposition` the abstract triple as described by an RDF-star triple term. Both are composed of subject, predicate and object, so ...I again tend to agree with you. . 

However, doesn't this jeopardize backwards compatability? So far it's possible to infer that an entity is of type `rdf:Statement` if it's the subject of an `rdf:subject|predicate|object` statement. From anecdotal evidence I gather that it's common practice to omit the type declaration and reduce triple count by one when using the RDF standard reification vocabulary. That practice becomes unsound as soon as unstar-ed triple terms enter the mix.

To work around this problem, we could stress that any application of the unstar mapping should refrain from applying the same optimization. Also, an unassuming RDF 1.1 environment would not be led into a completely wrong direction if it assumed that the immediate subject of an unstar operation - `_:gen1` in your proposed [Example 3](https://rawcdn.githack.com/w3c/rdf-concepts/f5d2deb28662ba8d520ebc38efb143625bddcaac/spec/index.html#ex-unstar-output) - represents an RDF standard reification. It doesn't since it would miss multi-part reifications, but maybe it's close enough.

On the other hand, backwards compatability is what the unstar mapping is all about, so why jeopardize it?

rat10:
> > To work around this problem, we could stress that any application of the unstar mapping should refrain from applying the same optimization.
> 
> Applying the `unstar` mapping means following the algorithm. If you don't follow it exactly, in particular if you don't add the triple (b, rdf:type, rdf:UnstarredTripleTerm), then it is not the `unstar` mapping anymore, and you may lose the properties listed at the top of Section 8 ("information preserving", "idempotent", "universal") can not be guaranteed...

Fair enough.

> > However, doesn't this jeopardize backwards compatability?
> 
> I do not suggest to change the domain of `rdf:subject` and co.... That would indeed jeopardize backwards compatibility.
> 
> So yes, it would mean that any bnode generated by the unstar mapping could be inferred to be of type `rdf:Statement`, but I don't think this is a problem. The important thing is that it is _also_ of type `rdf:UnstarredTripleTerm`, which is what should matter in that situation.

I discussed the "unassuming RDF 1.1 environment", and that is what matters to backwards compatability. In such an environment, `rdf:subject` etc would be assumed to refer to an RDF standard reification - which is exactly not what we want. Such an environment would probably not even be aware of the existence of the type `rdf:UnstarredTripleTerm`, let alone check for it. 

> Regarding the notion of `rdf:Proposition` discussed by the Semantics TF: I don't think that the `rdf:UnstarTripleTerm` class (used to _encode_ triple terms) needs to be related to `rdf:Proposition` (used to define the semantics of triple terms). But even if I did: according to WordReference, [several definitions of the term](https://www.wordreference.com/definition/proposition#unabridged) (and all of those related to logic or mathematics) define it as "a statement ...". This makes "proposition" a subclass of "statement", which is fine by me, and allows us to use `rdf:subject` and friends to describe instances of `rdf:Proposition`.

I disagree. No matter what the term "statement" means, the term `rdf:Statement` has a very specific meaning, and an RDF-star triple term is not a specialization of it, but arguably rather the other way round.

pchampin:
> No matter what the term "statement" means, the term rdf:Statement has a very specific meaning, and an RDF-star triple term is not a specialization of it, 

I stand corrected; re-reading the related sections [in RDF Semantics](https://www.w3.org/TR/rdf11-semantics/#Reif), it says "The subject of a reification is intended to refer to a concrete realization of an RDF triple, such as a document in a surface syntax, **rather than a triple considered as an abstract object**" (emphasis is mine). I will change my PR accordingly.

rat10:
W.r.t. vocabulary I’m still not convinced: defining a new `classicize` namespace creates its own issues. I find this more confusing then defining specialized properties like `rdf:tripleTermSubject`, etc. Also, doesn't this introduces not only 3 new predicates, but also a `classicize:TripleTerm` in addition to an `rdf:TripleTerm`?

rat10:
On a more basic level, I don’t think the proposal accomplishes all that it could and should accomplish. I would like the mapping to transform RDF 1.2 Full into a semantically equivalent RDF 1.1. IMO that is possible and desirable. The following is a sketch of a possible approach, extending and slightly redesigning the proposed PR.

**1: Triple Terms**
Of course there is no equivalent to the triple term in RDF 1.1, since RDF 1.1 doesn’t cater for the concept of an abstract and unasserted triple. But it can be described with standard triples, using to be defined triple term vocabulary. E.g. a triple term `<<( :tts :ttp :tto )>>` can be described as:
```
_:tt a rdf:TripleTermRepresentation ;
     rdf:tripleTermSubject :tts ;
     rdf:tripleTermPredicate :ttp ;
     rdf:tripleTermObject :tto .
```
Consequently an RDF 1.2 reification can be described with standard triples. Take the following RDF 1.2 Full construct:
```
_:r rdf:reifies <<( :tts :ttp :tto )>> ,
                <<( :tts2 :ttp2 :tto2 )>> .
```
Using the triple term vocabulary from above it can be mapped to standard RDF triples as follows:
```
_:r rdf:reifies _:tt ;
                _:tt2 .
_:tt a rdf:TripleTermRepresentation ;
     rdf:tripleTermSubject :tts ;
     rdf:tripleTermPredicate :ttp ;
     rdf:tripleTermObject :tto .
_:tt2 a rdf:TripleTermRepresentation ;
     rdf:tripleTermSubject :tts2 ;
     rdf:tripleTermPredicate :ttp2 ;
     rdf:tripleTermObject :tto2 .
```
That is, modulo vocabulary design, not much different from the proposed PR. The proposed PR also adds some useful comments that are not repeated here.

**2: Reifiers**
There is also no direct correspondent to a reifier in RDF 1.1, since RDF 1.1 reification has no concept of a group of reified statements, i.e it doesn’t support many-to-many reifications. But again this can be described with standard triples. Only a new property `rdf:reifiesGroupOfStatements` is needed to bridge the gap between RDF 1.1 many-to-one and RDF 1.2 many-to-many reification. Take the following RDF 1.2 Full construct:
```
_:r rdf:reifies <<( :tts :ttp :tto )>> ,
                <<( :tts2 :ttp2 :tto2 )>> .
```
Here `_:r` is semantically equivalent to the following RDF 1.1 construct:
```
_:rs rdf:reifiesGroupOfStatements _:sr1 ,
                                  _:sr2 .
_:sr1 a rdf:Statement ;
      rdf:subject :tts ;
      rdf:predicate :ttp ;
      rdf:object :tto .
_:sr2 a rdf:Statement ;
      rdf:subject :tts2 ;
      rdf:predicate :ttp2 ;
      rdf:object :tto2 .
```

**Semantic Equivalence**
Note that an single triple term reification is semantically equivalent to a `rdf:reifiesGroupOfStatements` composed of only one RDF 1.1 standard reification quad and to RDF 1.1 standard reification, i.e.
```
_:r rdf:reifies <<( :tts :ttp :tto )>> .
_:rs rdf:reifiesGroupOfStatements _:sr1 .
_:sr1 a rdf:Statement ;
      rdf:subject :tts ;
      rdf:predicate :ttp ;
      rdf:object :tto .
_:r owl:sameAs  _:rs ;
                _:sr1 .
```
IMO it is very useful to expressly state that this is the point where the two approaches meet.

**In conclusion** these mappings require the following addition to the RDF vocabulary:
```
rdf:TripleTermRepresentation
rdf:tripleTermSubject
rdf:tripleTermPredicate
rdf:tripleTermObject
rdf:reifiesGroupOfStatements
```
I think that is a reasonable expenditure. 

If I'm not mistaken a mapping algorithm based on this vocabulary is trivial, supports round-tripping and is idempotent. The sketched approach should even allow to mix both representations without any loss of information - that is of course not desirable, but can occur when integrating data, and would be manageable as an intermediary state.

pchampin:
> W.r.t. vocabulary I’m still not convinced: defining a new classicize namespace creates its own issues. 

I hesitated to do that, but since these terms are not meant to be used outside of the classicize algorithm, then it seemed better to isolate them in their own vocabulary rather than "pollute"  the `rdf:` vocabulary with terms that have a very specific purpose.

pchampin:
>  I would like the mapping to transform RDF 1.2 Full into a semantically equivalent RDF 1.1

In what level of entailment would you like this equivalent to stand? Simple entailment? RDF entailment? RDFS entailment? A new entailment level?

Also, the mapping does not suffice to encode RDF 1.2 Full into RDF 1.1. You would also need to transform literals with a base description. While there are [different](https://www.w3.org/TR/json-ld11/#the-i18n-namespace) [solutions](https://www.w3.org/TR/json-ld11/#the-rdf-compoundliteral-class-and-the-rdf-language-and-rdf-direction-properties) to do that, this would make the mapping (and the corresponding entailment regime) more complex.

rat10:
> > W.r.t. vocabulary I’m still not convinced: defining a new classicize namespace creates its own issues.
> 
> I hesitated to do that, but since these terms are not meant to be used outside of the classicize algorithm, then it seemed better to isolate them in their own vocabulary rather than "pollute" the `rdf:` vocabulary with terms that have a very specific purpose.

The classicize algorithm is meant to transform triple terms into something that RDF 1.2 Classic and possibly even older RDF environments can understand. So the triples using those new properties will be part of classic/older data. The application of the algorithm may be well isolated from classic/older data, but the result is not.

rat10:
> > I would like the mapping to transform RDF 1.2 Full into a semantically equivalent RDF 1.1
> 
> In what level of entailment would you like this equivalent to stand? Simple entailment? RDF entailment? RDFS entailment? A new entailment level?

I've illustrated the level of semantics equivalence that IMO is desirable with the [last example](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2539045176), namely that `_:r owl:sameAs  _:rs ; _:sr1 .`

> Also, the mapping does not suffice to encode RDF 1.2 Full into RDF 1.1. You would also need to transform literals with a base description. While there are [different](https://www.w3.org/TR/json-ld11/#the-i18n-namespace) [solutions](https://www.w3.org/TR/json-ld11/#the-rdf-compoundliteral-class-and-the-rdf-language-and-rdf-direction-properties) to do that, this would make the mapping (and the corresponding entailment regime) more complex.

I don't see how that problem is related to the topic of this discussion.

pchampin:
> The classicize algorithm is meant to transform triple terms into something that RDF 1.2 Classic and possibly even older RDF environments can understand. So the triples using those new properties will be part of classic/older data. The application of the algorithm may be well isolated from classic/older data, but the result is not.

I suggest we defer this discussion to later, and not block this PR on this basis. If we agree on the general algorithms, it is easy to change the vocabulary later. I just added an issue marking the `classicize:` namespace as AT RISK to force us rediscuss that (c054000). Would that work for you?

> I've illustrated the level of semantics equivalence that IMO is desirable with the https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2539045176, namely that _:r owl:sameAs  _:rs ; _:sr1 .

If you relly on the semantics of `owl:sameAs` to define this equivalence, then clearly it only exists in entailment regimes beyond those defined by our working group, right? Indeed, I don't aim to *prevent* anyone from defining this kind of equivalenve in a semantic extension, but I don't think we should bake that in the regimes we define ourselves. I also added a sentence to that effect (ec8663c).

With these two caveat, @rat10, would you agree to merge the PR and continue the discussion in further issues or PRs?

rat10:
> > The classicize algorithm is meant to transform triple terms into something that RDF 1.2 Classic and possibly even older RDF environments can understand. So the triples using those new properties will be part of classic/older data. The application of the algorithm may be well isolated from classic/older data, but the result is not.
> 
> I suggest we defer this discussion to later, and not block this PR on this basis. If we agree on the general algorithms, it is easy to change the vocabulary later. I just added an issue marking the `classicize:` namespace as AT RISK to force us rediscuss that ([c054000](https://github.com/w3c/rdf-concepts/commit/c054000fd3698eee2a4a63870f5cec0a9b2d214c)). Would that work for you?

Nonetheless that would make it harder to change later. Since defining a new namespace was never mentioned before I would not want it to be pushed right away. In fact, your choices of vocabulary so far run counter prior discussions both in Github issues and SemTF meetings, so I'd really prefer to discuss them first. 
 
> > I've illustrated the level of semantics equivalence that IMO is desirable with the [#115 (comment)](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2539045176), namely that _:r owl:sameAs  _:rs ; _:sr1 .
> 
> If you relly on the semantics of `owl:sameAs` to define this equivalence, then clearly it only exists in entailment regimes beyond those defined by our working group, right?

I think you misinterpret my use of vocabulary. The example uses `owl:sameAs` because the meaning of `owl:sameAs` is pretty well understood in these circles and therefore a practical tool in illustrating an argument. This doesn't mean that the level of equivalence that I mean to illustrate requires OWL, or any reasoning regime for that matter. Equivalence in meaning is a concept quite integral to RDF, so maybe it would be wiser to not use `owl:sameAs` in examples? OTOH it is pretty customary. Anyway: I mean equivalence in the same sense that two IRIs may be considered equivalent if they denote the same thing in the realm of discourse. Is that clear enough? Or is an example using `owl:sameAs` just more practical as a means of communication?

> Indeed, I don't aim to _prevent_ anyone from defining this kind of equivalenve in a semantic extension, but I don't think we should bake that in the regimes we define ourselves. I also added a sentence to that effect ([ec8663c](https://github.com/w3c/rdf-concepts/commit/ec8663cbc17bcfb4507ae63a883e4ac38f49d8f8)).

Requiring a semantic extension to do that work when we could so easily go the extra mile is what I would not like. In fact I think it helps understanding, acceptance and uptake of triple terms if we lay out the relation to RDF standard reification and especially the above mentioned equivalence as clearly as possible. 

> With these two caveat, @rat10, would you agree to merge the PR and continue the discussion in further issues or PRs?

Sorry, but no. I think we should first discuss it in a WG meeting.

pfps (changes requested):
The first description of the algorithm is missing the recursive aspect.   If the new triples have a triple term as object then the algorithm must be invoked on the new triples.

rat10:
> > Since defining a new namespace was never mentioned before
> 
> If I replaced all "classicise:TripleTerm" with "rdf:ClassicizedTripleTerm", "classicize:subject" with "rdf:classicizeSubject", and so on, keeping the marker explaining that those terms may change, would that be more acceptable for you?

At least that sounds so strange that it is almost guaranteed to not make it into the final version. However, "almost" is not really comforting given that we are under time pressure and things might get messy in teh end. So I’d rather we continue this discussion until we settle for something less ad hoc.

Is there a specific reason why you use the term "classic"? IIRC we voted for a proposal of @Antoine-Zimmermann which disambiguates between "Basic" and "Full". I find "basic" to be a more fitting name, eluding to the fact that "basic triples" are still the core of RDF modelling.

Also, IIRC, @doerthe and @kasei  were concerned about means to describe the constituents of triple terms, without any reference to backwards compatibility - just to keep in mind that backwards compatibility and mapping to RDF standard ("classic" ;) reification is not the only task the mapping may be used for. Therefore naming should be as technical and neutral as possible. I still favor `rdf:tripleTermSubject` etc over all other proposals. 

> > Requiring a semantic extension to do that work when we could so easily go the extra mile is what I would not like.
> 
> Having the unstar/classicize mapping impact the semantics was also "never mentioned before", to reuse your argument above, so this should not be blocking for this PR.

I remember a discussion in the Semantics TF a few weeks ago when the now obsolete graph-based mapping was criticized for not being semantics-preserving. So there is definitely an ongoing discussion, and maybe even a misunderstanding, see below.

> Furthermore, if going the extra mile is as easy as you claim, then it can be done in an additional PR once this one is merged.

I don’t see the need to divide this not overly complex topic in two PRs. I'd rather see taht we come to a complete solutiion. We seem to converge, and if we can't we need to discuss this in the WG first anyway.

> > In fact I think it helps understanding, acceptance and uptake of triple terms if we lay out the relation to RDF standard reification
> 
> I wholeheartedly agree with that, but since old-style reification has not specific formal semantics attached to it (in the sense that it does not impact entailment in any way),

IIUC RDF standard reification has a formal semantics, it just is not normative. But there seems to be a deeper misunderstanding. You seem to understand "entailment" when I say "semantics", but that’s not my drift. I’m not concerned about possible entailments, but about equivalences in meaning of the different constructs, old and new. 

> I think this can (and should) be done without a change in the formal semantics. Adding prose to that effect, in this section, in RDF-SEMANTICS or in RDF-PRIMER, would be sufficient.

I tend to agree. Adding prose (probably in various places, maybe foremost here) should be sufficient.

pchampin:
> The first description of the algorithm is missing the recursive aspect. If the new triples have a triple term as object then the algorithm must be invoked on the new triples.

My idea was that "each triple term <var>tt</var> used in the graph" and "all occurrences of <var>tt</var> used in the graph" referred to uses/occurrences *at any depth*. With this reading, I believe that the first description of the algorithm is consistent with the recursive pseudo-code version.

If I added the parenthesis: "all occurrences of <var>tt</var> used **(at any depth)** in the graph", would you find this sufficiently explicit? (despite the fact that the notion of "depth" was never explicitly defined)

I could also add a deeply nested triple term in the example, to make this clearer...

pchampin:
@rat10 , to respond to several points in your latest comment

> Is there a specific reason why you use the term "classic"?

Because that's how it is currently called in the spec! But you have a point: [this decision](https://www.w3.org/2023/10/05-rdf-star-minutes#r03) has not been entirely reflected in the spec (the term "classic" was introduced [prior](https://github.com/w3c/rdf-concepts/pull/32) to this decision).

> I don’t see the need to divide this not overly complex topic in two PRs.

The idea is to put what we can agree on out of the way. It is ok to merge things that are not fully baked yet -- especially when they are explicit about not beeing fully baked, with "issue markers" (the red boxes in the text). Conversely, keeping a PR open for a long time makes it increasingly harder to merge if other PRs are merged in the meantime.

>  I’m not concerned about possible entailments, but about equivalences in meaning

Since equivalence is [defined in terms of entailment](https://www.w3.org/TR/rdf11-concepts/#dfn-equivalence), I assume that you are not using the terms "equivalence" in its technical sense either...

I suggest (as I have done before) that we reserve the terms "semantics" (and "equivalence") for what is defined in RDF-SEMANTICS. Of course, the IRIs in the RDF vocabulary (or any other vocabulary, for that matter) have an *intended meaning* that is only partially conveyed by the formal semantics, and that is also important.

But when you write "semantic equivalence", and use `owl:sameAs` (which has drastic effects in terms of entailment) in your examples, you really sound like you are talking about the formal semantics rather than the (informal) intended meaning. So that's very confusing...

afs:
> The first description of the algorithm is missing the recursive aspect. If the new triples have a triple term as object then the algorithm must be invoked on the new triples.

If you could suggest text, that would be useful, or at least a comment on the particular text you are referring to.

afs:
> > The first description of the algorithm is missing the recursive aspect. If the new triples have a triple term as object then the algorithm must be invoked on the new triples.

@pfps a comment next to the text would be useful - I'm having to guess what you are referring to.

> If I added the parenthesis: "all occurrences of tt used **(at any depth)** in the graph", would you find this sufficiently explicit? (despite the fact that the notion of "depth" was never explicitly defined)

It _may_ be this:
> `(b, classicize:object, o)`

`o` needs processing (as it is in the algorithm).

> I could also add a deeply nested triple term in the example, to make this clearer...

An example with all cases would be useful.

This is another reason to merge - examples do not show up in the preview.

pchampin:
> examples do not show up in the preview.

I added a link to an "evergreen" preview that includes the examples, just above the automatic "pr-preview" links.

pfps:
> > The first description of the algorithm is missing the recursive aspect. If the new triples have a triple term as object then the algorithm must be invoked on the new triples.
> 
> My idea was that "each triple term tt used in the graph" and "all occurrences of tt used in the graph" referred to uses/occurrences _at any depth_. With this reading, I believe that the first description of the algorithm is consistent with the recursive pseudo-code version.
> 
> If I added the parenthesis: "all occurrences of tt used **(at any depth)** in the graph", would you find this sufficiently explicit? (despite the fact that the notion of "depth" was never explicitly defined)
> 
> I could also add a deeply nested triple term in the example, to make this clearer...

One problem is that you have to be careful when things happen.  If the replacement is done all at once then the objects of the added triples will not be subject to the replacement.  You have to ensure that the objects of all the added triples are subject to the replacement.  One way to do that is to work top-down recursively, but there are many other ways.

Another problem is that there is no definition for occur, so one might take it to mean "be a subject, predicate, or object of a triple in the graph".

As far as a suggestion on what to do, I'm unconvinced that this section is needed at all so my suggestion as to what to do if the section has a problem is to just not include it.

pchampin:
@pfps
> I'm unconvinced that this section is needed

It was [suggested earlier](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2508172643) that a prose description was more in line with the rest of the document than pseudo-code. That's why I added this section.

>  If the replacement is done all at once then the objects of the added triples will not be subject to the replacement.

If your graph contains `<< :a :b << :c :d :e >> >>`,
* you can start by replacing `<< :c :d :e >>` by a blank node _:cde *everywhere* (including when it is embedded in another triple term), then replacing `<< :a :b _:cde >>` with another blank node everywhere;
* but you can also start by replacing `<< :a :b << :c :d :e >> >>` by a blank node _:abcde, for which you will generate a triple `_:abcde classicize:object << :c :d :e >>`, then replacing `<< :c :d :e >>` with a blank node _:cde (including in the newly generated triples).
I believe that it works regarless of the order, provided that we agree on what "use" and "occurrence" means... which brings me to your point:

> there is no definition for occur

agreed. I suggest that we reintroduce the notion of [constituent term](https://www.w3.org/2021/12/rdf-star.html#dfn-constituent) that was defined in the RDF-star CG report. We could then rewrite the "From Full to Classic" section to:

> > It consists in assigning each [triple term](https://raw.githack.com/w3c/rdf-concepts/refs/heads/action-129/spec/index.html#dfn-triple-term) <var>tt</var> <ins>that is a [constituent term](https://www.w3.org/2021/12/rdf-star.html#dfn-constituent) of the graph</ins> to a fresh [blank node](https://raw.githack.com/w3c/rdf-concepts/refs/heads/action-129/spec/index.html#dfn-blank-node) <var>b</var> (i.e. a blank node not in use in the graph). All occurrences of <var>tt</var> in the graph, <ins>including as [constituent term](https://www.w3.org/2021/12/rdf-star.html#dfn-constituent) of other [triple term](https://raw.githack.com/w3c/rdf-concepts/refs/heads/action-129/spec/index.html#dfn-triple-term),</ins> are replaced with <var>b</var>...

WDYT?

pchampin:
>> (...)

> The "start by ... then" phrasing in both bullets shows that the replacement is not "all at once", to my mind.

Then I don't understand what "all at once" means in this context.

> Nor do I believe that "at any depth" will communicate what is meant nor necessary.

Fair enough. What about my proposal above using the concept of "constituent term"?

> Recursion is the correct term for the necessary process.

Recursion is an implementation detail of the pseudo-code version.

What the prose is trying to convey is basically two nested loops, which are (hopefully) easier to grasp, technically less efficient than the pseudo-code (hence the value of having both), and producing the same result.

gkellogg (approved):
I do think the mapping should be normative. And, we should create terms in the RDF vocabulary for this, rather than a separate `classicize` namespace.

An example with recursive triple terms would be helpful.

pfps:
Another way of illuminating a problem is whether s,p,o are pre-replacement or post-replacement in:
> All occurrences of tt in the graph are replaced with b, and the following triples are added to the graph (where s, p and o are respectively the [subject](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-subject), [predicate](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-predicate) and [object](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-object) of tt):
I read the text as indicating that they are pre-replacement, which then may require further application of the rule.  But there is another problem if any replacement is carried out inside-out as an inner replacement would create a new triple term which would then persist.  But outside-in also has problems.
So maybe the best solution is an iterative one - pick a tt, replace it everywhere with a fresh b, and add the new triples; repeat until no tt.
It may be possible to come up with wording that is one-shot, but I expect that the wording will either be lengthy or some math is required.

pchampin:
> Another way of illuminating a problem is whether s,p,o are pre-replacement or post-replacement in:
> 
> > All occurrences of tt in the graph are replaced with b, and the following triples are added to the graph (where s, p and o are respectively the [subject](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-subject), [predicate](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-predicate) and [object](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/115.html#dfn-object) of tt):
> 
> I read the text as indicating that they are pre-replacement, which then may require further application of the rule. But there is another problem if any replacement is carried out inside-out as an inner replacement would create a new triple term which would then persist.

Now I see it. 

> But outside-in also has problems. So maybe the best solution is an iterative one - pick a tt, replace it everywhere with a fresh b, and add the new triples; repeat until no tt.

Agreed. (and this is what was done in the CG report, by the way... :sweat_smile: )

> It may be possible to come up with wording that is one-shot, but I expect that the wording will either be lengthy or some math is required.

I'll update this PT accordingly, and see what I can do.

pfps (commented):
This looks good now.  A bit of wordsmithing might be useful - "it consists .... of ..."

rat10:
For the record: I don't see the issues I raised adequately addressed. I see no need to push this PR w.rt. avoiding conflicts with other PRs as it is really very well isolated. Questions w.r.t. naming (the properties, basic vs classic) IMHO simply arose because the PR strays from agreed upon terminology and should be rolled back until properly discussed (e.g. by raising an issue about basic vs classic). My proposal w.r.t.the most important issue - how triple terms relate to RDF standard refication - hasn't even been properly discussed. So: no, I'm still not okay with merging this PR.

spec/index.html
```diff
@@ -1450,6 +1470,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+
+  <dl>
+    <dt>Information preserving</dt>
+    <dd>It must be possible to reconstruct the input graph (resp. dataset) from the output graph (resp. dataset).
+      Note that, on the other hand, these transformations are not designed to be semantics preserving:
+      the output graph is not semantically [=equivalent=] to the input graph,
+      at least not in the entailment regimes defined in [[RDF12-SEMANTICS]].
+    </dd>
+    <dt>Idempotent</dt>
+    <dd>Applying a transformation several time to a graph (resp. dataset) should have the same effect as applying it once.
```
afs, line 1495:
```suggestion
    <dd>Applying a transformation several times to a graph (resp. dataset) should have the same effect as applying it once.
```

spec/index.html
```diff
@@ -1450,6 +1470,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+  </dl>
+
+  <section id="section-classicize-definition">
+    <h2>From [=Full=] to [=Classic=]</h2>
+
+    <p>
+      Encoding an [=RDF graph=] to ensure that it is consumable by an RDF [=Classic=] implementation is called <dfn data-lt="classicize|classicized">classicizing</dfn> it.
+      It consists, while the graph has a [=triple term=] <var>tt</var> in its [=constituent terms=], to mint a fresh [=blank node=] <var>b</var>
+      (i.e. a blank node not in use in the graph), and replace <var>tt</var> with <var>b</var> in all the triples of the graph having <var>tt</var> in their [=constituents=].
+      Then the following triples are added to the graph (where <var>s</var>, <var>p</var> and <var>o</var> are respectively the [=subject=], [=predicate=] and [=object=] of <var>tt</var>):
```
gkellogg, line 1511: This section is written considering that graphs are mutable (add or remove something to a graph), while we define a graph as being static:

> The RDF data model is atemporal: [RDF graphs](https://raw.githack.com/w3c/rdf-concepts/refs/heads/action-129/spec/index.html#dfn-rdf-graph) are static snapshots of information.

Would it not be cleaner to describe the creation of a new graph after applying such transformations to make these operations more functional?

gkellogg (commented):
I think the general sense of this is good, but we should be consistent on graph immutability.

spec/index.html
pchampin: > This section is written considering that graphs are mutable (add or remove something to a graph), while we define a graph as being static

Yes, I agree that could be confusing, but there are precedents of speaking of "replacing" nodes in a graph (sec 3.7 and 3.8). And I think that this is a concise and intuitive way to convey what the transformation is.

Note that the algorithm, on the other hand assumes immutable graphs and produces a new graph from scratch.

spec/index.html
```diff
@@ -1450,6 +1470,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+      There is actually <a href="#section-classicize-caveat">a minor caveat</a> to this property.
+    </dd>
+  </dl>
+
+  <section id="section-classicize-definition">
+    <h2>From [=Full=] to [=Classic=]</h2>
+
+    <p>
+      Encoding an [=RDF graph=] to ensure that it is consumable by an RDF [=Classic=] implementation is called <dfn data-lt="classicize|classicized">classicizing</dfn> it.
+      It consists, while the graph has a [=triple term=] <var>tt</var> in its [=constituent terms=], to mint a fresh [=blank node=] <var>b</var>
```
pchampin, line 1509:
```suggestion
      It consists, while the graph has a [=triple term=] <var>tt</var> in its [=constituent terms=], in minting a fresh [=blank node=] <var>b</var>
```

pchampin:
> Questions w.r.t. naming (the properties, basic vs classic) IMHO simply arose because the PR strays from agreed upon terminology and should be rolled back until properly discussed (e.g. by raising an issue about basic vs classic).

As I explained [above](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2546335721), the current version of the spec, on which this PR is based, defined "Full conformance" and "Classic conformance", so the PR sticks to this and refers to "Classic conformance". Using "Basic" in section 8 while section 2 says "Classic" would have been inconsistent and confusing.

Note that I personally don't have a strong preference between "basic" and "classic". It is not that I refuse to make the change, but I don't think that this PR should "casually" change a normatively defined term, when its purpose is somewhere else.

pchampin:
> My [comment](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2539045176) w.r.t.the most important issue - how triple terms relate to RDF standard refication - hasn't been properly discussed. So: no, I'm still not okay with merging this PR.

I just [added](https://github.com/w3c/rdf-concepts/pull/115/commits/9dfbd3eaaa5dcbb7efb2a7a7faa5544b59faf53d) a note about why we introduce another vocabulary rather than reuse the old reification vocabulary.

For more general discussion about the relation between triple terms and old-style reification, as I [responded earlier](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2540917616): I agree that it is needed, but this should be in rdf-semantics and/or rdf-primer. And anyway, this is orthogonal to this PR: the relationship between triple terms and old-style reification is about triple terms in general, not just triple terms being "classicized".

spec/index.html
```diff
@@ -888,6 +889,12 @@ <h3>Triple Terms</h3>
       Every <a>triple</a> whose <a>object</a> is not a <a>triple term</a> SHOULD NOT
       use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
       as its <a>predicate</a>.</p>
+
+    <p>The <dfn data-lt="constituent">constituent terms</dfn> (or simply constituents)
+      of a [=triple term=] (resp. an [=RDF triple=]) are its [=subject=], its [=predicate=], its [=object=],
```
TallTed, line 894: `resp.`? This expands in my head to `respectively` (and minimally requires a following comma, in every instance), but I don't know what the `respective` reference is.

Abbreviations like this should generally be avoided, especially when, as here, their meaning might be at all confusing.

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+
+  <p>This section provides transformations between [=Full=] [=RDF graphs=] (resp. [=RDF datasets=]) and [=Classic=] [=RDF graphs=] (resp. [=RDF datasets=]),
   to provide some level of interoperability between the different classes of <a href="#conformance">Conformance</a>.
 
   <p class=issue>Should we go even further and aim to provide interoperability between <em>RDF 1.1</em> and RDF 1.2 [=Full=]?</p>
 
-  <p>It defines the <a href="#section-unstar-algo">`unstar`</a> algorithm, which transforms an RDF [=Full=] dataset into an RDF [=Classic=] dataset by encoding all triple-terms into dedicated named graphs.
-    This algorithm is designed to be:
-  </p>
+  <p class=issue>AT RISK: the working group may decide to replace the terms `rdf:TripleTerm`, `rdf:ttSubject`, `rdf:ttPredicate` and `rdf:ttObject` used in this section by other terms, possibly in a different namespace.</p>
```
TallTed, line 1483:
```suggestion
  <p class=issue>AT RISK: The Working Group may decide to replace the terms `rdf:TripleTerm`, `rdf:ttSubject`, `rdf:ttPredicate`, and `rdf:ttObject` used in this section with other terms, possibly in a different namespace.</p>
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+
+  <p>These transformation are designed to be:</p>
 
   <dl>
     <dt>Information preserving</dt>
-    <dd>It must be possible to reconstruct the input dataset from the output dataset.
-      Note that, on the other hand, the algorithm is not designed to be semantics preserving:
-      the graphs in the produced dataset are not semantically <a>equivalent</a> to their corresponding graph in the input dataset.
+    <dd>It must be possible to reconstruct the input graph (resp. dataset) from the output graph (resp. dataset).
+      Note that, on the other hand, these transformations are not designed to be semantics preserving:
```
TallTed, line 1490:
```suggestion
      Note, however, that these transformations are not designed to preserve semantics:
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+      the output graph is not semantically [=equivalent=] to the input graph,
+      at least not in the entailment regimes defined in [[RDF12-SEMANTICS]].
     </dd>
     <dt>Idempotent</dt>
-    <dd>Transforming a dataset that is already complying with RDF [=Classic=] (i.e. containing no <a>triple term</a>) must result in the same dataset.
+    <dd>Applying a transformation several times to a graph (resp. dataset) should have the same effect as applying it once.
+      Moreover, [=classicizing=] a graph (resp. dataset) that is already complying with RDF [=Classic=] (i.e. containing no [=triple term=]) must result in the same graph (resp. dataset).
```
TallTed, line 1496: I think we will be well served by using `RDF Classic` as the defined term, rather than just `Classic`. I'm not making that change here; only suggesting it to you.
```suggestion
      Moreover, [=classicizing=] a graph (resp. dataset) that is already complying with RDF [=Classic=] (i.e., containing no [=triple term=]) must result in the same graph (resp. dataset).
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
     </dd>
     <dt>Universal</dt>
-    <dd>It should be possible to transform any RDF [=Full=] dataset using this method.
-      There is actually <a href="#section-unstar-caveat">a minor caveat</a> to this property.
+    <dd>It should be possible to transform any [=Full=] graph (resp. dataset) to a [=Classic=] graph (resp. dataset) using this method.
+      There is actually <a href="#section-classicize-caveat">a minor caveat</a> to this property.
```
TallTed, line 1500:
```suggestion
    <dd>It should be possible (with <a href="#section-classicize-caveat">a minor caveat</a>) to transform any [=Full=] graph (resp. dataset) to a [=Classic=] graph (resp. dataset) using this method.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-    This blank node is then use as the name of a new named graph containing a single triple `S P O`.
-    In order to distinguish those new named graphs from the original named graphs of the input dataset,
-    dedicated triples are added to a special `rdf:unstarMetadata` named graph.</p>
-
-  <p class=note>The blank nodes generated by `unstar` to replace <a>triple terms</a> should not be confused with the <a>reifiers</a> that are typically associated to these <a>triple terms</a>.</p>
+    <p>
+      Encoding an [=RDF graph=] to ensure that it is consumable by an RDF [=Classic=] implementation is called <dfn data-lt="classicize|classicized">classicizing</dfn> it.
+      It consists, while the graph has a [=triple term=] <var>tt</var> in its [=constituent terms=], in minting a fresh [=blank node=] <var>b</var>
+      (i.e. a blank node not in use in the graph), and replace <var>tt</var> with <var>b</var> in all the triples of the graph having <var>tt</var> in their [=constituents=].
+      Then the following triples are added to the graph (where <var>s</var>, <var>p</var> and <var>o</var> are respectively the [=subject=], [=predicate=] and [=object=] of <var>tt</var>):
```
TallTed, line 1511:
```suggestion
      [=Classicizing=] consists, while the graph has a [=triple term=] <var>tt</var> in its [=constituent terms=], of minting a fresh [=blank node=] <var>b</var>
      (i.e., a blank node not yet in use in the graph); replacing each <var>tt</var> with <var>b</var> in all the triples of the graph having <var>tt</var> in their [=constituents=];
      and then adding the following triples to the graph (where <var>s</var>, <var>p</var>, and <var>o</var> are respectively the [=subject=], [=predicate=] and [=object=] of <var>tt</var>):
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+    </p>
+    <ul>
+      <li>(<var>b</var>, `rdf:type`, `rdf:TripleTerm`)
+      <li>(<var>b</var>, `rdf:ttSubject`, <var>s</var>)
+      <li>(<var>b</var>, `rdf:ttPredicate`, <var>p</var>)
+      <li>(<var>b</var>, `rdf:ttObject`, <var>o</var>)
+    </ul>
+    <p>This process is repeated until the graph has no [=triple term=] [=constituent=], and is therefore compliant with RDF [=Classic=].</p>
```
TallTed, line 1519: I've moved this sentence into the introductory sentence, above.
```suggestion
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
 
-  <p class=note>Despite having the same name, the `unstar` mapping defined in this specification is different from the `unstar` mapping defined in [[RDF-STAR-CG]].</p>
+    <p>Note that this transformation is <em>information preserving</em> only when the input graph does not contain at the same time a [=triple term=]
+      and an [=asserted=] triple (<var>b</var>, `rdf:type`, `rdf:TripleTerm`) where <var>b</var> is a [=blank node=].
+      Implementations encountering this situation MUST report an error.
+      See Section <a href="#section-classicize-caveat"></a> for a discussion on this limitation.
```
TallTed, line 1524:
```suggestion
      This limitation is discussed in Section <a href="#section-classicize-caveat"></a>.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+    </p>
 
+    <p class=note>The blank nodes generated to replace [=triple terms=] should not be confused with the [=reifiers=] that are typically associated to these [=triple terms=].</p>
```
TallTed, line 1527: It would be good to link this to the discussion of those "typical associations".
```suggestion
    <p class=note>The blank nodes generated to replace [=triple terms=] should not be confused with the [=reifiers=] that are typically associated with these [=triple terms=].</p>
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
 
-  <section id="section-unstar-algo" class="algorithm">
-    <h2>The `unstar` algorithm</h2>
+    <p>
+      [=Classicizing=] an [=RDF dataset=] consists in [=classicizing=] its [=default graph=] and each of its [=named graph=].
```
TallTed, line 1530:
```suggestion
      [=Classicizing=] an [=RDF dataset=] consists of [=classicizing=] its [=default graph=] and each of its [=named graph=].
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+      In this case, the fresh [=blank node=] assigned to each [=triple term=] must not be used in any graph of the dataset.
+    </p>
 
-    <p>The algorithm expects one input variable <var>Dᵢ</var> which is an <a>RDF dataset</a>. It returns a [=Classic=] <a>RDF dataset</a>.
-      In the algorithm, we adopt the view presented in <a href="#section-dataset-quad"></a>.
+    <p>
+      See Section <a href="#section-classicize-algo"></a> for a detailed algorithm of the transformation.
```
TallTed, line 1535:
```suggestion
      A detailed algorithm of the transformation is found in Section <a href="#section-classicize-algo"></a>.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-        </ol></li>
-        <li>Add the quad (<var>s</var>, <var>p</var>, <var>o</var>, <var>g</var>) to <var>Dₒ</var>.</li>
-      </ol></li>
-      <li>Return <var>Dₒ</var>.</li>
-    </ol>
-  </section>
+    <section id="section-classicize-example">
+      <h2>Example</h2>
+
+      <p>The examples in this section are using the Turtle concrete syntax [[RDF12-TURTLE]].</p>
```
TallTed, line 1541:
```suggestion
      <p>The examples in this section are expressed in the Turtle concrete syntax [[RDF12-TURTLE]].</p>
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+  <section id="section-unclassicize-definition">
+    <h2>From [=Classic=] to [=Full=]</h2>
+
+    <p>Reverting a [=classicize=] graph to its original form consists,
+      for each [=blank node=] <var>b</var> that is the subject of an [=asserted=] triple (<var>b</var>, `rdf:type`, `rdf:TripleTerm`),
+      in locating the three other [=asserted=] triples (<var>b</var>, `rdf:ttSubject`, <var>s</var>),
+      (<var>b</var>, `rdf:ttPredicate`, <var>p</var>),
+      and (<var>b</var>, `rdf:ttObject`, <var>o</var>).
+      These four triples are removed from the graph.
+      All remaining occurrences of <var>b</var> as a [=constituent term=] of the graph are then replaced with the triple term (<var>s</var>, <var>p</var>, <var>o</var>).
```
TallTed, line 1577:
```suggestion
    <p>Reverting a [=classicized=] graph to its original form consists of locating
      each [=asserted=] triple (<var>b</var>, `rdf:type`, `rdf:TripleTerm`)
      that has a [=blank node=] <var>b</var> as its subject, 
      along with the three associated [=asserted=] triples 
      that have the same [=blank node=] <var>b</var> as their subjects, i.e.,
      (<var>b</var>, `rdf:ttSubject`, <var>s</var>),
      (<var>b</var>, `rdf:ttPredicate`, <var>p</var>),
      and (<var>b</var>, `rdf:ttObject`, <var>o</var>);
      removing these four triples from the graph;
      and replacing all remaining occurrences of <var>b</var> 
      as a [=constituent term=] of the graph 
      with the triple term (<var>s</var>, <var>p</var>, <var>o</var>).
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-        <li>Add the association (<var>t</var>, <var>b</var>) to <var>Mₒ</var>.</li>
-        <li>Add the quads (<var>s</var>, <var>p</var>, <var>o</var>, <var>b</var>) and (<var>b</var>, `rdf:type`, `rdf:TripleTerm`, `rdf:unstarMetadata`) in <var>D</var>.</li>
-        <li>Return <var>b</var>, <var>Mₒ</var> and <var>D</var>.</li>
-      </ol></li>
-    </ol>
+    <p>Implementations MUST report an error if, for a given <var>b</var>,
+      it can not unambiguously determine <var>s</var>, <var>p</var> or <var>o</var>
+      (i.e. if one of the `classicize:` properties of <var>b</var> is missing or duplicated).
+      Implementations MUST also report an error if the input graph contains at the same time a [=triple term=] and an [=asserted triple=] (<var>b</var>, `rdf:type`, `rdf:TripleTerm`) where <var>b</var> is a [=blank node=].
+      None of these situations can occur if the input graph was produced by the [=classicize=] transformation.
```
TallTed, line 1584:
```suggestion
    <p>An implementation MUST report an error if, for a given <var>b</var>,
      it can not unambiguously determine <var>s</var>, <var>p</var>, or <var>o</var>
      (i.e., if one of the `classicize:` properties
      — `rdf:ttSubject`, `rdf:ttPredicate`, or `rdf:ttObject` —
      of <var>b</var> is missing or duplicated).
      An implementation MUST also report an error if the input graph contains
      at the same time a [=triple term=] and an [=asserted triple=]
      (<var>b</var>, `rdf:type`, `rdf:TripleTerm`) 
      where <var>b</var> is the same [=blank node=].
      Note that none of these situations can occur if the input graph was produced by the [=classicize=] transformation.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-    <h2>Limitations of the `unstar` algorithm</h2>
+  <section id="section-classicize-caveat">
+    <h2>Limitations</h2>
 
-    <p>Steps <a href="#unstar-error1">4.1.1</a> and <a href="#unstar-error2">4.2.1</a> of the <a href="#section-unstar-algo">`unstar`</a> algorithm exit with an error.
-      This will occur if the input dataset contains at the same time <a>triple terms</a> and a graph named `rdf:unstarMetadata`.
-      The `unstar` algorithm is therefore not strictly universal as it can not transform this particular kind of datasets.
+    <p>The two transformations above are explicitly not supporting graphs or datasets containing at the same time a [=triple term=]
+      and an [=asserted triple=] (<var>b</var>, `rdf:type`, `rdf:TripleTerm`) where <var>b</var> is a [=blank node=].
+      This means, in particular, that the [=classicize=] transformation is not strictly <em>universal</em>.
```
TallTed, line 1601:
```suggestion
    <p>The two transformations above explicitly do not support graphs or datasets containing at the same time a [=triple term=] and an [=asserted triple=]
      (<var>b</var>, `rdf:type`, `rdf:TripleTerm`) 
      where <var>b</var> is the same [=blank node=].
      This means that the [=classicize=] transformation is not <em>strictly</em> universal.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
     </p>
 
     <p>This limitation should not be an issue in practice.
-      The special graph name `rdf:unstarMapping` is unlikely to be in used in any published dataset,
-      as it was not defined in the RDF namespace prior to this specification.
+      The `rdf:TripleTerm` type is unlikely to be in used in any published graph or dataset,
+      as it was not defined prior to this specification.
       For this reason, using it would actually have been bad practice.
```
TallTed, line 1607:
```suggestion
      as it was not defined prior to this specification,
      and, therefore, using it would actually have been bad practice.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-      As for future datasets, their authors should consider the graph name `rdf:unstarMetadata` to be reserved, in order to prevent interference with the `unstar` algorithm.
+      As for future graphs and datasets, their authors should consider this type to be reserved, in order to prevent interference with the [=classicize=] transformation.
```
TallTed, line 1608:
```suggestion
      For future graphs and datasets, this type should be considered to be reserved for use within the [=classicize=] transformation, and not used otherwise.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
-      would result in a "hybrid" dataset that `unstar` can not process.
-      Such application should make sure to apply `unstar` to every dataset priori to merging them.
-      Since `unstar` is idempotent, there is no harm in applying it more than necessary.</p>
+    <p class=note>
+      This is one of the reasons why this transformation introduces a new vocabulary
+      (`rdf:TripleTerm`, `rdf:ttSubject`, `rdf:ttPredicate`, `rdf::ttObject`),
+      rather than repurposing the existing <a data-cite="RDF12-SCHEMA#ch_reificationvocab">reification vocabulary</a>
+      (`rdf:Statement`, `rdf:subject`, `rdf:predicate`, `rdf:object`).
+      Contrarily to `rdf:TripleTerm`, `rdf:Statement` is known to used in widely used datasets (e.g. <a href="https://www.uniprot.org/">Uniprot</a>),
+      so deprecating its usage as "reserved" was not an option.
```
TallTed, line 1617:
```suggestion
      This is one reason why this transformation introduces new vocabulary terms
      (`rdf:TripleTerm`, `rdf:ttSubject`, `rdf:ttPredicate`, `rdf::ttObject`),
      rather than repurposing the existing <a data-cite="RDF12-SCHEMA#ch_reificationvocab">reification vocabulary</a>
      (`rdf:Statement`, `rdf:subject`, `rdf:predicate`, `rdf:object`).
      Unlike `rdf:TripleTerm`, `rdf:Statement` is known to be found in
      widely used datasets (e.g., <a href="https://www.uniprot.org/">Uniprot</a>),
      so reserving its use for the [=classicize=] transformation was not an option.
```

spec/index.html
```diff
@@ -1466,164 +1473,233 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
...
+    </p>
 
+    <p>Another consequence of this restriction is that users should be careful when merging graphs in an application that [=classicize=] graphs or datasets.
+      More precisely, merging a [=Full=] [=RDF graph=] (containing at least one [=triple term=])
+      with a [=classicized=] [=RDF graph=] (and therefore potentially containing [=blank node=] instances of `rdf:TripleTerm`)
+      could result in a "hybrid" graph that can not be transformed.
+      Such applications should make sure to [=classicize=] every graph prior to merging them.
+      Conversely, applications supporting RDF [=Full=] should make sure to apply the reverse tranformation to any graph that is known or likely to be [=classicized=],
+      to avoid creating such "hybrid" graphs.
+      Since these transformations are <em>idempotent</em>, there is no harm in applying them more than necessary.
```
TallTed, line 1627: Its troubling that this caution was written to *users* and then later, to *applications*, rather than *implementers* who ought to bear the burden. Further editing will be good. Also, this segment --
```
      Therefor, such applications should [=classicize=] every graph prior to merging them.
      Conversely, applications supporting RDF [=Full=] should make sure to apply the reverse transformation
      to any graph that is known or likely to have been [=classicized=],
      to avoid creating such "hybrid" graphs.
```
— would benefit from rewriting to make it a single caution that covers both `[=Classic=]` applications classicizing, and `[=Full=]` applications "de-classicing" (there must be a better word! "the reverse transformation" doesn't cut it...), every graph


```suggestion
    <p>Another consequence of this restriction is that users will need to be aware and careful when merging graphs in an application that [=classicizes=] graphs or datasets.
      The concern is that merging a [=Full=] [=RDF graph=] containing at least one [=triple term=]
      with a [=classicized=] [=RDF graph=] (which might contain [=blank node=] instances of `rdf:TripleTerm`)
      could result in a "hybrid" graph that cannot be transformed to a consistent [=Full=] nor [=Classic=] [=RDF graph=].
      Therefor, such applications should [=classicize=] every graph prior to merging them.
      Conversely, applications supporting RDF [=Full=] should make sure to apply the reverse transformation
      to any graph that is known or likely to have been [=classicized=],
      to avoid creating such "hybrid" graphs.
      Since these transformations are designed to be <em>idempotent</em>, there is no harm in applying them more than necessary.
```

TallTed (changes requested):
I've reviewed spec/index.html up to the Algorithms. I'll come back for that. There are a number of requested changes above. I didn't think I'd find so many, nor that I'd have this much time to do so, or I'd have bundled them into a review... Sorry for the extra clicks these will take to apply!

rat10:
> > Questions w.r.t. naming (the properties, basic vs classic) IMHO simply arose because the PR strays from agreed upon terminology and should be rolled back until properly discussed (e.g. by raising an issue about basic vs classic).
> 
> As I explained [above](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2546335721), the current version of the spec, on which this PR is based, defined "Full conformance" and "Classic conformance", so the PR sticks to this and refers to "Classic conformance". Using "Basic" in section 8 while section 2 says "Classic" would have been inconsistent and confusing.
> 
> Note that I personally don't have a strong preference between "basic" and "classic". It is not that I refuse to make the change, but I don't think that this PR should "casually" change a normatively defined term, when its purpose is somewhere else.

I understand - although I differ w.r.t. "casually" and "normatively". However,  I'd rather say that the mistake should not be made worse by repeating it but rather pointed out where it was made. Also, we had the discussion a year ago and I don't want to repeat it unnecessarily, but I do indeed like the pair "Basic" and "Full" much more than "Classic" and "Full". On an unrelated note I meanwhile would argue to not have that distinction at all, but it's probably too late for that.

rat10:
> > My [comment](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2539045176) w.r.t.the most important issue - how triple terms relate to RDF standard refication - hasn't been properly discussed. So: no, I'm still not okay with merging this PR.
> 
> I just [added](https://github.com/w3c/rdf-concepts/pull/115/commits/9dfbd3eaaa5dcbb7efb2a7a7faa5544b59faf53d) a note about why we introduce another vocabulary rather than reuse the old reification vocabulary.

Thank you, that addresses my respective concern. 

> For more general discussion about the relation between triple terms and old-style reification, as I [responded earlier](https://github.com/w3c/rdf-concepts/pull/115#issuecomment-2540917616): I agree that it is needed, but this should be in rdf-semantics and/or rdf-primer. And anyway, this is orthogonal to this PR: the relationship between triple terms and old-style reification is about triple terms in general, not just triple terms being "classicized".

So, do you propose to open a new issue on that topic?

niklasl:
@rat10 
> So, do you propose to open a new issue on that topic?

We have https://github.com/w3c/rdf-semantics/issues/61 .

rat10:
@niklasl 
> @rat10
> 
> > So, do you propose to open a new issue on that topic?
> 
> We have [w3c/rdf-semantics#61](https://github.com/w3c/rdf-semantics/issues/61) .

Okay, I wasn't aware. The tag says 'editorial'. What does that mean?

afs (approved):

spec/index.html
pchampin: The idea was that the definition of constituent term applies to both triple terms and RDF triples.
Anyway, it will go away, as I will replace it with "appears in" as defined (currently) in RDF-Semantics.

I'll fix all the "resp." into "respectively,".

spec/index.html
pchampin: Good point about the link, will add it.

spec/index.html
pchampin: Good point about "user"/"application"/"implementers" → I will improve this.

pfps (approved):
OK

pchampin:
Rebased this PR on top of the latest main, and added an entry in section 'Changes between RDF 1.1 and RDF 1.2'

TallTed (approved):

spec/index.html
```diff
@@ -1489,6 +1502,253 @@ <h2>Generalizations of RDF Triples, Graphs, and Datasets</h2>
 </section>
 
 
+<section id="section-classic-full-interop" class="informative">
+  <h2>Interoperability between RDF [=Classic=] and RDF [=Full=]</h2>
+
+  <p class=issue>Should we make this section normative?</p>
```
afs, line 1508: Yes, I think it should be.

It should be the way of doing the mapping, not to be taken as an illustration/suggestion.

PR #124, opened 2024-12-21
Remove 'recognized datatype IRI' in favor of RDF Semantics definition and use

### LINKED ISSUE #60: Drop the requirement to support ill-typed literals with recognized datatype IRIs
### Observation

RDF 1.1 requires that implementations support ill-typed literals, including ill-typed literals with recognized datatype IRIs.

Ill-typed literals with recognized datatype IRIs do not have any known use cases. They are semantically inconsistent, do not denote anything, have no value, and any triple that contains them is false in every interpretation.

Notice that there is nothing wrong with requiring implementations to support ill-typed literals with unrecognized datatype IRIs. For example, it is good that RDF implementations are required to support literals like [1] that have a datatype IRI that is not broadly recognized.

```turtle
[1] '### Header'^^<https://example.com/markdown>
```

However, it is unclear why implementations are *allowed* to support, let alone are *required* to support, ill-typed literals with recognized datatype IRIs.

### Example

Suppose a triple store recognizes the RDF datatype IRIs + the XSD datatype IRIs + the GeoSPARQL datatype IRIs. Such a triple store can upon data ingest immediately detect that [2] and [3] are ill-typed literals with recognized datatype IRI.

```turtle
[2] 'Yes'^^xsd:boolean
[3] 'The sea is everything. It covers seven tenths of the terrestrial globe.'^^xsd:boolean
```

The RDF 1.1 standard forbids triple stores to throw an error upon encountering data that contains [2] or [3], even though this may be the preferred data quality approach for many users.

### Suggestion

In RDF 1.2, let's weaken the RDF 1.1 phrase "Implementations MUST accept ill-typed literals" to:
1. "Implementations MUST support ill-typed literals with unrecognized datatype IRIs."
2. "Implementations MAY support ill-typed literals with recognized datatype IRIs."

Implementations MUST support the RDF datatype IRIs, and MAY support any other datatype IRIs that they believe important enough for their users. The notion "recognized datatype IRI" is used as defined in RDF 1.1 Semantics.

### Ramifications

The proposed change makes it possible for RDF 1.2 data to be accepted in one implementation, but not in another implementation. For example, it is possible to upload data that contains literals [2] and [3] into an implementation that does not recognize the `xsd:boolean` datatype IRI. But it is not possible to upload the same data into an implementation that does recognize the `xsd:boolean` datatype IRI.

This differentiation is a good thing, because it allows stricter implementations to be created, rather than requiring all implementations to support the exact same ill-typed nonsense data.

Notice that RDF 1.1 Semantics already allows implementations to differ from one another in their support for more/fewer recognized datatype IRIs. Implementations that differ in their recognized datatype IRIs already differ in their behavior in RDF 1.1.

### LINKED ISSUE #122: should Concepts include information about REQUIRED datatypes
Semantics states that support of several datatypes is REQUIRED for RDF entailment but Concepts does not have any analogue of this requirement.  I think that Concepts should discuss what is needed for RDF entailment.

afs:
Important: 

RDF-Semantics has never referenced RDF-Concepts for "recognized datatypes".
It has its own definition [dfn-recognized](https://www.w3.org/TR/rdf12-semantics/#dfn-recognize) so no changes are necessary for RDF Semantics.

There are two places that cover accepting literals. One is for literals with unknown dataypes, [section DatatypeIRIs](https://www.w3.org/TR/rdf12-concepts/#datatype-iris) and one is for known-to-be [ill-typed literals](https://www.w3.org/TR/rdf12-concepts/#section-Graph-Literal). In RDF 1.1 they were MUST and SHOULD. In this PR, they are aligned as SHOULD. RDF Semantics provides further restrictions for D-entailment and it's definition of "_recognized_" datatype IRIs that has additional conditions.

This closes #122.
This closes #60.

[Preview not depending on PR-Preview](https://raw.githack.com/w3c/rdf-concepts/recognized-datatype-iris/spec/index.html).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/124.html" title="Last updated on Jan 9, 2025, 1:42 PM UTC (55e77fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/124/c60c4ed...55e77fa.html" title="Last updated on Jan 9, 2025, 1:42 PM UTC (55e77fa)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -743,10 +743,10 @@
       <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
         a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
         likewise in that order.</li>
-      <li>If the literal's <a>datatype IRI</a> is handled by an RDF processor,
+      <li>If the literal's <a>datatype</a> is handled by an RDF implmentation,
         <ul>
           <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
-            of datatype, then the literal value is the result of applying
+            of the <a>datatype</a>, then the literal value is the result of applying
             the <a>lexical-to-value mapping</a> of the datatype to the
             <a>lexical form</a>.</li>
           <li>otherwise, the literal is ill-typed and no literal value can be
@@ -758,7 +758,7 @@
         </ul>
       </li>
       <li>If the literal's <a>datatype IRI</a> is <em>not</em>
-        handled by an RDF processor then the literal value is
+        handled by an RDF implementation, then the literal value is
         not defined by this specification.</li>
     </ul>
 
@@ -1099,7 +1099,7 @@
 
   <p>A <dfn>datatype</dfn> consists of a <a>lexical space</a>,
     a <a>value space</a> and a <a>lexical-to-value mapping</a>, and
-    is denoted by one or more <a>IRIs</a>.</p>
+    is identified by one or more <a>IRIs</a>.</p>
 
   <p>The <dfn>lexical space</dfn> of a datatype is a set of <a>strings</a>.</p>
 
@@ -1291,12 +1291,10 @@
   <section>
     <h3>Datatype IRIs</h3>
 
-    <p>Datatypes are identified by <a>IRIs</a> and the <a>referent</a>
-      of the IRI of a datatype defines a partial mapping from the
-      lexical form to a value.</p>
-
-   <p>If any IRI of the form
-      <code>http://www.w3.org/2001/XMLSchema#xxx</code> is handled by an RDF processor, it
+    <p>Datatypes are identified by <a>IRIs</a>.</p>
+    <p>
+      If any IRI of the form
+      <code>http://www.w3.org/2001/XMLSchema#xxx</code> is handled by an RDF implementation, it
       MUST refer to the RDF-compatible XSD type named <code>xsd:xxx</code> for
       every XSD type listed in <a href="#xsd-datatypes">section 5.1</a>.</p>
 
@@ -1312,12 +1310,12 @@
         refers to the datatype <code><a>rdf:JSON</a></code>.</li>
     </ul>
 
-    <p>RDF processors are not required to handle all datatypes.
-      Any literal typed with a datatype IRI not handled by an RDF processor
-      is treated just like an unknown IRI, i.e. as referring to an unknown thing.
+    <p>RDF implementations are not required to handle all datatypes.
+      Any literal typed with a datatype not handled by an RDF implementation
+      is treated just like an unknown IRI, i.e., as referring to an unknown thing.
       Applications MAY give a warning message if they are unable to determine the
-      referent of an IRI used in a typed literal. RDF processors SHOULD
-      not reject literals with unknown IRI as either a syntactic or
+      referent of an IRI used in a typed literal. RDF implementations SHOULD
+      not reject a literal with an unknown datatype as either a syntactic or
       semantic error.<p>
 
     <p>Other specifications MAY impose additional constraints on
@@ -1342,7 +1340,7 @@
 
     <p class="note" id="note-recognized-datatype-iris">
       In RDF 1.1, <em><span id="dfn-recognized-datatype-iris">Recognized datatype IRIs</span></em>
-      were defined RDF Concepts, overlapping with 
+      were defined in RDF Concepts, overlapping with 
       <a data-cite="RDF12-SEMANTICS#dfn-recognized">RDF Semantics, "recognizing"</a>
       datatype IRIs for <a data-cite="RDF12-SEMANTICS#dfn-semantic-extension">semantic extensions</a>.
     </p>
@@ -1469,7 +1467,7 @@
 
 <section id="section-additional-datatypes" class="appendix">
   <h2>Additional Datatypes</h2>
-  <p>This section defines additional <a>datatypes</a> that RDF processors MAY support.</p>
+  <p>This section defines additional <a>datatypes</a> that RDF implementations MAY support.</p>
 
   <section id="section-html">
     <h3>The <code>rdf:HTML</code> Datatype</h3>
@@ -1970,6 +1968,7 @@
     <li>Refer to the definition and discussion of 
       <a data-cite="RDF12-SEMANTICS#dfn-recognized">RDF Semantics, "recognizing"</a>
       datatype IRIs, instead of <em>Recognized datatype IRIs</em>.</li>
+    <li>The informal terminolgy "RDF processor" has been removed.</li>
   </ul>
 
   <p class="note">A detailed overview of the differences between RDF versions&nbsp;1.1
--- a/spec/index.html
+++ b/spec/index.html
@@ -740,28 +740,25 @@
       <li>If the literal is a <a>language-tagged string</a>,
         then the literal value is a pair consisting of its <a>lexical form</a>
         and its <a>language tag</a>, in that order.</li>
-      <li>if the literal is a <a>directional language-tagged string</a>, then the literal value is
+      <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
         a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
         likewise in that order.</li>
-
-      <li>If the literal's <a>datatype IRI</a> is in the set of
-        <a>recognized datatype IRIs</a>, let <var>d</var> be the
-        <a>referent</a> of the datatype IRI.
+      <li>If the literal's <a>datatype IRI</a> is handled by an RDF processor,
         <ul>
-          <li>If the literal's <a>lexical form</a> is in the <a>lexical space</a>
-            of <var>d</var>, then the literal value is the result of applying
-            the <a>lexical-to-value mapping</a> of <var>d</var> to the
+          <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
+            of datatype, then the literal value is the result of applying
+            the <a>lexical-to-value mapping</a> of the datatype to the
             <a>lexical form</a>.</li>
-          <li>Otherwise, the literal is ill-typed and no literal value can be
+          <li>otherwise, the literal is ill-typed and no literal value can be
              associated with the literal. Such a case produces a semantic
              inconsistency but is not <em>syntactically</em> ill-formed.
-             Implementations MUST accept ill-typed literals and produce RDF
+             Implementations SHOULD accept ill-typed literals and produce RDF
              graphs from them. Implementations MAY produce warnings when
              encountering ill-typed literals.</li>
         </ul>
       </li>
-      <li>If the literal's <a>datatype IRI</a> is <em>not</em> in the set of
-        <a>recognized datatype IRIs</a>, then the literal value is
+      <li>If the literal's <a>datatype IRI</a> is <em>not</em>
+        handled by an RDF processor then the literal value is
         not defined by this specification.</li>
     </ul>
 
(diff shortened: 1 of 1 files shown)


afs:
cc @rdfguy

gkellogg (approved):

spec/index.html
```diff
@@ -1317,12 +1312,13 @@ <h3>Datatype IRIs</h3>
...
-      MAY give a warning message if they are unable to determine the
-      referent of an IRI used in a typed literal, but they SHOULD NOT
-      reject such RDF as either a syntactic or semantic error.<p>
+    <p>RDF processors are not required to handle all datatypes.
+      Any literal typed with a datatype IRI not handled by an RDF processor
+      is treated just like an unknown IRI, i.e. as referring to an unknown thing.
+      Applications MAY give a warning message if they are unable to determine the
+      referent of an IRI used in a typed literal. RDF processors SHOULD
+      not reject literals with unknown IRI as either a syntactic or
+      semantic error.<p>
```
TallTed, line 1321:
```suggestion
    <p>RDF processors are not required to handle all datatypes.
      Any literal typed with a datatype IRI not handled by an RDF processor
      is treated just like an unknown IRI, i.e., as referring to an unknown thing.
      Applications MAY give a warning message if they are unable to determine the
      referent of an IRI used in a typed literal. RDF processors SHOULD NOT
      reject a literal with an unknown IRI as either a syntactic or
      semantic error.<p>
```

spec/index.html
```diff
@@ -1343,6 +1339,14 @@ <h3>Datatype IRIs</h3>
       user-defined simple XML Schema datatypes</a>
       is suggested in [[SWBP-XSCH-DATATYPES]]. RDF implementations
       are not required to support either of these facilities.</p>
+
+    <p class="note" id="note-recognized-datatype-iris">
+      In RDF 1.1, <em><span id="dfn-recognized-datatype-iris">Recognized datatype IRIs</span></em>
+      were defined RDF Concepts, overlapping with
```
TallTed, line 1345:
```suggestion
      In RDF 1.1, <em><span id="dfn-recognized-datatype-iris">Recognized datatype IRIs</span></em>
      were defined in RDF Concepts, overlapping with 
```

TallTed: or possibly
```suggestion
      In RDF 1.1, <em><span id="dfn-recognized-datatype-iris">Recognized datatype IRIs</span></em>
      were defined by RDF Concepts, overlapping with 
```

spec/index.html
```diff
@@ -740,28 +740,25 @@ <h3>Literals</h3>
...
-        <a>referent</a> of the datatype IRI.
+      <li>If the literal's <a>datatype IRI</a> is handled by an RDF processor,
         <ul>
-          <li>If the literal's <a>lexical form</a> is in the <a>lexical space</a>
-            of <var>d</var>, then the literal value is the result of applying
-            the <a>lexical-to-value mapping</a> of <var>d</var> to the
+          <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
+            of datatype, then the literal value is the result of applying
+            the <a>lexical-to-value mapping</a> of the datatype to the
             <a>lexical form</a>.</li>
```
TallTed, line 751:
```suggestion
            <a>lexical form</a>;</li>
```

TallTed (changes requested):
Small, editorial, mostly punctuation

spec/index.html
```diff
@@ -740,28 +740,25 @@ <h3>Literals</h3>
...
-             Implementations MUST accept ill-typed literals and produce RDF
+             Implementations SHOULD accept ill-typed literals and produce RDF
              graphs from them. Implementations MAY produce warnings when
              encountering ill-typed literals.</li>
         </ul>
       </li>
-      <li>If the literal's <a>datatype IRI</a> is <em>not</em> in the set of
-        <a>recognized datatype IRIs</a>, then the literal value is
+      <li>If the literal's <a>datatype IRI</a> is <em>not</em>
+        handled by an RDF processor then the literal value is
```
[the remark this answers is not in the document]
doerthe: I think the RDF processor needs further explanation. Is it introduced somewhere? Could we add a link?

spec/index.html
hartig: There is a slight inconsistency in this paragraph. According to the first sentence, the things that RDF processors may handle are the datatypes. Yet, according to the second sentence, it is datatype IRIs that RDF processors may (or may not) handle. We should settle on one or the other (datatypes or datatype IRIs).

My preference would be for datatypes instead of datatype IRIs. In this case, the second sentence here should be change to: "Any literal typed with a datatype IRI that denotes a datatype that is not handled by ..."

Then, the new line 746 (see above) should be changed accordingly: "If the <a>datatype</a> denoted by the literal's <a>datatype IRI</a> is handled by ..."

hartig (changes requested):

spec/index.html
```diff
@@ -1295,13 +1291,12 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
     <h3>Datatype IRIs</h3>
 
-    <p>Datatypes are identified by <a>IRIs</a>. If
-      <var>D</var> is a set of IRIs which are used to refer to
-      datatypes, then the elements of <var>D</var> are called
-      <span id="dfn-recognized-datatype-iris"><!-- obsolete term--></span><dfn data-lt="recognized datatype IRI">recognized datatype IRIs</dfn>.
-      Recognized IRIs have fixed
-      <a href="#referents">referents</a>. If any IRI of the form
-      <code>http://www.w3.org/2001/XMLSchema#xxx</code> is recognized, it
+    <p>Datatypes are identified by <a>IRIs</a> and the <a>referent</a>
```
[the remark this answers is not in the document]
pchampin: > To be consistent with the terminology in the definition of [datatype](https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/124.html#dfn-datatype) :

actually, the term "identified" is the correct one, see 1st paragraph of https://www.w3.org/TR/rdf12-semantics/#datatypes . So the definition of datatypes in rdf-concepts should probably be fixed...

spec/index.html
```diff
@@ -740,28 +740,25 @@ <h3>Literals</h3>
...
-      <li>If the literal's <a>datatype IRI</a> is in the set of
-        <a>recognized datatype IRIs</a>, let <var>d</var> be the
-        <a>referent</a> of the datatype IRI.
+      <li>If the literal's <a>datatype IRI</a> is handled by an RDF processor,
         <ul>
-          <li>If the literal's <a>lexical form</a> is in the <a>lexical space</a>
-            of <var>d</var>, then the literal value is the result of applying
-            the <a>lexical-to-value mapping</a> of <var>d</var> to the
+          <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
+            of datatype, then the literal value is the result of applying
```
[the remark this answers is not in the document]
pchampin: I think that replacing "datatype" with "the datatype" should be enough and less verbose.

If we go the verbose way, note that the correct word should be "identified" rather than "denote" (see my other comment on this below).

spec/index.html
```diff
@@ -740,28 +740,25 @@ <h3>Literals</h3>
...
         and its <a>language tag</a>, in that order.</li>
-      <li>if the literal is a <a>directional language-tagged string</a>, then the literal value is
+      <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
         a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
         likewise in that order.</li>
-
-      <li>If the literal's <a>datatype IRI</a> is in the set of
-        <a>recognized datatype IRIs</a>, let <var>d</var> be the
-        <a>referent</a> of the datatype IRI.
+      <li>If the literal's <a>datatype IRI</a> is handled by an RDF processor,
```
pchampin, line 746: I find the wording "an RDF processor" strange, all the more that it is repeated below...
Implicitly, we are referring to the *same* processor all along this paragraph, so replacing this (and the other occurrence) with "the RDF processor" would seem better.

Also, the term "RDF processor" is currently only used twice in RDF-Concepts (a little more after this PR), and only in relation with datatypes. Anywhere else in the spec, we talk about Applications and Implementations. Wouldn't it be better to stick to one of these?
```suggestion
      <li>If the literal's <a>datatype IRI</a> is handled by the implementation,
```

spec/index.html
pchampin, line 761:
```suggestion
        handled by the implementation, then the literal value is
```
(see my comment above about "an RDF processor")

spec/index.html
afs: I'd be happy to remove "RDF Processor" from RDF Concepts  - it's informal but looks formal; it is only for datatypes.

"RDF processor" is used in RDF Semantics more (5 times) so it looks like another thing that has come over from there.

spec/index.html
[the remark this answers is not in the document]
afs: RDF Semantics then says "denoted".

>Datatypes are [identified](https://www.w3.org/TR/rdf12-semantics/#dfn-identify) by IRIs. Interpretations will vary according to which IRIs are recognized as denoting datatypes.

_I think_ the point is "identified" is about the usage made in the `^^` position. If so, "identified" would be better in line 1294.

I have changed the definition "and is denoted by" to and is identified by" for consistency.

If there are further thoughts, can we deal with these in a separate issue/pr? This is orthogonal to removing "recognized datatype IRI" from RDF Concepts.

spec/index.html
afs: OK - I will make the change here and L746.

As for more general "datatype IRI" -> "datatype", please could we have a separate PR? By bundling changes into one PR, it becomes increasingly harder to ensure, and review, that it is done outside the areas touched by the initial PR.

spec/index.html
afs: For reasons unknown, the most (not all) bulleted lists in the doc have `'` on each list item. It would be better to deal with this separately because this PR is growing in scope already and I think keeping it as focused as possible will help get it accepted.

spec/index.html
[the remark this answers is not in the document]
afs: > I think that replacing "datatype" with "the datatype" should be enough and less verbose.

Agreed.

spec/index.html
```diff
@@ -1295,13 +1291,12 @@ <h3>The XML Schema Built-in Datatypes</h3>
...
-    <p>Datatypes are identified by <a>IRIs</a>. If
-      <var>D</var> is a set of IRIs which are used to refer to
-      datatypes, then the elements of <var>D</var> are called
-      <span id="dfn-recognized-datatype-iris"><!-- obsolete term--></span><dfn data-lt="recognized datatype IRI">recognized datatype IRIs</dfn>.
-      Recognized IRIs have fixed
-      <a href="#referents">referents</a>. If any IRI of the form
-      <code>http://www.w3.org/2001/XMLSchema#xxx</code> is recognized, it
+    <p>Datatypes are identified by <a>IRIs</a> and the <a>referent</a>
+      of the IRI of a datatype defines a partial mapping from the
+      lexical form to a value.</p>
```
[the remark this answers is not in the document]
afs: Removing the second part looks good. This isn't the definition of "datatype".

spec/index.html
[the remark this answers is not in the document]
hartig: Yes @afs , based on @pchampin's comment I now also agree the line 1294 should use "identified", exactly as you did in this PR. So, the only thing that remains to be done for line 1294 in this PR is to apply @TallTed's [suggestion of adding a comma](https://github.com/w3c/rdf-concepts/pull/124#discussion_r1901930143).

spec/index.html
afs: I've pushed changes that include changing "RDF processor" to "RDF implenentation".

spec/index.html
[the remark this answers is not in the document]
afs: The subsequent text has been removed as per [your suggestion](https://github.com/w3c/rdf-concepts/pull/124#discussion_r1907584379).

hartig (approved):

spec/index.html
TallTed: https://github.com/w3c/rdf-concepts/issues/127

Issue #63, opened 2023-09-13
Reference for punycode-encoding of IDN


domel:
According to the [comment](https://github.com/w3c/rdf-concepts/pull/59#discussion_r1323501083):

> The real issue is that this is probably not just about updating references. With the warning that I have not studied this thread enough to be confident that I fully understand the issues involved, let me try to explain.
> 
> If the discussion is about internatiionalized domain names (IDNs), there should be no references to RFC 3490 or 3491 (aka "IDNA2003") at all. Such references should be replaced by references to one or more of RFC 5890ff and the surrounding text should be checked carefully to make sure it is not dependent on assumptions or conventions of IDNA2003 (RFC 3490/3491) that were changed in IDNA2008 (RFC 5890ff),
> 
> Now, as far as Punycode is concerned, IDNA2008 generally, and 5891 in particular, do not change the algorithm itself at all (see RFC 5891 Section 4.4). What it changes are the references inside 3942 that point to IDNA2003 that are updatedto point to IDNA2008. However, depending on how you are using "punycode" terminology, that is fairly important. The issues goes back to the core differences between IDNA2003 and IDNA2008. Conceptually, IDNA2008 is not a tuned and updated version of IDNA2003. It is as much of a change as the change of the default CCS for HTML from ISO 8859-1 to Unicode. In the particular case of IDNs, one of those conceptual differences is that IDNA2003 was quite permissive as to what names were valid, even allowing different Unicode strings to convert to the same string after Punycode conversion although (obviously) the reverse conversion could only produce one of those strings. IDNA2008 is much more focused on identifier integrity and restrictions intended to reduce or eliminate ambiguity, mapping problems, etc. As a result, the "to_asciii" and "to_unicode" terminology and, if one needs to be precise, "Punycode string" terminology, are no longer useful and, IIR, do not appear in the IDNA2008 documents at all. And strings that can be generated by the Punycode algorithm that are not valid IDNA2008 labels are, to use a technical term, garbage: there is no way to talk about their validity. So, in general, contemporary documents should not be referring to Punycode at all, especially as a type of string, but should be using the IDNA2008 terminology of "A-labels" and "U-labels", defined in RFC 5890.

I think that it's not only the issue to change from

> Punycode-encoding of Internationalized Domain Names in IRIs [[RFC3492](https://w3c.github.io/rdf-concepts/spec/#bib-rfc3492)]

to 

[...]

PR #132, opened 2025-01-16
Make the IRI advice list be positive form advice

afs:
The IRI advice list had some negative forms which could be misread because other points were positive form.


gkellogg (approved):

hartig (approved):

spec/index.html
```diff
@@ -643,27 +643,26 @@ <h3>IRIs</h3>
...
+          is preferred over <code>http://example:80/</code>.</li>
+        <li>Use uppercase hexadecimal letters within percent-encoding
           triplets (i.e., “<code>%3F</code>” is preferred over
-          “<code>%3f</code>”)</li>
-        <li>Punycode-encoding of Internationalized Domain Names
-          in IRIs [[RFC3492]]</li>
-        <li>IRIs that are not in Unicode [=Normalization Form C=] [[I18N-Glossary]]</li>
+          “<code>%3f</code>”).</li>
+        <li>An empty path in an HTTP IRI
+          <code>http://example/</code> is perferred over
```
TallTed, line 659:
```suggestion
          <code>http://example/</code> is preferred over
```

TallTed (changes requested):
It might be worth putting all the `preferred` examples into parentheses, or removing all those parentheses, and making the punctuation the same. (See lines 653-660)

afs:
> It might be worth putting all the `preferred` examples into parentheses, or removing all those parentheses, and making the punctuation the same. (See lines 653-660)

Done.

spec/index.html
TallTed: This typo has not been fixed. Why is it marked "resolved"?

TallTed:
Typo of `perferred` remains, though it is marked "resolved".

Please don't mark anything "resolved" unless you've clicked the COMMIT button and/or I've confirmed that you've made changes that resolve my concern.

afs:
Fixed.

spec/index.html
```diff
@@ -648,27 +648,26 @@ <h3>IRIs</h3>
...
+        <li>Use uppercase hexadecimal letters within percent-encoding
+          triplets;"<code>%3F</code>" is preferred over
+          "<code>%3f</code>".</li>
+        <li>An empty path in an HTTP IRI
+          <code>http://example/</code> is preferred over
+          having no path <code>http://example</code>.</li>
+        <li>Normalize IRIs to remove "<code>/./</code>" and "<code>/../</code>" in the path
+          component of an IRI.</li>
+        <li>Avoid using punycode-encoding of Internationalized Domain Names 
+          in IRIs [[RFC5890]].</li>
```
aphillips, line 669: This should probably say:

```suggestion
        <li>Avoid using the A-label (ASCII, punycode-encoded name) for Internationalized Domain Names [[RFC5890]] in IRIs.</li>
```

spec/index.html
```diff
@@ -648,27 +648,26 @@ <h3>IRIs</h3>
...
-          component of an IRI</li>
-        <li>Lowercase hexadecimal letters within percent-encoding
-          triplets (i.e., “<code>%3F</code>” is preferred over
-          “<code>%3f</code>”)</li>
-        <li>Punycode-encoding of Internationalized Domain Names
-          in IRIs [[RFC3492]]</li>
-        <li>IRIs that are not in Unicode [=Normalization Form C=] [[I18N-Glossary]]</li>
+        <li>Use lowercase characters in scheme names and domain names.</li>
+        <li>Only use percent-encoding of characters where 
+          necessary for IRI syntax.</li>
```
aphillips, line 657: Probably should should be slightly different:

```suggestion
        <li>Only use percent-encoding of characters where
          required by the IRI syntax.</li>
```

spec/index.html
```diff
@@ -648,27 +648,26 @@ <h3>IRIs</h3>
...
-          <code>http://example.com/</code> is preferable</li>
-        <li>“<code>/./</code>” or “<code>/../</code>” in the path
-          component of an IRI</li>
-        <li>Lowercase hexadecimal letters within percent-encoding
-          triplets (i.e., “<code>%3F</code>” is preferred over
-          “<code>%3f</code>”)</li>
-        <li>Punycode-encoding of Internationalized Domain Names
-          in IRIs [[RFC3492]]</li>
-        <li>IRIs that are not in Unicode [=Normalization Form C=] [[I18N-Glossary]]</li>
+        <li>Use lowercase characters in scheme names and domain names.</li>
```
aphillips, line 655: Some care needs to be used here. IDNs are not necessarily case-insensitive outside the ASCII range. Mapping the U-label to lowercase might result in a domain name that does not work.

```suggestion
        <li>Use lowercase ASCII characters in scheme names and, where applicable, domain names.
           Non-ASCII domain names should be chosen carefully.</li>
```

aphillips (commented):
Unfortunately, @xfq was on vacation when this PR was being reviewed. I made some comments after the fact and will call them out in #63.

spec/index.html
gkellogg: Just to be clear, we’re using UTF-8, but limited to the ASCII range – we’re not using an ASCII character set.

afs: We could split this into scheme advice and domain name advice if they have different text.

I agree with @gkellogg's point about UTF-8/ASCII . For an advice note, is it better to say "ASCII" or is there Unicode-terminology - I found is "Basic Latin" but I don't think that is helpful.

== _scheme name advice_

Scheme names are restricted by syntax anyway so "lowercase" is unambiguous.

With or without "ASCII" --

_text:_  * Use lowercase ASCII characters in [scheme names](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1).

== _domain name advice_

For an advice note, "chosen carefully", is a bit non-committal :smile:. 

What is the state of play in DNS names? Does the upper/lower case advice apply to characters outside basic latin?

_text:_  * Use lowercase ASCII characters in domain names in preference to uppercase ASCII characters.

Possible link: The https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2 but the key text is not exactly at that point.

aphillips: So between making this comment last night and this morning I double checked the state of IDNA wrt non-ASCII case-sensitivity to ensure I am giving the proper advice. 

Basically, ASCII letters in domain names are case-insensitive (and guidance to lowercase these is the right course of action). Non-ASCII characters are case-sensitive, because upper and lowercase letters encode differently into Punycode and because case folding is at least sometimes locale-sensitive. This is all in RFC5890.

However, RFC5892, which says which code points can be registered, includes section 2.2, which says (in the most oblique and indirect way possible) that uppercase letters (among others) should be avoided.

You're right that scheme names are already restricted to ASCII, so stating ASCII for them is unnecessary (except, perhaps, as a reminder that the letters are restricted to ASCII??). 

For domain names, _only the ASCII letters_ are case-insensitive. If a domain name is registered, in spite of the guidance noted above, with an uppercase non-ASCII letter, that name will encode to a different A-label (and is thus distinct in DNS).

> For an advice note, "chosen carefully", is a bit non-committal 😄.

Yeah, that's an exceptionally lame bit of guidance.

> For an advice note, is it better to say "ASCII" or is there Unicode-terminology - "Basic Latin", but I don't think that is helpful.

ASCII is usually the clearest way to talk about this. "Basic Latin" is Unicode-ese, but only Unicadetti will make the connection to "code points less than 0x80"

Perhaps having two bullets would simplify matters?

```suggestion
        <li>Use lowercase characters in scheme names.</li>
        <li>Use lowercase characters in domain names. 
         Note that non-ASCII characters in domain names are case-sensitive [[RFC5890]], 
         but that domains are generally only registered with lowercase letters [[RFC5892]].</li>
```

afs: Part of #155.

spec/index.html
afs: Done in #155.

spec/index.html
afs: Part of #155 with A-label as a link to RFC5890 section 2.3.2.1

Issue #128, opened 2025-01-16
update to graph isomorphism


pfps:
As far as I can tell, graph isomorphism needs the following extra clause

- M(tt) is the triple term ( M(s), M(p), M(o) ) for tt a triple term of the form (s, p, o)

This is slightly clunky but doesn't use any concrete syntax.

hartig:
I was just about to implement the change as proposed in [the issue description above](https://github.com/w3c/rdf-concepts/issues/128#issue-2793376260). Yet, it turns out that it may not be exactly that simple, for the following reasons: By the current definition, the domain of the bijection_M_ is the set of all [nodes](https://w3c.github.io/rdf-concepts/spec/#dfn-node) of the graph _G_. Yet, it is not guaranteed that any of the three elements of the triple term _tt_ = (_s_, _p_, _o_) is such a node. Therefore, _M_(_s_) may be undefined, and so may _M_(_p_) and _M_(_o_).

The fix to this problem may be to change the domain and the co-domain of the bijection _M_ to both be the set of all RDF terms. With this change, the extra clause mention in [the issue description above](https://github.com/w3c/rdf-concepts/issues/128#issue-2793376260) can then be added without problems. So, to be explicit, in addition to adding this extra clause we need to change the sentence before the bullet points in the section. I propose to change that sentence as follows:

> Two [RDF graphs](https://w3c.github.io/rdf-concepts/spec/#dfn-rdf-graph) _G_ and _G'_ are isomorphic (that is, they have an identical form) if there is a bijection _M_ from the set of all [RDF terms](https://w3c.github.io/rdf-concepts/spec/#dfn-rdf-term) into that same set, such that all of the following properties hold:

I will implement this change once PR #137 is merged.

pfps:
Good catch

pchampin:
Also, note that M(p) is not necessary in the recursion (p is always an IRI, so M(p) is always equal to p).

PR #140, opened 2025-01-17
Updates the definition of graph isomorphism

hartig:
.. as proposed in #128, including the additional change proposed in https://github.com/w3c/rdf-concepts/issues/128#issuecomment-2598233599

Notice that I also removed the "that is a node of G" part of bullet point 2 and 3. This change is needed for the new recursive part of the definition to be correct (i.e., when applying _M_ to IRIs or literals inside triple terms).

Additionally, I also changed the last (now fifth) bullet point by putting _M_(_p_) in the second triple mentioned in this bullet point. While this change is not strictly necessary (because _p_ is an IRI and _M_ maps each IRI to itself), using _M_ here looks a bit more consistent.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/140.html" title="Last updated on Jan 17, 2025, 4:28 PM UTC (b906255)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/140/4d1cc8c...b906255.html" title="Last updated on Jan 17, 2025, 4:28 PM UTC (b906255)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -961,11 +961,9 @@
 
     <ul>
       <li><var>M</var> maps blank nodes to blank nodes.</li>
-      <li><var>M</var>(<var>lit</var>)=<var>lit</var> for every <a>RDF literal</a> <var>lit</var> that
-        is a node of <var>G</var>.</li>
+      <li><var>M</var>(<var>lit</var>)=<var>lit</var> for every <a>RDF literal</a> <var>lit</var>.</li>
 
-      <li><var>M</var>(<var>iri</var>)=<var>iri</var> for every <a>IRI</a> <var>iri</var>
-        that is a node of <var>G</var>.</li>
+      <li><var>M</var>(<var>iri</var>)=<var>iri</var> for every <a>IRI</a> <var>iri</var>.</li>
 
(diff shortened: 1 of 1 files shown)


pchampin (approved):
LGTM.

I notice that M was also applied to the predicate of (asserted) triples and triple terms, whereas the original definition kept _p_ as is. Since _p_ is always an IRI, this is absolutely equivalent, but more verbose...

I have a slight preference for the terse version (_p_ instead of M(_p_)), but I can live with both.

hartig:
> I have a slight preference for the terse version (p instead of M(p)), but I can live with both.

I'm fine with changing it back to _p_ if others prefer that as well. However, in this case, I would also change the _M_(_p_) in the new fourth bullet point to _p_. Just to be consistent.

afs:
I prefer `M(p)` because it extends to generalized RDF (and symmetric RDF if we have to add blank node predicates for sementics).

Also, it looks clearer to me because it is mapping `p` even if it only ever results in `p`.

afs (approved):

pfps (approved):
Looks good to me

gkellogg (approved):

hartig:
Given the approvals of all other editors, and also of @pfps who created the issue that this PR addresses, I will merge the PR now.

PR #151, opened 2025-01-24
update RDF dataset discussion with explicit comparison of `{ RDF | SPARQL } { 1.2 | 1.1 | 1.0 }`

TallTed:
Add explicit comparison of SPARQL 1.2 and RDF 1.2, vs SPARQL 1.1, RDF 1.1, and RDF 1.0. Might also want specific difference with SPARQL 1.0.

This specificity may not be needed for users who have never before encountered RDF or SPARQL, but I think it helpful for users who *have* worked with any previous version(s) — like me.

***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/rdf-concepts/pull/151.html" title="Last updated on Mar 3, 2025, 8:47 PM UTC (4e40fdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/151/f231b67...TallTed:4e40fdd.html" title="Last updated on Mar 3, 2025, 8:47 PM UTC (4e40fdd)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1021,11 +1021,13 @@
       can avoid interoperability issues by not ascribing importance to
       the presence or absence of empty named graphs.</p>
 
-    <p>SPARQL 1.2 [[SPARQL12-QUERY]] uses 
+    <p>Version 1.2 of the SPARQL Query Language [[SPARQL12-QUERY]] uses 
       <a data-cite="SPARQL12-QUERY#sparqlDataset">the same concept of an RDF Dataset</a>
+      as RDF versions 1.1 and 1.2,
       in which the graph names of named graphs may be IRIs or blank nodes.
       In contrast, version 1.1 of the SPARQL Query Language [[SPARQL11-QUERY]]
-      only allows IRIs as graph names.</p>
+      and RDF version 1.0
+      only allowed graph names to be IRIs.</p>
 
   </div>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -1026,7 +1026,6 @@
       as RDF versions 1.1 and 1.2,
       in which the graph names of named graphs may be IRIs or blank nodes.
       In contrast, version 1.1 of the SPARQL Query Language [[SPARQL11-QUERY]]
-      and RDF version 1.0
       only allowed graph names to be IRIs.</p>
 
   </div>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1021,7 +1021,7 @@
       can avoid interoperability issues by not ascribing importance to
       the presence or absence of empty named graphs.</p>
 
-    <p>Version 1.2 of the SPARQL Query Language [[SPARQL12-QUERY]] uses 
+    <p>SPARQL version 1.2 [[SPARQL12-CONCEPTS]] uses 
       <a data-cite="SPARQL12-QUERY#sparqlDataset">the same concept of an RDF Dataset</a>
       as RDF versions 1.1 and 1.2,
       in which the graph names of named graphs may be IRIs or blank nodes.


gkellogg:
Note that we do discuss changes to RDF in the Abstract:

> RDF 1.2 introduces [triple terms](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-term) as another kind of [RDF term](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-term) which can be used as the [object](https://www.w3.org/TR/rdf12-concepts/#dfn-object) of another [triple](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-triple). RDF 1.2 also introduces [directional language-tagged strings](https://www.w3.org/TR/rdf12-concepts/#dfn-dir-lang-string), which contain a [base direction](https://www.w3.org/TR/rdf12-concepts/#dfn-base-direction) element that allows the initial text direction to be specified for presentation by a user agent.

> RDF 1.2 Concepts introduces key concepts and terminology for RDF 1.2, discusses datatyping, and the handling of [fragment identifiers](https://www.w3.org/TR/rdf12-concepts/#dfn-fragment-identifier) in IRIs within RDF graphs.

RDF 1.0 did not have named graphs, and therefore a discussion of the use of IRIs for graph names is not appropriate. (SPARQL 1.0 had it's own concept of named graphs).

hartig (approved):

spec/index.html
```diff
@@ -1021,11 +1021,13 @@ <h2>RDF Datasets</h2>
       can avoid interoperability issues by not ascribing importance to
       the presence or absence of empty named graphs.</p>
 
-    <p>SPARQL 1.2 [[SPARQL12-QUERY]] uses 
+    <p>Version 1.2 of the SPARQL Query Language [[SPARQL12-QUERY]] uses
```
afs, line 1024: It's all SPARQL specs which is why it says SPARQL 1.2. 

The link is provided because that is the starting point. Arguably, the link could be removed.

spec/index.html
```diff
@@ -1021,11 +1021,13 @@ <h2>RDF Datasets</h2>
...
 
-    <p>SPARQL 1.2 [[SPARQL12-QUERY]] uses 
+    <p>Version 1.2 of the SPARQL Query Language [[SPARQL12-QUERY]] uses 
       <a data-cite="SPARQL12-QUERY#sparqlDataset">the same concept of an RDF Dataset</a>
+      as RDF versions 1.1 and 1.2,
       in which the graph names of named graphs may be IRIs or blank nodes.
       In contrast, version 1.1 of the SPARQL Query Language [[SPARQL11-QUERY]]
-      only allows IRIs as graph names.</p>
+      and RDF version 1.0
+      only allowed graph names to be IRIs.</p>
```
afs, line 1030: Better to remove this. 

RDF 1.0 didn't mention datasets at all nor were they concurrent.
This is a note. RDF 1.1 is 10 years old.

spec/index.html
TallTed: > The link is provided because that is the starting point. Arguably, the link could be removed.

It seems odd for [[SPARQL12-QUERY](https://w3c.github.io/sparql-query/spec/)] to be the starting point.

[[SPARQL12-CONCEPTS](https://w3c.github.io/sparql-concepts/spec/)] or *possibly* [[SPARQL12-NEW](https://w3c.github.io/sparql-new/spec/)] makes more sense to me, and not only because these are listed before [[SPARQL12-QUERY](https://w3c.github.io/sparql-query/spec/)] in the **[List of Documents](https://w3c.github.io/sparql-query/spec/#related)** (to which I cannot find a way to link directly, nor to auto-expand after load)

> It's all SPARQL specs which is why it says SPARQL 1.2.

It seems to me, this lends weight to linking to [[SPARQL12-CONCEPTS](https://w3c.github.io/sparql-concepts/spec/)] or *possibly* [[SPARQL12-NEW](https://w3c.github.io/sparql-new/spec/)] instead of [[SPARQL12-QUERY](https://w3c.github.io/sparql-query/spec/)].

Removing the link is indeed also an option.

TallTed:
> Note that we do discuss changes to RDF in the Abstract:

I don't know about you, but when I am looking things up in a spec, and even when I am trying to read the entire document, I often overlook things in the Abstract (because I don't understand enough to absorb them properly) — and may even skip it entirely, because I expect it to echo things said later but at a higher level of, sorry, abstraction. For these reasons, I suggest that anything we mean for the reader to actively digest should be placed elsewhere in the document, even if it is also said (preferably not word-for-word) in the Abstract.

> RDF 1.0 did not have named graphs, and therefore a discussion of the use of IRIs for graph names is not appropriate. (SPARQL 1.0 had it's own concept of named graphs).

My bad, probably. I should have re-re-re-re-re-re-read the RDF 1.0 & 1.1 and SPARQL 1.0 & 1.1 docs more closely. (In fact, I did go back to those docs, and based my own writing on what they said. I should have linked to my source material. I'll see if I can re-find it.)

spec/index.html
```diff
@@ -1021,11 +1021,13 @@ <h2>RDF Datasets</h2>
...
       the presence or absence of empty named graphs.</p>
 
-    <p>SPARQL 1.2 [[SPARQL12-QUERY]] uses 
+    <p>Version 1.2 of the SPARQL Query Language [[SPARQL12-QUERY]] uses 
       <a data-cite="SPARQL12-QUERY#sparqlDataset">the same concept of an RDF Dataset</a>
+      as RDF versions 1.1 and 1.2,
       in which the graph names of named graphs may be IRIs or blank nodes.
       In contrast, version 1.1 of the SPARQL Query Language [[SPARQL11-QUERY]]
-      only allows IRIs as graph names.</p>
+      and RDF version 1.0
```
pchampin, line 1029:
```suggestion
```

pchampin (dismissed):
Apart from the erroneous reference to named graphs in RDF 1.0, LGTM.

spec/index.html
TallTed: Does this work for you, @afs?
```suggestion
    <p>SPARQL Version 1.2 [[SPARQL12-CONCEPTS]] uses 
```

spec/index.html
TallTed: @afs -- Does this work for you, as also suggested by @pchampin?
```suggestion
      only allowed graph names to be IRIs.</p>
```

afs: Yes.

spec/index.html
afs: Yes - though "version" not "Version".

spec/index.html
TallTed: done in [4e40fdd](https://github.com/w3c/rdf-concepts/pull/151/commits/4e40fdd369ca046550bdc95535d4fcbad60828b8)

afs (approved):

gkellogg (approved):

pchampin (approved):

PR #158, opened 2025-02-13
unify terminology for triples

pchampin:
redefine "RDF triple" so that "triple term" and "asserted triple" are two different roles for triples,
rather than objects of different nature, per my action w3c/rdf-star-wg#145.

This also induced some changes in Section 1 "Introduction", but the rest of the text works well with this change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/158.html" title="Last updated on Feb 18, 2025, 4:43 PM UTC (b5cc7fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/158/df7b9db...b5cc7fd.html" title="Last updated on Feb 18, 2025, 4:43 PM UTC (b5cc7fd)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -416,7 +416,7 @@
   <section id="entailment">
     <h3>Equivalence, Entailment and Inconsistency</h3>
 
-    <p>An <a>RDF triple</a> encodes a <dfn class="export">proposition</dfn>—a
+    <p>An <a>RDF triple</a> encodes a <dfn class="export">proposition</dfn> — a
       simple logical expression, describing a relationship between two entities.
       An [=asserted triple=] is a claim that the corresponding proposition is true.
       An <a>RDF graph</a> is the conjunction (logical <em>AND</em>) of
@@ -576,7 +576,7 @@
     <p>An [=RDF triple=] used as the [=object=] of another [=triple=] is called a <dfn class=export >triple term</dfn>.
       An [=RDF triple=] is said to be <dfn id="dfn-asserted-triple" class=export data-lt="asserted triple|asserted triples" data-lt-no-plural=""
       >asserted</dfn> in an [=RDF graph=] if it is an element of the [=RDF graph=].
-      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or both.
+      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
     </p>
 
     <p><a>IRIs</a>, <a>literals</a>,
--- a/spec/index.html
+++ b/spec/index.html
@@ -884,7 +884,7 @@
 
     <ul>
       <li>The [=subject=], [=predicate=] and [=object=] of |t| [=appear=] in |t|.
-      <li>If the [=object=] |o| of |t| is a [=triple term=], any [=RDF term=] [=appearing=] in |o| also [=appears=] in |t|.
+      <li>If a [=triple term=] |t2| appears in |t|, then any [=RDF term=] [=appearing=] in |t2| also [=appears=] in |t|.
     </ul>
 
     <p>By extension, an [=RDF term=] is said to [=appear=] in an [=RDF graph=] if it appears in an [=asserted triple=] of that graph. An [=RDF triple=] is said to [=appear=] in an [=RDF graph=] if it is either an [=asserted triple=] of that graph or a [=triple term=] [=appearing=] in that graph.</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -585,7 +585,7 @@
   </section>
 
   <section id="section-terms">
-    <h3>RDF terms</h3>
+    <h3>RDF Terms</h3>
 
     <p><a>IRIs</a>, <a>literals</a>,
       <a>blank nodes</a>, and <a>triple terms</a> are collectively known as
@@ -874,7 +874,7 @@
   </section>
 
   <section id="section-triple-terms">
-    <h3>Triple terms</h3>
+    <h3>Triple Terms</h3>
 
     <p>An [=RDF triple=] used as the [=object=] of another [=triple=] is called a <dfn class=export >triple term</dfn>.
       In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
--- a/spec/index.html
+++ b/spec/index.html
@@ -553,7 +553,7 @@
 
     <p>An <dfn data-local-lt="triple" class="export">RDF triple</dfn>
      (often simply called "triple")
-     is a 3-tuple that is defined recursively as follows:</p>
+     is a 3-tuple that is defined inductively as follows:</p>
 
     <ul>
       <li>
@@ -596,7 +596,7 @@
       It is possible for a [=predicate=] [=IRI=] to also occur as a [=node=] in
       the same graph.</p>
 
-    <p>The set of [=RDF terms=] <dfn class=export data-lt="appear">appearing</dfn> in an [=RDF triple=] |t| is defined recursively as follows:</p>
+    <p>The set of [=RDF terms=] <dfn class=export data-lt="appear">appearing</dfn> in an [=RDF triple=] |t| is defined inductively as follows:</p>
 
     <ul>
       <li>The [=subject=], [=predicate=] and [=object=] of |t| [=appear=] in |t|.
--- a/spec/index.html
+++ b/spec/index.html
@@ -548,6 +548,10 @@
 
   <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
+  <p>An [=RDF triple=] is said to be <dfn id="dfn-asserted-triple" class=export data-lt="asserted triple|asserted triples" data-lt-no-plural=""
+      >asserted</dfn> in an [=RDF graph=] if it is an element of the [=RDF graph=].
+  </p>
+
   <section id="section-triples">
     <h3>Triples</h3>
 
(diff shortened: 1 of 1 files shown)


pchampin:
@franconi @pfps @doerthe I took the initiative to redefine "appears in" in RDF-concepts, while defining triple terms.
It seems that this is the natural "home" for this definition (also, because 'unstar' will need it, eventually).

https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/158.html#dfn-appear

There is one change compared to the one that his currently in RDF-Semantics:
it first defines "term appears in triple", then extends it to "term appears in graph", and "triple appears in graph" (as asserted or triple-term).

WDYT?

franconi:
The terminology defined in [issue #87](https://github.com/w3c/rdf-semantics/issues/87) and [PR #91](https://github.com/w3c/rdf-semantics/pull/91) (In Semantics) should be used to support the unification of the terminology of triples.

franconi:
> @franconi @pfps @doerthe I took the initiative to redefine "appears in" in RDF-concepts, while defining triple terms. It seems that this is the natural "home" for this definition (also, because 'unstar' will need it, eventually).
> 
> https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/158.html#dfn-appear
> 
> There is one change compared to the one that his currently in RDF-Semantics: it first defines "term appears in triple", then extends it to "term appears in graph", and "triple appears in graph" (as asserted or triple-term).
> 
> WDYT?

I'd use "inductively" instead of "recursively".

afs (approved):
+ @hartig 's suggestions

pchampin:
@franconi 

> The terminology defined in [issue #87](https://github.com/w3c/rdf-semantics/issues/87) and [PR #91](https://github.com/w3c/rdf-semantics/pull/91) (In Semantics) should be used to support the unification of the terminology of triples.

This was partially done:  Section 1.8 now states that "An RDF triple encodes a **proposition**" (it used to say "statement").

I didn't introduce the term "fact", partly because I was under the impression that it was still disputed, partly because I didn't find any natural place to introduce it.

hartig (approved):

afs:
"fact" would work for me. "facts" can be disputed etc.

spec/index.html
```diff
@@ -417,12 +416,14 @@ <h3>Working with Multiple RDF Graphs</h3>
   <section id="entailment">
     <h3>Equivalence, Entailment and Inconsistency</h3>
 
-    <p>An <a>RDF triple</a> encodes a <a>statement</a>—a
-      simple <dfn class="export">logical expression</dfn>, or claim about the world.
+    <p>An <a>RDF triple</a> encodes a <dfn class="export">proposition</dfn>—a
```
TallTed, line 419:
```suggestion
    <p>An <a>RDF triple</a> encodes a <dfn class="export">proposition</dfn> — a
```

spec/index.html
```diff
@@ -547,21 +548,37 @@ <h2>RDF Graphs</h2>
...
+        If |s| is an [=IRI=] or a [=blank node=],
+        |p| is an [=IRI=], and
+        |o| is an [=IRI=], a [=blank node=], or a [=literal=],
+        then (|s|, |p|, |o|) is an [=RDF triple=].
+      </li>
+
+      <li>
+        If |s| is an [=IRI=] or a [=blank node=],
+        |p| is an [=IRI=], and
+        |o| is an [=RDF triple=],
```
TallTed, line 569: Why not?
```suggestion
        |o| is an [=IRI=], a [=blank node=], a [=literal=], or an [=RDF triple=],
```

spec/index.html
```diff
@@ -547,21 +548,37 @@ <h2>RDF Graphs</h2>
...
+        then (|s|, |p|, |o|) is an [=RDF triple=].
+      </li>
     </ul>
 
+    <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
+
+    <p>An [=RDF triple=] used as the [=object=] of another [=triple=] is called a <dfn class=export >triple term</dfn>.
+      An [=RDF triple=] is said to be <dfn id="dfn-asserted-triple" class=export data-lt="asserted triple|asserted triples" data-lt-no-plural=""
+      >asserted</dfn> in an [=RDF graph=] if it is an element of the [=RDF graph=].
+      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or both.
```
TallTed, line 579: Either...
```suggestion
      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or as both.
```

TallTed: ...or...
```suggestion
      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
```

TallTed (changes requested):
Small things.

spec/index.html
hartig: Because all these other options are already covered by the first point.

TallTed: Well, yes, my point was "why two `<li>`? why not fold them together?" 

There's no difference between them, except the nested `[=RDF triple=]`. maybe I should have replaced the entirety of the two `<li>` with one `<li>`, instead of replacing the last line of the first `<li>` through the last line of the second `<li>` with one last line that combines the two last lines....

hartig: Because "[a recursive definition must always have base cases, cases that satisfy the definition _without_ being defined in terms of the definition itself](https://en.m.wikipedia.org/wiki/Recursive_definition#Form_of_recursive_definitions)." In the given definition, the first bullet point is that base case.

pchampin: what @hartig said. Otherwise, it wouldn't be a well-formed inductive/recursive definition.

TallTed (approved):

TallTed:
I will live with "facts". It still doesn't make me happy, and I have hopes that someone will have a brilliant better idea, some day. But I won't block on that basis.

afs:
The new "triples" section is doing a lot of work. Because triples-as-objects are new to RDF 1.2, the all-in-one triples section 
makes it look like a bigger change than it is. The core idea of triples hasn't changed.

> "triple term" and "asserted triple"  are two different roles for triples,

There could still be a section on "Triple terms", starting 

"""
An RDF triple used as the object of another triple is called a triple term.
"""

then that section includes term equality for triple terms and the second ("rdf:reifies").

Should there be a subsection "Graphs"?

At the moment there is intro [one line](https://www.w3.org/TR/rdf12-concepts/#section-rdf-graph). A graph subsection would include "set of nodes"  and "appears" and "graph comparison".

 The "3. Graphs" text would be introducing the section, and maybe the place define "RDF term".

pchampin:
@afs I like your suggestions, but do we agree that they should not hold this PR? (given that you have approved the PR, I assume we agree on this, but just checking)

doerthe:
I am fine with the changes, only wonder whether that means that we need to extend the definition for symmetric RDF as we use it in RDF semantics. Or should we do that here?

afs:
> @afs I like your suggestions, but do we agree that they should not hold this PR? (given that you have approved the PR, I assume we agree on this, but just checking)

That was before I saw the impact of triple term equality - part of what makes the "triples" section bigger.

afs (commented):
> @afs I like your suggestions, but do we agree that they should not hold this PR? (given that you have approved the PR, I assume we agree on this, but just checking)

That was before I saw the impact of triple term equality - part of what makes the "triples" section bigger.

I'm not blocking this merging. If it is only as an intermediate, and removes the "triple term" section, which goes back in again, it might be easier to sketch the final form here first.

afs:
> I am fine with the changes, only wonder whether that means that we need to extend the definition for symmetric RDF as we use it in RDF semantics. Or should we do that here?

We should take that discussion to a separate issue.

At
https://www.w3.org/TR/rdf12-semantics/#:~:text=ex%3Ad%20_%3Ab%20ex%3Ac%20.%20by%20rdfs7
(Appendix A - non-normative) there is a blank node in the predicate position which isn't symmetric RDF (as it is today).

https://www.w3.org/TR/rdf12-semantics/#notation talks about generalized RDF.

pchampin:
> I am fine with the changes, only wonder whether that means that we need to extend the definition for symmetric RDF as we use it in RDF semantics. Or should we do that here?

In my opinion the definitions of symmetric/generalized RDF do not need to be changed. They still "work" with the new perspective on triples/triple terms/asserted triples.

pchampin:
> If it is only as an intermediate, and removes the "triple term" section, which goes back in again, it might be easier to sketch the final form here first.

Fair enough. I just pushed a change, which is exclusively moving text, without changing it (with one exception, see below).

The text that was before in the 'Triples' section is now moved as follows:

* the definition of 'asserted triple' is now introduced just after the definition of RDF graph
* the generic notions about terms ('RDF term', 'nodes', terms are distinguishable) are now in a new 'RDF Terms' section, before the 'IRI' section
* the definition of 'triple term' and 'appears', as well as the note recommending to only use triple terms with rdf:reifies, are now in the resurrected 'Triple terms' section

The only change in the text is in the note about recursive triples not allowing cycles: it was talking about "triple terms", now talks about "triples" (in fact, I should have done this change before).

Re-requesting reviews.

spec/index.html
```diff
@@ -547,21 +548,45 @@ <h2>RDF Graphs</h2>
...
+
+
+    <p class="note">The definition of <a>triple</a> is recursive.
+      That is, a <a>triple</a> can itself have an
+      <a>object</a> component which is another <a>triple</a>.
+      However, by this definition, cycles of <a>triples</a> cannot be created.</p>
+  </section>
+
+  <section id="section-terms">
+    <h3>RDF terms</h3>
```
hartig, line 588:
```suggestion
    <h3>RDF Terms</h3>
```
To be consistent with the way the other section titles are capitalized.

spec/index.html
```diff
@@ -849,51 +874,28 @@ <h3>Blank Nodes</h3>
   </section>
 
   <section id="section-triple-terms">
-    <h3>Triple Terms</h3>
+    <h3>Triple terms</h3>
```
hartig, line 877:
```suggestion
    <h3>Triple Terms</h3>
```
To be consistent with the way the other section titles are capitalized.

hartig (approved):
I am happy with moving things around as done in the last commit.

I have only two minor issues that are easily fixed.

doerthe:
> > I am fine with the changes, only wonder whether that means that we need to extend the definition for symmetric RDF as we use it in RDF semantics. Or should we do that here?
> 
> In my opinion the definitions of symmetric/generalized RDF do not need to be changed. They still "work" with the new perspective on triples/triple terms/asserted triples.

I agree, my concern was different: we defined the notion of "appears in" to deal with the entailment patterns which assume symmetric RDF. Now, after we moved that definition to RDF concepts, it does not cover the case that triple terms could occur in subject position. My question was now: how do we best deal with that detail?

pchampin:
> we defined the notion of "appears in" to deal with the entailment patterns which assume symmetric RDF. Now, after we moved that definition to RDF concepts, it does not cover the case that triple terms could occur in subject position. My question was now: how do we best deal with that detail?

oh! Now I get it. :face_with_diagonal_mouth: 
The way I see it, either we extend the definition of "appears in" in Section 7, or RDF-Semantics extends the definition of appears in for its own purpose. I slightly prefer the latter option.

pchampin:
> The way I see it, either we extend the definition of "appears in" in Section 7, or RDF-Semantics extends the definition of appears in for its own purpose. I slightly prefer the latter option.

In fact, another option is to make the definition of "appears" more general (which is how it is defined in RDF-Semantics), i.e. replacing

> If the object o of t is a triple term, any RDF term appearing in o also appears in t. 

with

> If the triple term t2 appear in t, then any RDF term appearing in t2 also appears in t.

This is maybe a little less legible, but generalizes nicely to symmetric and generalize RDF.

doerthe:
I like that option.

hartig:
I am happy with that option as well.

hartig (approved):

afs (approved):

TallTed (approved):

Issue #157, opened 2025-02-13
Graph isomorphism and Dataset isomorphism should be unified


pchampin:
The two notions are very similar, but currently defined in slightly different ways, which is confusing.
Furthermore, the definition for Datasets has not been adapted to handle triple terms yet.

PR #159, opened 2025-02-13
unify definitions of graph/dataset isomorphism

pchampin:
address #157

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -899,26 +899,25 @@
   <section id="graph-isomorphism">
     <h3>Graph Comparison</h3>
 
+    <p>Given a bijection |M| from [=blank nodes=] to [=blank nodes=],
+      we can extend |M| to map [=RDF terms=] to [=RDF terms=],
+      [=RDF triples=] to [=RDF triples=],
+      and [=RDF graphs=] to [=RDF graphs=],
+      as follows:
+
+    <ul>
+      <li>|M|(|iri|)=|iri| if |iri| is an [=IRI=].
+      <li>|M|(|lit|)=|lit| if |lit| is a [=literal=].
+      <li>|M|(|t|)=( |M|(|s|), |M|(|p|), |M|(|o|) ) if |t| is an [=RDF triple=] of the form ( |s|, |p|, |o| ).
+      <li>|M|(|G|)={ |M|(|t|<sub>1</sub>), ..., |M|(|t|<sub>|n|</sub>) } if |G| = { |t|<sub>1</sub>, ..., |t|<sub>|n|</sub> } is an [=RDF graph=].
+    </ul>
+
     <p id="section-graph-equality">Two
       <a>RDF graphs</a> <var>G</var> and <var>G'</var> are
       <dfn data-lt="graph isomorphism|isomorphic" data-lt-noDefault class="export">isomorphic</dfn>
       (that is, they have an identical form)
-      if there is a bijection <var>M</var>
-      from the set of all <a>RDF terms</a> into that same set,
-      such that all of the following properties hold:</p>
-
-    <ul>
-      <li><var>M</var> maps blank nodes to blank nodes.</li>
-      <li><var>M</var>(<var>lit</var>)=<var>lit</var> for every <a>RDF literal</a> <var>lit</var>.</li>
-
-      <li><var>M</var>(<var>iri</var>)=<var>iri</var> for every <a>IRI</a> <var>iri</var>.</li>
-
-      <li><var>M</var>(<var>tt</var>) is the triple term ( <var>M</var>(<var>s</var>), <var>M</var>(<var>p</var>), <var>M</var>(<var>o</var>) ) if <var>tt</var> is a triple term of the form ( <var>s</var>, <var>p</var>, <var>o</var> ).</li>
-
-      <li>The triple ( <var>s</var>, <var>p</var>, <var>o</var> ) is in <var>G</var> if and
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1053,17 +1052,11 @@ <h3>RDF Dataset Comparison</h3>
       graph <var>NG1</var> and the RDF dataset <var>D2</var> with default graph
       <var>DG2</var> and any named graph <var>NG2</var>)
       are <dfn data-lt="dataset isomorphism" class="export">dataset-isomorphic</dfn> if and only if
-      there is a bijection <var>M</var> between the nodes, triples and graphs in
-      <var>D1</var> and those in <var>D2</var> such that all of the following properties hold:</p>
+      there is a bijection <var>M</var> from [=blank nodes=] to [=blank nodes=],
```
gkellogg, line 1055: NG1 can be a blank node.

gkellogg (changes requested):

spec/index.html
```diff
@@ -899,26 +899,25 @@ <h3>Triple Terms</h3>
...
+      <li>|M|(|iri|)=|iri| if |iri| is an [=IRI=].
+      <li>|M|(|lit|)=|lit| if |lit| is a [=literal=].
+      <li>|M|(|t|)=( |M|(|s|), |M|(|p|), |M|(|o|) ) if |t| is an [=RDF triple=] of the form ( |s|, |p|, |o| ).
+      <li>|M|(|G|)={ |M|(|t|<sub>1</sub>), ..., |M|(|t|<sub>|n|</sub>) } if |G| = { |t|<sub>1</sub>, ..., |t|<sub>|n|</sub> } is an [=RDF graph=].
+    </ul>
+
     <p id="section-graph-equality">Two
       <a>RDF graphs</a> <var>G</var> and <var>G'</var> are
       <dfn data-lt="graph isomorphism|isomorphic" data-lt-noDefault class="export">isomorphic</dfn>
       (that is, they have an identical form)
```
afs, line 918:
```suggestion
      (that is, they have equivalent form)
```
"identical" would mean the blank nodes are the elements from the set of all blank nodes. "identical meaning" might also work.

spec/index.html
```diff
@@ -1053,17 +1052,11 @@ <h3>RDF Dataset Comparison</h3>
...
     <ul>
-      <li><var>M</var> maps <a>blank nodes</a> to blank nodes;</li>
-      <li><var>M</var> is the identity map on <a>literals</a> and URIs;</li>
-      <li>For every triple &lt;s p o&gt;, <var>M</var>(&lt;s, p, o&gt;)=
-        &lt;<var>M(s)</var>, <var>M(p)</var>, <var>M(o)</var>&gt;;</li>
-      <li>For every graph <var>G</var>={t1, ..., tn},
-        <var>M(G)</var>={<var>M(t1)</var>, ..., <var>M(tn)</var>};</li>
-      <li><var>DG2</var> = <var>M(DG1)</var>; and</li>
+      <li><var>M(DG1)</var> = <var>DG2</var>; and</li>
       <li>&lt;n, G&gt; is in <var>NG1</var> if and only if
```
afs, line 1060: It might be better to switch away from `&lt; .. &gt;` because it is now used around `s,p,o`.

Reusing () - `(A, B)` - or `[ ]`

spec/index.html
pchampin: @gkellogg I don't understand your comment.

That being said, I reused the language about NG1 and NG2 as is, but I find it quite confusing. I'll change it into something that I find more understandable. Hopefully it will solve your comment above as well :)

spec/index.html
pchampin: I agree that "identical" may sound too strong. I'm not sure about "equivalent", because this term has a precise meaning in RDF-SEMANTICS, which is not 1-1 with "isomorphic" (even though isomorphic implies equivalent).

I will go for "the same form".

hartig:
While I agree with the general direction and the need to unify the definitions, I don't find this proposal particularly clean from a formal perspective; especially the part about extending _M_. The other major issue (which you have just [mentioned yourself](https://github.com/w3c/rdf-concepts/pull/159#discussion_r1955781539) and which was there already before this PR) is the one about NG1 and NG2. Give me a few minutes to make another proposal.

spec/index.html
gkellogg: My reading of this is that it covers isomorphism of terms within the default and named graphs, but not on the graph names, themselves. The graph names also need to be part of _M_.

pchampin:
Turning this PR into a draft, as #160 is likely to replace it anyway. Should be closed if/when #160 gets merged.

spec/index.html
pchampin: The previous wording was indeed ambiguous, as it said "bijection M between nodes, triples and graphs", sounding like graph names were excluded. Actually, there were *not* excluded, as later it said:

> <n, G> is in NG1 if and only if <M(n), M(G)> is in NG2. 

so graph named are also mapped via M.

And the proposed wording in this PR (as well as Olaf's counterproposal in #160) is described as mapping from *terms* to *terms*, regardless on where those terms are used. So hopefully, this should be even clearer that graph names are also mapped.

hartig:
@pchampin I think this PR can be closed now (without merging) as it had been replaced by #160, which is merged now.

afs:
Labelled "propose closing" because 160 replaces this PR.

pchampin:
closing in favor of #160

PR #161, opened 2025-02-25
Adds RDF term equality definitions

### LINKED ISSUE #154: RDF term equality definitions
There are definitions for [IRI equality](https://www.w3.org/TR/rdf12-concepts/#dfn-iri-equality) and [literal term equality](https://www.w3.org/TR/rdf12-concepts/#dfn-literal-term-equality). 

There aren't definitions for "blank node equality", "triple term equality" or a general "RDF term equality".

Would it be useful to add these?

SPARQL needs definitions for term equality in [`sameTerm`](https://www.w3.org/TR/sparql12-query/#func-sameTerm) and other places. 

SPARQL 1.1 has it's own definition, and while this could updated to referring to RDF Concepts for IRI and literals and otherwise keeping the current text, if RDF Concepts has "RDF term equality", SPARQL 1.2 ought to use that.

hartig:
Closes #154 by adding definitions of 'blank node equality', 'triple term equality', 'RDF term equality', and 'triple equality'. Additionally, this PR makes the definitions of graph comparison and dataset comparison more explicit by using these notions of equality.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/161.html" title="Last updated on Mar 5, 2025, 6:06 PM UTC (fdcaaf9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/161/b8b739d...fdcaaf9.html" title="Last updated on Mar 5, 2025, 6:06 PM UTC (fdcaaf9)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1092,7 +1092,8 @@
         the following are true:
         <ul>
           <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
-          <li>The triple (|s|, |p|, |o|) is in |G| if and only if
+          <li>For every triple |t|=(|s|, |p|, |o|), it holds that
+            |t| is in |G| if and only if
             the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
         </ul>
       </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -579,7 +579,7 @@
 
     <p><dfn>Triple equality</dfn>:
       Two triples (|s|, |p|, |o|) and (<var>s'</var>, <var>p'</var>, <var>o'</var>)
-      are considered equal if and only if all of the following three conditions hold.</p>
+      are equal (the same [=RDF triple=]) if and only if all of the following three conditions hold.</p>
 
     <ul>
       <li>|s| and <var>s'</var> are [=RDF term equality|equal=].</li>
@@ -610,7 +610,7 @@
       <code>http://example.org/</code>.</p>
 
     <p><dfn>RDF term equality</dfn>:
-      Two [=RDF terms=] |t| and <var>t'</var> are considered equal if and only if
+      Two [=RDF terms=] |t| and <var>t'</var> are equal (the same [=RDF term=]) if and only if
       one of the following four conditions holds:</p>
 
     <ul>
@@ -870,7 +870,7 @@
       any internal structure of blank nodes.</p>
 
     <p><dfn>Blank node equality</dfn>:
-      Two blank nodes are considered equal if and only if they are the same blank node.</p>
+      Two blank nodes are equal if and only if they are the same blank node.</p>
 
     <div class="note" id="note-bnode-id">
       <p><span id="dfn-blank-node-identifiers"><!-- obsolete term--></span><dfn data-lt="blank node identifier">Blank node identifiers</dfn>
--- a/spec/index.html
+++ b/spec/index.html
@@ -614,10 +614,10 @@
       one of the following four conditions holds:</p>
 
     <ul>
-      <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal=].</li>
-      <li>|t| and <var>t'</var> are [=literals=] that are [=literal term equality|equal=].</li>
-      <li>|t| and <var>t'</var> are [=blank nodes=] that are [=blank node equality|equal=].</li>
-      <li>|t| and <var>t'</var> are [=triple terms=] that are [=triple term equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal=] (per [=IRI equality=]).</li>
+      <li>|t| and <var>t'</var> are [=literals=] that are [=literal term equality|equal=] (per [=literal term equality=]).</li>
+      <li>|t| and <var>t'</var> are [=blank nodes=] that are [=blank node equality|equal=] (per [=blank node equality=]).</li>
+      <li>|t| and <var>t'</var> are [=triple terms=] that are [=triple equality|equal=] (per [=triple equality=]).</li>
     </ul>
 
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
@@ -913,9 +913,9 @@
 
     <p>By extension, an [=RDF term=] is said to [=appear=] in an [=RDF graph=] if it appears in an [=asserted triple=] of that graph. An [=RDF triple=] is said to [=appear=] in an [=RDF graph=] if it is either an [=asserted triple=] of that graph or a [=triple term=] [=appearing=] in that graph.</p>
 
-    <p><dfn>Triple term equality</dfn>:
+    <p>Triple term equality:
       Since triple terms are [=triples=], equality of triple terms is the same as [=triple equality=].</p>
-
+      
     <p class="note">Every <a>triple</a> with a <a>triple term</a> as its [=object=] SHOULD
       use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
       as its <a>predicate</a>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -1092,7 +1092,7 @@
         the following are true:
         <ul>
           <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
-          <li>The triple (|s|, |p|, |o|) is in |G| and
+          <li>The triple (|s|, |p|, |o|) is in |G| if and only if
             the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
         </ul>
       </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -579,7 +579,7 @@
 
     <p><dfn>Triple equality</dfn>:
       Two triples (|s|, |p|, |o|) and (<var>s'</var>, <var>p'</var>, <var>o'</var>)
-      are considered equal if and only if all of the following three conditions holds.</p>
+      are considered equal if and only if all of the following three conditions hold.</p>
 
     <ul>
       <li>|s| and <var>s'</var> are [=RDF term equality|equal=].</li>
@@ -901,7 +901,7 @@
     <h3>Triple Terms</h3>
 
     <p>An [=RDF triple=] used as the [=object=] of another [=triple=] is called a <dfn class=export >triple term</dfn>.
-      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or both.
+      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
     </p>
 
     <p>The set of [=RDF terms=] <dfn class=export data-lt="appear">appearing</dfn> in an [=RDF triple=] |t| is defined inductively as follows:</p>
@@ -1089,9 +1089,10 @@
         the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in |DG2|.</li>
       <li>The [=named graph=] (|n|, |G|) is in |NG1| if and only if
         there is a [=named graph=] (<var>n'</var>, <var>G'</var>) in |NG2| such that
+        the following are true:
         <ul>
-          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var> and</li>
-          <li>the triple (|s|, |p|, |o|) is in |G| if and only if
+          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
+          <li>The triple (|s|, |p|, |o|) is in |G| and
             the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
         </ul>
       </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -577,6 +577,16 @@
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
+    <p><dfn>Triple equality</dfn>:
+      Two triples (|s|, |p|, |o|) and (<var>s'</var>, <var>p'</var>, <var>o'</var>)
+      are considered equal if and only if all of the following three conditions holds.</p>
+
+    <ul>
+      <li>|s| and <var>s'</var> are [=RDF term equality|equal=].</li>
+      <li>|p| and <var>p'</var> are [=RDF term equality|equal=].</li>
+      <li>|o| and <var>o'</var> are [=RDF term equality|equal=].</li>
+    </ul>
+
 
     <p class="note">The definition of <a>triple</a> is recursive.
       That is, a <a>triple</a> can itself have an
@@ -598,6 +608,17 @@
       is not equal to the IRI <code>http://example.org/</code>,
       nor to a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
+
+    <p><dfn>RDF term equality</dfn>:
+      Two [=RDF terms=] |t| and <var>t'</var> are considered equal if and only if
+      one of the following four conditions holds:</p>
+
+    <ul>
+      <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=literals=] that are [=literal term equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=blank nodes=] that are [=blank node equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=triple terms=] that are [=triple term equality|equal=].</li>
+    </ul>
 
     <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
       is the set of <a>subjects</a> and <a>objects</a> of the <a>asserted triples</a> of the graph.
@@ -848,6 +869,9 @@
       the set of possible blank nodes is arbitrary.  RDF makes no reference to
       any internal structure of blank nodes.</p>
 
+    <p><dfn>Blank node equality</dfn>:
+      Two blank nodes are considered equal if and only if they are the same blank node.</p>
+
     <div class="note" id="note-bnode-id">
       <p><span id="dfn-blank-node-identifiers"><!-- obsolete term--></span><dfn data-lt="blank node identifier">Blank node identifiers</dfn>
       are local identifiers that are used in some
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -925,8 +952,6 @@ <h3>Graph Comparison</h3>
       the triple (|s|, |p|, |o|) is in |G| if and only if
       the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</p>
 
-    <p>See also: <a>IRI equality</a>, <a>literal term equality</a>.</p>
```
hartig, line 928: Note that I have removed this one because the cross-references to these definitions are integrated directly into the definition above now.

spec/index.html
```diff
@@ -599,6 +609,17 @@ <h3>RDF Terms</h3>
...
 
+    <p><dfn>RDF term equality</dfn>:
+      Two [=RDF terms=] |t| and <var>t'</var> are considered equal if and only if
+      one of the following four conditions holds:</p>
+
+    <ul>
+      <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=literals=] that are [=literal term equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=blank nodes=] that are [=blank node equality|equal=].</li>
+      <li>|t| and <var>t'</var> are [=triple terms=] that are [=triple term equality|equal=].</li>
```
pchampin, line 620:
```suggestion
      <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal=] (per [=IRI equality=]).</li>
      <li>|t| and <var>t'</var> are [=literals=] that are [=literal term equality|equal=] (per [=literal term equality=]).</li>
      <li>|t| and <var>t'</var> are [=blank nodes=] that are [=blank node equality|equal=] (per [=blank node equality=]).</li>
      <li>|t| and <var>t'</var> are [=triple terms=] that are [=triple equality|equal=] (per [=triple equality=]).</li>
```

I would not fight for this change, but I think it makes the rendered text more readable. It can easily be overlooked that the four occurrences of "equal" do not refer to the same definition.

Also, I think it is better to refer directly to "triple equality" than to "triple term equality". See my remark below.

spec/index.html
```diff
@@ -889,6 +913,9 @@ <h3>Triple Terms</h3>
 
     <p>By extension, an [=RDF term=] is said to [=appear=] in an [=RDF graph=] if it appears in an [=asserted triple=] of that graph. An [=RDF triple=] is said to [=appear=] in an [=RDF graph=] if it is either an [=asserted triple=] of that graph or a [=triple term=] [=appearing=] in that graph.</p>
 
+    <p><dfn>Triple term equality</dfn>:
+      Since triple terms are [=triples=], equality of triple terms is the same as [=triple equality=].</p>
+
```
pchampin, line 918:
```suggestion
    <p>Triple term equality:
      Since triple terms are [=triples=], equality of triple terms is the same as [=triple equality=].</p>
      
```
For the sake of regularity, I understand that it is desirable to have a paragraph about "Triple term equality". But I think that making it a definition is not required and adds confusion.  As this paragraph explains, triples terms *are* triples, so triple equality applies here.

pchampin (changes requested):
The only change that I really "request" is the one in 'Graph Comparison', because (unless I'm missing something) it introduces an error.

The others are merely an expression of my preferences, but I can live without them.

afs:
There isn't a preview/diff? Is this because the boilerplate was not included in the description?

spec/index.html
afs: Or the pattern:
```
  <li>|t| and <var>t'</var> are [=IRIs=] that are [=IRI equality|equal as IRIs=].</li>
```

afs (approved):
One suggestion, non-blocking.

hartig:
> There isn't a preview/diff? Is this because the boilerplate was not included in the description?

Strange. I have never put anything special in my PRs before. What would this boilerplate be?

spec/index.html
```diff
@@ -913,9 +940,9 @@ <h3>Graph Comparison</h3>
     <ul>
       <li>|M| is bijective.</li>
       <li>For every [=blank node=] |b|, |M|(|b|) is a [=blank node=] (but not necessarily the same as |b|).</li>
-      <li>For every [=literal=] |lit|, |M|(|lit|) = |lit|.</li>
-      <li>For every [=IRI=] |iri|, |M|(|iri|) = |iri|.</li>
-      <li>For every [=triple term=] |tt| of the form (|s|, |p|, |o|), |M|(|tt|) is the triple term ( |M|(|s|), |M|(|p|), |M|(|o|) ).</li>
+      <li>For every [=literal=] |lit|, |M|(|lit|) is a [=literal=] that is [=literal term equality|equal=] to |lit|.</li>
+      <li>For every [=IRI=] |iri|, |M|(|iri|) is an [=IRI=] that is [=IRI equality|equal=] to |iri|.</li>
+      <li>For every [=triple term=] |tt|, |M|(|tt|) is a [=triple term=] that is [=triple term equality|equal=] to |tt|.</li>
```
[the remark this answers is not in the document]
hartig: > I suggest to revert that change, which is actually changing the semantics...

Yes absolutely! Stupid mistake on my side.

hartig:
> It should be added when the PR is created via the github UI.

That's what is strange. I did create the PR via the GitHub UI, exactly as I have done it for earlier PRs. But this time the link to the preview was not added.

Oh, wait, the only thing that I can imagine being the reason is that I edited the PR text a few minutes after having created the PR. Maybe that edit made a concurrently running auto-edit fail?

afs:
Could well be!

spec/index.html
```diff
@@ -577,6 +577,16 @@ <h3>Triples</h3>
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
+    <p><dfn>Triple equality</dfn>:
+      Two triples (|s|, |p|, |o|) and (<var>s'</var>, <var>p'</var>, <var>o'</var>)
+      are considered equal if and only if all of the following three conditions holds.</p>
```
TallTed, line 582:
```suggestion
      are considered equal if and only if all of the following three conditions hold.</p>
```

spec/index.html
```diff
@@ -877,7 +901,7 @@ <h3>Blank Nodes</h3>
     <h3>Triple Terms</h3>
 
     <p>An [=RDF triple=] used as the [=object=] of another [=triple=] is called a <dfn class=export >triple term</dfn>.
-      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
+      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or both.
```
TallTed, line 904: There should either be one `as` that leads into the list (and no `as` within any list item), _or_ have have one `as` within each list item. I previously requested the former (which I still prefer), and that is what is being partially edited in this PR; so now I request the latter, based on the edit already in this PR. Perhaps folks could +1/-1 this change request.
```suggestion
      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], as an [=asserted triple=], or as both.
```

spec/index.html
```diff
@@ -1064,9 +1089,12 @@ <h3>RDF Dataset Comparison</h3>
...
       <li>The [=named graph=] (|n|, |G|) is in |NG1| if and only if
         there is a [=named graph=] (<var>n'</var>, <var>G'</var>) in |NG2| such that
-        |M|(|n|) = <var>n'</var> and
-        the triple (|s|, |p|, |o|) is in |G| if and only if
-        the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
+        <ul>
+          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var> and</li>
+          <li>the triple (|s|, |p|, |o|) is in |G| if and only if
+            the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
+        </ul>
```
TallTed, line 1096:
```suggestion
        the following are true:
        <ul>
          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
          <li>The triple (|s|, |p|, |o|) is in |G| and
            the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
        </ul>
```

TallTed (changes requested):

spec/index.html
hartig: > There should either be one as that leads into the list (and no as within any list item), or have have one as within each list item. I previously requested the former (which I still prefer), ...

Okay. I'm fine with reverting my change here.
```suggestion
      In a given [=RDF graph=], a [=triple=] can appear as a [=triple term=], an [=asserted triple=], or both.
```

spec/index.html
```diff
@@ -913,9 +940,9 @@ <h3>Graph Comparison</h3>
     <ul>
       <li>|M| is bijective.</li>
       <li>For every [=blank node=] |b|, |M|(|b|) is a [=blank node=] (but not necessarily the same as |b|).</li>
-      <li>For every [=literal=] |lit|, |M|(|lit|) = |lit|.</li>
-      <li>For every [=IRI=] |iri|, |M|(|iri|) = |iri|.</li>
-      <li>For every [=triple term=] |tt| of the form (|s|, |p|, |o|), |M|(|tt|) is the triple term ( |M|(|s|), |M|(|p|), |M|(|o|) ).</li>
+      <li>For every [=literal=] |lit|, |M|(|lit|) is a [=literal=] that is [=literal term equality|equal=] to |lit|.</li>
+      <li>For every [=IRI=] |iri|, |M|(|iri|) is an [=IRI=] that is [=IRI equality|equal=] to |iri|.</li>
```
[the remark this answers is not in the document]
pchampin: > In this case, it would actually be a better idea to replace the = by "is" (e.g., "M(iri) is iri" instead of "M(iri) = iri").

Yes, I like that.

> Notice, however, that this wording makes a difference for literals: 

You gave me a lot to think about with this puzzle :) My conclusion (which I will explain in more detail in the main conversation of this PR) is that this is not (or should not be) an issue.

pchampin:
This PR got me thinking (and re-reading the specs) for a long time... but I think I finally put my finger on what bothers me.

My major issue is that it approaches "term equality" (for each category of terms) as if the abstract syntax had a notion of equality that was different from the notion of identity. And in fact, it *does not*. And after my long mulling, I'm very much convinced that it *must not* make such a distinction (see pathological example below).

Those "term equality" sections exist mostly to emphasize how variations in *concrete syntaxes and internal representations* are not relevant for the abstract syntax. But granted, the pre-existing sections (esp. the one about literals) were already not very clear about that. (I just noticed how the last sentence of the definition of language tag is baroque in that respect: "Two language tags are **the same** if they only **differ** by case." -- and I may very well be responsible for this wording...).

The sections added in this PR are going further in the direction of "distinct but equal", with wording such as "are considered equal".

I'll try to suggest changes to this PR to clarify all this, or possibly make a counter-proposal.

--------

Pathological example showing that the distinction between identity and equality in the abstract syntax is detrimental:

Consider the following two graphs
```
# G1
:s :p1 "chat"@en-US.
:s :p1 "chat"@en-US.
```
```
# G2
:s :p1 "chat"@en-us.
:s :p2 "chat"@EN-US.
```
I would like to consider that they are isomorphic, right? Because all the objects are equal. But if we consider the literals to be *distinct* in the abstract syntax, then per our definition of isomorphism, they are not isomorphic. We would need a mapping M that maps `"chat"@en-US` sometimes to `"chat"@en-us` and sometimes to `"chat@EN-US`!...

Note also that in RDF 1.1, the two graphs above are not isomorphic, nor are they simply-equivalent (but they are D-equivalent if D contains rdf:langString, and therefore RDF-equivalent).

afs:
Typo? `:p1`, `:p1` in G1 but `:p1`, `:p2` in G2.

spec/index.html
```diff
@@ -577,6 +577,16 @@ <h3>Triples</h3>
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
+    <p><dfn>Triple equality</dfn>:
+      Two triples (|s|, |p|, |o|) and (<var>s'</var>, <var>p'</var>, <var>o'</var>)
+      are considered equal if and only if all of the following three conditions hold.</p>
```
pchampin, line 582:
```suggestion
      are equal (the same [RDF triple]) if and only if all of the following three conditions hold.</p>
```

spec/index.html
```diff
@@ -599,6 +609,17 @@ <h3>RDF Terms</h3>
       nor to a blank node with the <a>blank node identifier</a>
       <code>http://example.org/</code>.</p>
 
+    <p><dfn>RDF term equality</dfn>:
+      Two [=RDF terms=] |t| and <var>t'</var> are considered equal if and only if
```
pchampin, line 613:
```suggestion
      Two [=RDF terms=] |t| and <var>t'</var> are equal (the same [=RDF term=]) if and only if
```

spec/index.html
```diff
@@ -848,6 +869,9 @@ <h3>Blank Nodes</h3>
       the set of possible blank nodes is arbitrary.  RDF makes no reference to
       any internal structure of blank nodes.</p>
 
+    <p><dfn>Blank node equality</dfn>:
+      Two blank nodes are considered equal if and only if they are the same blank node.</p>
```
pchampin, line 873:
```suggestion
      Two blank nodes are equal if and only if they are the same blank node.</p>
```

afs:
Wording (not in this PR) that I found potentially weak:

"the two language tags (if any) compare equal"

what if one has a language and one does not?  "abc" and "abc"@en. Only one language tag.

pchampin:
> Wording (not in this PR) that I found potentially weak:
> 
> "the two language tags (if any) compare equal"
> 
> what if one has a language and one does not? "abc" and "abc"@en. Only one language tag.

I'll prepare a PR to improve the section on literals, in the light of my remarks above. For the rest, I think the suggestions I just made on this PR are enough.

spec/index.html
```diff
@@ -1064,9 +1089,13 @@ <h3>RDF Dataset Comparison</h3>
...
       <li>The [=named graph=] (|n|, |G|) is in |NG1| if and only if
         there is a [=named graph=] (<var>n'</var>, <var>G'</var>) in |NG2| such that
-        |M|(|n|) = <var>n'</var> and
-        the triple (|s|, |p|, |o|) is in |G| if and only if
-        the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
+        the following are true:
+        <ul>
+          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
+          <li>The triple (|s|, |p|, |o|) is in |G| and
+            the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
```
pchampin, line 1096:
```suggestion
          <li>The triple (|s|, |p|, |o|) is in |G| if and only if
            the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
```
I assume that the change of "if and only if" to "and" was not intended.

pchampin (approved):
Thanks a lot. I'm very happy with this PR, now, modulo what I believe to be a typo.

hartig:
@TallTed I applied your edit suggestions. Are you okay with this PR now?

spec/index.html
TallTed: No, I intentionally changed the `if and only if` to `and`. Try reading the conditional from the start with less markup, and leaving out the first, simple bullet item from the `<ul>` —
```
The named graph (n, G) is in NG1 if and only if
   there is a named graph (n', G') in NG2 such that
      the triple (s, p, o) is in G if and only if
         the triple ( M(s), M(p), M(o) ) is in G'.
```
— Vs. my simple `and` —
```
The named graph (n, G) is in NG1 if and only if
   there is a named graph (n', G') in NG2 such that
      the triple (s, p, o) is in G and
         the triple ( M(s), M(p), M(o) ) is in G'.
```

spec/index.html
```diff
@@ -1064,9 +1089,13 @@ <h3>RDF Dataset Comparison</h3>
         the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in |DG2|.</li>
       <li>The [=named graph=] (|n|, |G|) is in |NG1| if and only if
         there is a [=named graph=] (<var>n'</var>, <var>G'</var>) in |NG2| such that
-        |M|(|n|) = <var>n'</var> and
-        the triple (|s|, |p|, |o|) is in |G| if and only if
-        the triple ( |M|(|s|), |M|(|p|), |M|(|o|) ) is in <var>G'</var>.</li>
+        the following are true:
+        <ul>
+          <li>|M|(|n|) is [=RDF term equality|equal=] to <var>n'</var>.</li>
+          <li>The triple (|s|, |p|, |o|) is in |G| if and only if
```
TallTed, line 1095: See https://github.com/w3c/rdf-concepts/pull/161#discussion_r1976135601
```suggestion
          <li>The triple (|s|, |p|, |o|) is in |G| and
```

gkellogg:
> There isn't a preview/diff? Is this because the boilerplate was not included in the description?

The PR Preview workflow gets run with every commit. So, after some commits, there is now a preview/diff.

spec/index.html
afs: Leave as is.

 "if and only if" is a two way implication -- `<=>`. There are many "if and only if" in this document in definition sections.

afs: It is repeating the text from 'graph isomorphism', which I find helpful.

TallTed: > It is repeating the text from 'graph isomorphism', which I find helpful.

I have nothing against such repetition, but again, I find this phrasing problematic.

It does not communicate clearly to me.

Given that this text is repeated from 'graph isomorphism', I think it should be changed in both locations, as it is no less problematic at the other location.

hartig: Changing the _"if and only if"_ to _"and"_ (as proposed by @TallTed ) is not correct because, then, it is not clear what the _s_,  _p_, and  _o_ are that this sub-bullet point is talking about.

Yet, I also see @TallTed's point that having the _"if and only if"_ within the _"such that"_ condition of another _"if and only if"_ is problematic; it may not be incorrect but at least confusing.

To fix this issue, I propose to qualify the inner _"if and only if"_ statement as follows:

> * The named graph (_n_, _G_) is in _NG1_ if and only if there is a named graph (_n'_, _G'_) in _NG2_ such that the following are true:
>   *  ...
>   * For every triple _t_=(_s_, _p_, _o_), it holds that _t_ is in _G_ if and only if the triple ( _M_(_s_), _M_(_p_), _M_(_o_) ) is in _G'_.

Hence, the edit suggestion here would then be:
 ```suggestion
           <li>For every triple |t|=(|s|, |p|, |o|), it holds that
             |t| is in |G| if and only if
```
What do you think?

hartig: Thanks Ted. I have committed my proposed change. Are you okay with this PR now?

TallTed (approved):

PR #162, opened 2025-02-26
improve definition of Literal

pchampin:
This PR was motivated by the problem raised [here](https://github.com/w3c/rdf-concepts/pull/161#issuecomment-2684633342), aiming to fix the definition of "literal term equality".
But ended up in a more involved refactoring of the definition of Literal.

Below is a summary of the changes

- reorganizing the content, putting some parts in separate subsections ("Representation of literals", "Literal value")
- simplifying some parts (lexical value now references 'RDF string', removed some redundancies)
- insisting on the fact that the (upper/lower) case is *not part* of the language tag in the abstract syntax (so "chat"@fr and "chat"@FR are not just equal, they are really the same literal)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/162.html" title="Last updated on Mar 17, 2025, 1:07 PM UTC (aec28ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/162/a6212d7...aec28ec.html" title="Last updated on Mar 17, 2025, 1:07 PM UTC (aec28ec)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -754,125 +754,139 @@
     <p>Literals are used for values such as strings, numbers, and dates.</p>
 
     <p>A <dfn data-local-lt="RDF literal">literal</dfn> in an <a>RDF graph</a> consists of
-      two, three, or four elements, as follow:</p>
+      two, three, or four elements, as follow.</p>
 
     <ol>
-      <li>a <dfn>lexical form</dfn> consisting of a sequence of
-        <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a> [[!UNICODE]]
-        which are <a data-cite="I18N-GLOSSARY#dfn-scalar-value">Unicode scalar values</a>,
-        and therefore do not contain
-        <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">Unicode surrogate code points</a></li>
-      <li>a <dfn>datatype IRI</dfn>, being an <a>IRI</a>
+      <li>A <dfn>lexical form</dfn>, being a [=string=].
+      <li>A <dfn>datatype IRI</dfn>, being an <a>IRI</a>
         identifying a datatype that determines how the lexical form maps
-        to a <a>literal value</a></li>
-      <li>if and only if the <a>datatype IRI</a> is
-        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>, a
+        to a <a>literal value</a>.</li>
+      <li>If and only if the <a>datatype IRI</a> is
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> or
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>, a
         non-empty <dfn>language tag</dfn> as defined by [[!BCP47]]. The
         language tag MUST be well-formed according to
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]],
         and MUST be treated consistently, that is, in a case insensitive manner.
-        Two language tags are the same if they only differ by case.</li>
-      <li>if and only if the <a>datatype IRI</a> is
+        Two [[!BCP47]]-complying strings that differ only by case represent the same [=language tag=].</li>
+      <li>If and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>,
-        a non-empty <a>language tag</a>
-        that MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
-        of [[!BCP47]],
-        and MUST be treated consistently, that is, in a case insensitive manner,
-        and a <dfn>base direction</dfn> that MUST be either `ltr` or `rtl`.</li>
+        a <dfn>base direction</dfn> that MUST be either<ul>
+          <li>`ltr`, indicating that the initial text direction is set to left-to-right, or</li>
+          <li>`rtl`, indicating that the initial text direction is set to right-to-left.</li>
+        </ul></li>
     </ol>
 
     <p>A literal is a <dfn>language-tagged string</dfn> if the third element
       is present and the fourth element is not present.
-      Lexical representations of language tags MAY be case normalized,
-      (for example, by canonicalizing as defined by 
-      <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
-      </p>
-
-    <p>A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
+      A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
       if both the third element and fourth elements are present.
-      The third element, the language tag, is treated identically as in a <a>language-tagged string</a>,
-      and the fourth element, <a>base direction</a>, MUST be either `ltr` or `rtl`, which MUST be in lower case.</p>
-
-    <p>The meanings of the <a>base direction</a> values are:</p>
-    <ul>
-      <li>`ltr`: indicates that the initial text direction is set to left-to-right.</li>
-      <li>`rtl`: indicates that the initial text direction is set to right-to-left.</li>
-    </ul>
-
-    <p>Please note that concrete syntaxes MAY support
-      <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
-      <a>lexical form</a> without any <a>datatype IRI</a>, <a>language tag</a>, or <a>base direction</a>.
-      Simple literals are syntactic sugar for abstract syntax
-      <a>literals</a>
-      with the <a>datatype IRI</a>
-      <code>http://www.w3.org/2001/XMLSchema#string</code>
-      (which is commonly abbreviated as <code>xsd:string</code>).
-      Similarly, most concrete syntaxes represent
-      <a>language-tagged strings</a> and <a>directional language-tagged strings</a> without
-      the <a>datatype IRI</a> because it always equals either
-      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>)
-      or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>), respectively.</p>
-
-    <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is:</p>
-
-    <ul>
-      <li>If the literal is a <a>language-tagged string</a>,
-        then the literal value is a pair consisting of its <a>lexical form</a>
-        and its <a>language tag</a>, in that order.</li>
-      <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
-        a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
-        likewise in that order.</li>
-      <li>If the literal's <a>datatype</a> is handled by an RDF implementation,
-        <ul>
-          <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
-            of the <a>datatype</a>, then the literal value is the result of applying
-            the <a>lexical-to-value mapping</a> of the datatype to the
-            <a>lexical form</a>.</li>
-          <li>otherwise, the literal is <dfn data-lt-no-plural>ill-typed</dfn> and no literal value can be
-             associated with the literal. Such a case produces a semantic
-             inconsistency but is not <em>syntactically</em> ill-formed.
-             Implementations SHOULD accept [=ill-typed=] literals and produce RDF
-             graphs from them. Implementations MAY produce warnings when
-             encountering [=ill-typed=] literals.</li>
-        </ul>
-      </li>
-      <li>If the literal's <a>datatype IRI</a> is <em>not</em>
-        handled by an RDF implementation, then the literal value is
-        not defined by this specification. Implementations SHOULD accept
-        literals with unknown datatype IRIs and produce RDF graphs from them.
-      </li>
-    </ul>
+    </p>
 
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
-      Two literals are term-equal (the same <a>RDF literal</a>)
+      two literals are term-equal (the same <a>RDF term</a>)
       if and only if:</p>
 
     <ul>
-      <li>the two <a>lexical forms</a> compare equal</li>
-      <li>the two <a>datatype IRIs</a> compare equal</li>
-      <li>the two <a>language tags</a> (if any) compare equal</li>
-      <li>the two <a>base directions</a> (if any) compare equal</li>
+      <li>the two <a>lexical forms</a> compare equal,</li>
+      <li>the two <a>datatype IRIs</a> compare equal,</li>
+      <li>the two <a>language tags</a> are either both absent, or both present and compare equal,</li>
+      <li>the two <a>base directions</a> are either both absent, both `ltr`, or both `rtl`.</li>
     </ul>
-    <p>Comparison is performed using 
+    <p>Comparison of the [=lexical forms=] and of the [=datatype IRIs=] is performed using
       <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">case sensitive matching</a>
-      (see description of string comparison in 
-      <a href="#rdf-strings" class="sectionRef"></a>)
-      except for language tags, where the comparison is performed using 
+      (see description of string comparison in
+      <a href="#rdf-strings" class="sectionRef"></a>).
+      Comparison of the [=language tags=] is performed using
       <a data-cite="I18N-GLOSSARY#dfn-case-sensitive">ASCII case-insensitive matching</a>.
-      Thus, two literals can have the same value
-      without being the same <a>RDF term</a>.
-      For example:</p>
+    </p>
 
-    <pre>
-      "1"^^xs:integer
-      "01"^^xs:integer
-    </pre>
+    <section>
+      <h2>Representation of literals</h2>
 
-    <p>denote the same <a data-lt="literal value">value</a>, but are not the
-      same literal <a>RDF terms</a> and are not
-      <a>term-equal</a> because their
-      <a>lexical forms</a> differ.</p>
+      <p>Some concrete syntaxes MAY support
+        <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
+        <a>lexical form</a> without any <a>datatype IRI</a>, <a>language tag</a>, or <a>base direction</a>.
+        Simple literals are syntactic sugar for abstract syntax
+        <a>literals</a>
+        with the <a>datatype IRI</a>
+        <code>http://www.w3.org/2001/XMLSchema#string</code>
+        (which is commonly abbreviated as <code>xsd:string</code>).
+      </p>
+
+      <p>
+        Similarly, most concrete syntaxes represent
+        <a>language-tagged strings</a> and <a>directional language-tagged strings</a> without
+        the <a>datatype IRI</a> because it always equals either
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>)
+        or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>), respectively.
+      </p>
+
+      <p>
+        Any [=string=] complying with [[!BCP47]] MAY be used to represent a [=language tag=] in concrete syntaxes or implementation.
+        Such strings MAY be case normalized,
+        (for example, by canonicalizing as defined by
+        <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
+        On the contrary, an implementation MAY preserve the case from the original representation,
+        provided that it processes it in a case-insensitive manner.
+      </p>
+
+      <aside class=note>
+        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
+        In RDF 1.1, `"chat"@fr` and `"chat"@FR` were representing two distinct terms, but implementations had license to replace one with the other (which most did).
+        In RDF 1.2, they are now representing the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
+      </aside>
+
+    </section>
+
+    <section>
+      <h2>Literal value</h2>
+
+      <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is defined as follow.</p>
+
+      <ul>
+        <li>If the literal is a <a>language-tagged string</a>,
+          then the literal value is a pair consisting of its <a>lexical form</a>
+          and its <a>language tag</a>, in that order.</li>
+        <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
+          a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
+          likewise in that order.</li>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
-      <li>the two <a>datatype IRIs</a> compare equal</li>
-      <li>the two <a>language tags</a> (if any) compare equal</li>
-      <li>the two <a>base directions</a> (if any) compare equal</li>
+      <li>the two <a>lexical forms</a> compare equal,</li>
+      <li>the two <a>datatype IRIs</a> compare equal,</li>
+      <li>the two <a>language tags</a> are either both absent, or both present and compare equal,</li>
+      <li>the two <a>base directions</a> are either both absent, both `ltr`, or both `rtl`.</li>
     </ul>
-    <p>Comparison is performed using 
+    <p>Comparison of the [=lexical forms=] and of the [=datatype IRIs=] is performed using
```
hartig, line 776: For the datatype IRIs, shouldn't this better be covered by [IRI equality](https://w3c.github.io/rdf-concepts/spec/#dfn-iri-equality)?

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+        <a>literals</a>
+        with the <a>datatype IRI</a>
+        <code>http://www.w3.org/2001/XMLSchema#string</code>
+        (which is commonly abbreviated as <code>xsd:string</code>).
+      </p>
+
+      <p>
+        Similarly, most concrete syntaxes represent
+        <a>language-tagged strings</a> and <a>directional language-tagged strings</a> without
+        the <a>datatype IRI</a> because it always equals either
```
hartig, line 800:
```suggestion
        the <a>datatype IRI</a> because it is always either
```

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> (<code>rdf:langString</code>)
+        or <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code> (<code>rdf:dirLangString</code>), respectively.
+      </p>
+
+      <p>
+        Any [=string=] complying with [[!BCP47]] MAY be used to represent a [=language tag=] in concrete syntaxes or implementation.
```
hartig, line 806:
```suggestion
        Any [=string=] complying with [[!BCP47]] MAY be used to represent a [=language tag=] in concrete syntaxes or implementations.
```

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+        Such strings MAY be case normalized,
+        (for example, by canonicalizing as defined by
+        <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
```
hartig, line 809:
```suggestion
        Such strings MAY be case normalized
        (for example, by canonicalizing as defined by
        <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
```

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+        On the contrary, an implementation MAY preserve the case from the original representation,
```
hartig, line 810:
```suggestion
        Alternatively, an implementation MAY preserve the case from the original representation,
```

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+        provided that it processes it in a case-insensitive manner.
+      </p>
+
+      <aside class=note>
+        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
+        In RDF 1.1, `"chat"@fr` and `"chat"@FR` were representing two distinct terms, but implementations had license to replace one with the other (which most did).
+        In RDF 1.2, they are now representing the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
+      </aside>
```
gkellogg, line 818: I think this is too strong. RDF 1.1 was careful to allow implementations to normalize (typically lower-case) the language tag when parsing, so differences in case may not result in different terms and triples could collapse.

Perhaps just say it was a point of ambiguity but 1.2 behavior generally represents what implementations actually do.

spec/index.html
```diff
@@ -733,125 +733,139 @@ <h3>Literals</h3>
...
+
+    </section>
+
+    <section>
+      <h2>Literal value</h2>
+
+      <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is defined as follow.</p>
```
gkellogg, line 825: English (American English, anyway) prefers the use of the plural "follows" here.

```suggestion
        <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is defined as follows.</p>
```

gkellogg (commented):
Minor points.

spec/index.html
pchampin: fair point. I reused existing language, which didn't mention IRI equality.  This is equivalent, because IRI equality is also based on string comparison, but this would be clearer.

spec/index.html
pchampin: comma followed by opening parenthesis, really? @TallTed ?

spec/index.html
pchampin: Proposal:
* replace "were representing two distinct terms" with "were *theoretically* representing two distinct terms"
* explain a little more than "replacing" in RDF 1.1 uses some form of normalization
* add a sentence explicitly stating that most implementations will not be affected, because they normalize already

```suggestion
      <aside class=note>
        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
        In RDF 1.1, `"chat"@fr` and `"chat"@FR` were theoretically representing two distinct terms, but implementations had license to replace one with the other via some form of normalization.
        In RDF 1.2, they are now representing the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
        Since most RDF 1.1 implementation do normalize language tags internally, they will not be impacted by this change?
      </aside>
```

hartig: For the last of the new sentences that you propose: This is a claim. I wouldn't make claims here. So, better: Hence, RDF 1.1 implementations that normalize language tags internally will not be impacted by this change.

If you want to keep your sentence, however, then: "implementation" -> "implementations", and no question mark at the end.

Otherwise, this change looks good to me.

pchampin: Thanks @hartig . I propose to change "most" into "many", which is closer to what we can actually claim. 

But I can back this claim with data. Out of the 10 implementations I have tested (Comunica, Corese, Dydra, GraphDB, Jena, Oxigraph, Python RDFlib, Qlever, Ruby RDF, Virtuoso), only Virtuoso considers that `"chat"@fr` and `"chat"@FR` are not the same term (but, ironically, displays them both as `"chat"@fr` :smile:).

```suggestion
      <aside class=note>
        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
        In RDF 1.1, `"chat"@fr` and `"chat"@FR` were theoretically representing two distinct terms, but implementations had license to replace one with the other via some form of normalization.
        In RDF 1.2, they are now representing the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
        Since many RDF 1.1 implementations do normalize language tags internally, they will not be impacted by this change.
      </aside>
```

hartig: Looks good!

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+        (for example, by canonicalizing as defined by
+        <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
+        Alternatively, an implementation MAY preserve the case from the original representation,
+        provided that it processes it in a case-insensitive manner.
+      </p>
+
+      <aside class=note>
+        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
+        In RDF 1.1, `"chat"@fr` and `"chat"@FR` were theoretically representing two distinct terms, but implementations had license to replace one with the other via some form of normalization.
+        In RDF 1.2, they are now representing the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
```
pchampin, line 817:
```suggestion
        In RDF 1.1, `"chat"@fr` and `"chat"@FR` theoretically represent two distinct terms, but implementations may replace one with the other via some form of normalization.
        In RDF 1.2, they represent the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
```
as suggested by @afs

pfps (changes requested):
A language tag is not a string.  BCP 47 does not provide a good foundation for RDF language tags.

RDF Concepts could say that a language tag is a lowercase string that meets the requirements of BCP 47 or it could say that a language tag is a sequence of ASCII case-insensitive characters where the string constructed by taking any of members of the equivalence sets in sequence meets the requirements of BCP 47.  ASCII case-insensitive characters are then equivalence sets of characters under the equivalence relation that treats two characeters as equivalent if they are both the same when converted to lower case using ASCII case conversion.  The former is simpler but the latter provides guidance on how to treat surface syntax language tags.

Saying that language tags are strings and then going on to define an equality over them is like saying that language tags are cats and then going on to say that two language tags are the same if they have the same colour - the right way here is to say either that language tag strings are cat colours or that they are equivalence classes of cats under the same-colour equivalence.

afs:
In what way does RDF Concepts not say that?

A change might be saying that language tags are _represented_ by strings forming to RFC 5646.

pfps:
> In what way does RDF Concepts not say that?

Not say what?

pfps:
RDF Concepts says this, as far as I can tell:
>  language tags are strings and then [goes] on to define an equality over them 

RDF Concepts does not say either of these, as far as I can tell:
> RDF Concepts could say that a language tag is a lowercase string that meets the requirements of BCP 47 or it could say that a language tag is a sequence of ASCII case-insensitive characters where the string constructed by taking any of members of the equivalence sets in sequence meets the requirements of BCP 47. ASCII case-insensitive characters are then equivalence sets of characters under the equivalence relation that treats two characters as equivalent if they are both the same when converted to lower case using ASCII case conversion. The former is simpler but the latter provides guidance on how to treat surface syntax language tags.

pchampin:
> RDF Concepts says this, as far as I can tell:
> 
> > language tags are strings and then [goes] on to define an equality over them

Not exactly. The text in this PR says

 > a non-empty language tag as defined by [BCP47]. [...] Two [BCP47]-complying strings that differ only by case represent the same language tag.

The goal is to convey the idea that RDF language tags are *an abstraction* of the string complying with BCP-47, without using such scary language. But ok, maybe it's too handwavy.

I would be happy with changing the definition of language tags to lower-case BCP47-compliant strings (as proposed by @pfps). The 3rd paragraph of 3.4.1, in my opinion, explains clearly enough that concrete syntaxes and implementations are free to use the case they want (as long as they ignore it when comparing language tags).

Note that I'm off for 1 week starting 1h ago, so this will not progress unless another editor takes custody of this PR.

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
     <p>Literals are used for values such as strings, numbers, and dates.</p>
 
     <p>A <dfn data-local-lt="RDF literal">literal</dfn> in an <a>RDF graph</a> consists of
-      two, three, or four elements, as follow:</p>
+      two, three, or four elements, as follow.</p>
```
TallTed, line 736: This was intentionally a colon. A full-stop puts a bit too much break.
```suggestion
      two, three, or four elements, as follow:</p>
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
-        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code>, a
+        to a <a>literal value</a>.</li>
+      <li>If and only if the <a>datatype IRI</a> is
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> or
+        <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>, a
         non-empty <dfn>language tag</dfn> as defined by [[!BCP47]]. The
         language tag MUST be well-formed according to
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]],
         and MUST be treated consistently, that is, in a case insensitive manner.
```
TallTed, line 750:
```suggestion
        and MUST be treated consistently in a case insensitive manner.
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+      <li>If and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>,
-        a non-empty <a>language tag</a>
-        that MUST be well-formed according to <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
-        of [[!BCP47]],
-        and MUST be treated consistently, that is, in a case insensitive manner,
-        and a <dfn>base direction</dfn> that MUST be either `ltr` or `rtl`.</li>
+        a <dfn>base direction</dfn> that MUST be either<ul>
+          <li>`ltr`, indicating that the initial text direction is set to left-to-right, or</li>
+          <li>`rtl`, indicating that the initial text direction is set to right-to-left.</li>
```
TallTed, line 756:
```suggestion
        a <dfn>base direction</dfn> that MUST be one of the following:<ul>
          <li>`ltr`, indicating that the initial text direction is set to left-to-right</li>
          <li>`rtl`, indicating that the initial text direction is set to right-to-left</li>
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
-        not defined by this specification. Implementations SHOULD accept
-        literals with unknown datatype IRIs and produce RDF graphs from them.
-      </li>
-    </ul>
+    </p>
 
     <p><dfn data-local-lt="term-equal">Literal term equality</dfn>:
-      Two literals are term-equal (the same <a>RDF literal</a>)
+      two literals are term-equal (the same <a>RDF term</a>)
       if and only if:</p>
```
TallTed, line 768:
```suggestion
      if and only if the following are all true:</p>
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
 
     <ul>
-      <li>the two <a>lexical forms</a> compare equal</li>
-      <li>the two <a>datatype IRIs</a> compare equal</li>
-      <li>the two <a>language tags</a> (if any) compare equal</li>
-      <li>the two <a>base directions</a> (if any) compare equal</li>
+      <li>the two <a>lexical forms</a> compare equal,</li>
+      <li>the two <a>datatype IRIs</a> compare equal,</li>
+      <li>the two <a>language tags</a> are either both absent, or both present and compare equal,</li>
+      <li>the two <a>base directions</a> are either both absent, both `ltr`, or both `rtl`.</li>
```
TallTed, line 774:
```suggestion
      <li>The two <a>lexical forms</a> compare equal.</li>
      <li>The two <a>datatype IRIs</a> compare equal.</li>
      <li>The two <a>language tags</a> are either both absent, or both present and compare equal.</li>
      <li>The two <a>base directions</a> are either both absent, both `ltr`, or both `rtl`.</li>
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+        (for example, by canonicalizing as defined by
+        <a data-cite="bcp47#section-4.5">BCP 47 section 4.5</a>).
+        Alternatively, an implementation MAY preserve the case from the original representation,
+        provided that it processes it in a case-insensitive manner.
+      </p>
+
+      <aside class=note>
+        The treatment of language tags has changed between RDF 1.1 and RDF 1.2.
+        In RDF 1.1, `"chat"@fr` and `"chat"@FR` theoretically represent two distinct terms, but implementations may replace one with the other via some form of normalization.
+        In RDF 1.2, they represent the exact same literal, i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
```
TallTed, line 817:
```suggestion
        In RDF 1.1, `"chat"@fr` and `"chat"@FR` represent two distinct terms,
        but implementations may replace either with the other via some form of normalization.
        In RDF 1.2, they represent the exact same literal,
        i.e., the case difference in the concrete syntax does not propagate into the abstract syntax.
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+        <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is defined as follows.</p>
+
+      <ul>
+        <li>If the literal is a <a>language-tagged string</a>,
+          then the literal value is a pair consisting of its <a>lexical form</a>
+          and its <a>language tag</a>, in that order.</li>
+        <li>If the literal is a <a>directional language-tagged string</a>, then the literal value is
+          a tuple of its <a>lexical form</a>, its <a>language tag</a>, and its <a>base direction</a>,
+          likewise in that order.</li>
+        <li>If the literal's <a>datatype</a> is handled by an RDF implementation,
```
TallTed, line 835:
```suggestion
        <li>If the literal's <a>datatype</a> is handled by an RDF implementation, then one of the following applies:
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+            <li>if the literal's <a>lexical form</a> is in the <a>lexical space</a>
+              of the <a>datatype</a>, then the literal value is the result of applying
+              the <a>lexical-to-value mapping</a> of the datatype to the
+              <a>lexical form</a>.</li>
+            <li>otherwise, the literal is <dfn data-lt-no-plural>ill-typed</dfn> and no literal value can be
+               associated with the literal. Such a case produces a semantic
+               inconsistency but is not <em>syntactically</em> ill-formed.
+               Implementations SHOULD accept [=ill-typed=] literals and produce RDF
+               graphs from them. Implementations MAY produce warnings when
+               encountering [=ill-typed=] literals.</li>
```
TallTed, line 846:
```suggestion
            <li>If the literal's <a>lexical form</a> is in the <a>lexical space</a>
              of the <a>datatype</a>, then the literal value is the result of applying
              the <a>lexical-to-value mapping</a> of the datatype to the
              <a>lexical form</a>.</li>
            <li>Otherwise, the literal is <dfn data-lt-no-plural>ill-typed</dfn> and no literal value can be
               associated with the literal. Such a case produces a semantic
               inconsistency, but it is not <em>syntactically</em> ill-formed.
               Implementations SHOULD accept [=ill-typed=] literals and produce RDF
               graphs from them. Implementations MAY produce warnings when
               encountering [=ill-typed=] literals.</li>
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+        </li>
+        <li>If the literal's <a>datatype IRI</a> is <em>not</em>
+          handled by an RDF implementation, then the literal value is
+          not defined by this specification. Implementations SHOULD accept
+          literals with unknown datatype IRIs and produce RDF graphs from them.
+        </li>
+      </ul>
+
+      <p>
+        Thus, two literals can have the same value
```
TallTed, line 857:
```suggestion
        It follows from the above that two literals can have the same value
```

spec/index.html
```diff
@@ -733,125 +733,140 @@ <h3>Literals</h3>
...
+        without being the same <a>RDF term</a>.
+        For example:</p>
+
+      <pre>
+        "1"^^xsd:integer
+        "01"^^xsd:integer
+      </pre>
+
+      <p>denote the same <a data-lt="literal value">value</a>, but are not the
+        same literal <a>RDF terms</a> because their
```
TallTed, line 867:
```suggestion
        same literal <a>RDF term</a> because their
```

TallTed (changes requested):
Some tweaks for clarity, grammar, and consistency.

gkellogg:
> A language tag is not a string. BCP 47 does not provide a good foundation for RDF language tags.

According to [I18N Glossary](https://www.w3.org/TR/i18n-glossary/), it's a string:

> Language tag. A string used as an identifier for a language, usually referring explicitly to a [[BCP47](https://www.w3.org/TR/i18n-glossary/#bib-bcp47)] language tag. These language tags consist of one or more [language subtags](https://www.w3.org/TR/i18n-glossary/#dfn-language-subtag).

It's just the way that language-tags are compared where case is not significant.

BCP47 describes them using ABNF and discusses maximum character length (for example), so I would say they are strings.

My own implementation turns them into symbols, but that's mostly about uniqueness of symbols vs. strings in Ruby.

> I would be happy with changing the definition of language tags to lower-case BCP47-compliant strings (as proposed by @pfps). The 3rd paragraph of 3.4.1, in my opinion, explains clearly enough that concrete syntaxes and implementations are free to use the case they want (as long as they ignore it when comparing language tags).

I would bo okay with saying that they can be converted to lower-case only for the purpose of comparison. Internally, we should be vague enough for implementations using standard formatting (e.g., `"foo"@en-US`) have a reasonable expectation of getting out what the put in, even though semantically, it is treated equivalently to the lower-case normalized version.

pfps:
> I would be okay with saying that they can be converted to lower-case only for the purpose of comparison. Internally, we should be vague enough for implementations using standard formatting (e.g., `"foo"@en-US`) have a reasonable expectation of getting out what the put in, even though semantically, it is treated equivalently to the lower-case normalized version.

So then Semantics still has to keep the lower-case processing.

pfps (commented):
The proposed wording is (minimally) acceptable.  But this wording still requires Semantics to transform language tags into lower case, or perform some other mechanism for implementing the case-insensitivity aspect of language tags.

pfps (approved):
I'm only approving to override my block - it's not that I particularly like the change, but it can be fixed in Semantics.

spec/index.html
pchampin: Since the bullets of the list have been converted to full sentences (starting with a capital letter, ended with a period), it seemed more grammatically correct to mark the end of a sentence here. But if you confirm that this is ok, I defer to your better knowledge of English grammar.

spec/index.html
pchampin: I think we are losing something with this change. Namely, the idea is that they "MUST be treated consistently [with BCP47]". With the proposed wording, "consistently" seems to only mean "always the same way".

spec/index.html
TallTed: The colon takes the place of the full-stop. It is correct English punctuation.

spec/index.html
TallTed: The existing text says `same literal RDF terms` (i.e., plural), which I'm suggesting be changed to `same literal RDF term` (i.e., singular), which seems to me to conform to what [you're expressing](https://github.com/w3c/rdf-concepts/pull/162#discussion_r1991402727) you prefer?

pchampin: yes, sorry, I misread the diff.

spec/index.html
pchampin: > The that is implies that consistent treatment is equivalent to case insensitive treatment.

Again, since [BCP47] is very shortly before the word "consistently", this equivalence must be understood as "consistent *with BCP47*"  and "case insensitive". At least that's how I understand this sentence (which, I now realize, was not even written by me :).

Would it work better if we replaced "consistently" with "accordingly"?

```suggestion
        and MUST be treated accordingly, that is, in a case insensitive manner.
```

afs (approved):

gkellogg (approved):

TallTed (approved):
Approving, given the change from `consistently` to `accordingly` on line 750.

spec/index.html
pchampin: @hartig are you happy to merge this PR and potentially raise a separate issue for the point above? Otherwise, could you suggest a change? thx

hartig: @pchampin please merge and I will create a separate PR for this issue afterwards.

pchampin:
I just force-pushed a rebased version of this branch on the latest `main`.

Issue #181, opened 2025-04-02
Normative statement in a note


pchampin:
In https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#section-triple-terms , we have a note containing two normative statements (SHOULD, SHOULD NOT)

> Every [triple](https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#dfn-rdf-triple) whose [object](https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#dfn-object) is not a [triple term](https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#dfn-triple-term) SHOULD NOT use http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies (rdf:reifies) as its [predicate](https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#dfn-predicate).

This makes no sense, as notes are non normative.

I see two options:
* turn this note into a normal (normative) paragraph
* keep it as a note, but replace the "SHOULD"s with a plain english (non-normative) "should"s

csarven:
Agree but the solution depends :)

If what's currently expressed is the Note is preferable as a requirement, simply get it out of there and include it part of the normative section.

If what's intended is non-normative or advisory level information, I'd suggest to express the statement differently and also terms along the lines of:

>“strongly encouraged”, “strongly discouraged”, “encouraged", “discouraged", “can", “cannot”, “could”, “could not”, “might”, and “might not”

See also https://solidproject.org/ED/protocol#advisement-levels and https://infra.spec.whatwg.org/#conformance using the above terms.

For clarity, I would also suggest to refrain from using requirement level terms as mentioned in https://www.w3.org/TR/2025/WD-rdf12-concepts-20250327/#conformance even in lowercase. So, limit the use of requirement level terms for normative statements, and use advisement level terms for non-normative.

gkellogg:
IIRC, this language was added at a point where we were still deciding how firm we wanted to be about restricting the use of triple terms. At this time, I think the advice is non-normative, and we should use language that is just suggestive, as @csarven suggests.

gkellogg:
Fixed elsewhere.

PR #182, opened 2025-04-02
Declare advisement level keywords for non-normative content in Conformance

csarven:
This is a [correction class 2 change](https://www.w3.org/policies/process/#class-2) updating the #conformance section to clarify the use of #advisement-levels terms.

This proposal is based on the suggestion in https://github.com/w3c/rdf-concepts/issues/181#issuecomment-2772191395 to better distinguish the terminology used in normative and non-normative content. That is, continue to reserve RFC keywords in uppercase for normative content, and only use advisement level terms in lowercase for non-normative content.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -503,6 +503,10 @@
 </section>
 
 <section id="conformance">
+  <p id="advisement-levels">The key words “strongly encouraged”, “strongly
+  discouraged”, “encouraged", “discouraged", “can", “cannot”, “could”, “could
+  not”, “might”, and “might not” are usedfor non-normative content.</p>
+
   <p>This specification, <em>RDF 1.2 Concepts and Abstract Syntax</em>,
     defines a data model and related terminology for use in
     other specifications, such as


pfps:
The point of using UPPER CASE and calling out wording as normative is to distinguish between normative and non-normative wording.  This seems to me to be adequate for the purpose.

Adding a paragraphp on what is non-normative does not seem to me to be helpful.  On the contrary, I view it as diluting the message about normative wording.

Unless there is direction from W3C I think that this change should not be applied.

csarven:
The addition does not dilute the existing terms for normative content in any way. In fact, the proposal is about clarifying the language used for non-normative content. This approach promotes uniformity and minimizes potential misunderstandings.

Moreover, screen readers do not inherently distinguish between uppercase and lowercase words when reading text aloud. For example, "SHOULD" and "should" are typically read the same way. This raises accessibility concerns, making the distinction between uppercase SHOULD for normative content and lowercase should for non-normative content less straightforward.

It is more effective to encourage the use of distinct terms for normative and non-normative content—both for authors and editors, in terms of how content is articulated, and for readers, in terms of how content is interpreted while minimizing oversight.

(Sorry, I don't have hard data, surveys, or studies to back this up if necessary, but I see it as an obvious point. Additionally, as mentioned in the comment, some of these terms are already used in existing reports.)

spec/index.html
```diff
@@ -503,6 +503,10 @@ <h3>RDF Documents and Syntaxes</h3>
 </section>
 
 <section id="conformance">
+  <p id="advisement-levels">The key words “strongly encouraged”, “strongly
+  discouraged”, “encouraged", “discouraged", “can", “cannot”, “could”, “could
```
hartig, line 507: Nitpick: Let's be consistent about the quotation marks used ;-)
```suggestion
  discouraged”, “encouraged”, “discouraged”, “can”, “cannot”, “could”, “could
```

spec/index.html
```diff
@@ -503,6 +503,10 @@ <h3>RDF Documents and Syntaxes</h3>
...
+  not”, “might”, and “might not” are usedfor non-normative content.</p>
```
hartig, line 508:
```suggestion
  not”, “might”, and “might not” are used for non-normative content.</p>
```

hartig (changes requested):
Just two typos to be fixed.

Other than that, I don't see any harm in adding this paragraph. (The real task will then be do go over the doc and replace the non-normative uses of words like "may", "must", etc.

afs:
> Other than that, I don't see any harm in adding this paragraph. (The real task will then be do go over the doc and replace the non-normative uses of words like "may", "must", etc.

Yes, the impact needs to be assessed, noting that sometimes "should" may be the right wording.
The suggested words are sometimes opinionated in ways that current text isn't. "could not” reads as implying a logical impossibility.

Presumably the next request will be to do this across all RDF documents.

afs:
We have significant resource constraints.

> re going through the rest of this document as well as the other RDF documents to ensure consistency, I fully agree (I forgot to suggest that in my initial comment but glad it is caught and can be followed up).

I can not support this PR until there is a roadmap of the work.

pfps:
> Peter, I'm not quite following your argument. Can you point to specific lines or directly quote the proposed addition that conflicts with anything?
> 
> Again, this proposal does not modify normative wording - it only clarifies the language for non-normative content. Since the normative language remains unchanged, the argument about requiring a broader W3C directive doesn't apply. As far as I know, there's no directive mandating a uniform Conformance section across all W3C reports. There's no need to update all W3C documents simultaneously. What matters is that each specification's Conformance section is meaningful, useful, and internally consistent. This proposal aims to improve that, specifically for non-normative content.
> 
> You're also overlooking the accessibility issue, which is a concrete concern. Writing "SHOULD" and "should" in different cases is not a good practice no matter how you slice it. If anything, even the existence of RFC 8174 argues for making that distinction.

Whether SHOULD vs should is a good way of signalling normative content is a separate matter.  You should take that up with W3C and IETF.

But anything adjacent to the referencing of BCP 14 on the same subject has the possibility (and often the probability) of diluting the message.  The wording from BCP 14 on when the meaning from it is to be read into a phrase is very explicit and does not need clarification.

csarven:
This PR doesn't modify BCP 14 or its applicability. It is entirely separate. RFC 8174 clarifies that lowercase terms are not for normative content and leaves them for normal English usage. This PR focuses on clarifying non-normative language and doesn't restrict the use of lowercase terms. Rejecting it just because it deals with conformance language overlooks the proposed improvement without showing any real issue.

If there is anything further I can clarify, happy to do so. Am content to leave it up to the group to evaluate it and go from there. If there is rough consensus to (eventually) introduce this into the document along with further clarifications on the non-normative content, I can follow up on that.

csarven:
I find this part of the meeting quite baffling:

>It is really asking for a change in how BCP 14 is signalled in W3C documents.

This is despite having asked multiple times in this PR for a direct quote, either from the proposed text or from this conversation, that actually supports this claim. Yet I still haven't seen a single shred of evidence.

I do appreciate that some of you have acknowledged the suggestion as useful, with some understandable concerns about its practical implications. That's quite fair but I would ask than what constitutes a legitimate, or any, change from earlier work?

That said, I'd just like to clarify that the suggestion here was specifically and only for rdf-concepts, not to all of this group's documents. If the group later chooses to adopt it in other documents as well, that's great, and that bridge could be crossed later. 

Finally, as I've mentioned in the referenced issue, FWIW, https://infra.spec.whatwg.org/ is one of the specifications that uses the terminology that's proposed in this PR, and it is also the same specifications that's referenced from rdf-concepts - so it is not a completely random or strange thing that's out there using these terms, and I'd presume that the contributors to INFRA have also thought this through, i.e., the interplay between normative (BCP 14) and non-normative content!

... okie dokie.

pfps:
> I find this part of the meeting quite baffling:
> 
> > It is really asking for a change in how BCP 14 is signalled in W3C documents.
> 
> This is despite having asked multiple times in this PR for a direct quote, either from the proposed text or from this conversation, that actually supports this claim. Yet I still haven't seen a single shred of evidence.

You are asking for a change in the section of the document that describes how the keywords in BCP 14 are to interpreted.  This is precisely a change in how BCP 14 is to be signalled in this W3C document. 

> 
> I do appreciate that some of you have acknowledged the suggestion as useful, with some understandable concerns about its practical implications. That's quite fair but I would ask than what constitutes a legitimate, or any, change from earlier work?
> 
> That said, I'd just like to clarify that the suggestion here was specifically and only for rdf-concepts, not to all of this group's documents. If the group later chooses to adopt it in other documents as well, that's great, and that bridge could be crossed later.

If this document has different wording concerning BCP 14 then it would be different from the other WG documents, which is not something that should be done.  So in effect you are asking for a change in all the WG documents.  But then the WG documents would be different from other W3C documents, which is not something that should be done.

> Finally, as I've mentioned in the referenced issue, FWIW, https://infra.spec.whatwg.org/ is one of the specifications that uses the terminology that's proposed in this PR, and it is also the same specifications that's referenced from rdf-concepts - so it is not a completely random or strange thing that's out there using these terms, and I'd presume that the contributors to INFRA have also thought this through, i.e., the interplay between normative (BCP 14) and non-normative content!

That spec says "This is a [willful violation](https://infra.spec.whatwg.org/#willful-violation) of RFC 8174, motivated by legibility and a desire to preserve long-standing practice in many non-IETF-published pre-RFC 8174 documents. [[RFC8174]](https://infra.spec.whatwg.org/#biblio-rfc8174)".  I see this as a bad thing.

Now it may be that BCP 14 or how to use BCP 14 should be updated because the use of case is a bad idea.   Feel free to make this argument to IETF or W3C.

csarven:
Can you please explain what of the proposed text:

>The key words “strongly encouraged”, “strongly discouraged”, “encouraged”, “discouraged”, “can”, “cannot”, “could”, “could not”, “might”, and “might not” are used for non-normative content.

have anything to do with the current text in the Conformance section:

>The key words MAY, MUST, MUST NOT, RECOMMENDED, SHOULD, and SHOULD NOT in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/info/bcp14) [[RFC2119](https://www.w3.org/TR/rdf12-concepts/#bib-rfc2119)] [[RFC8174](https://www.w3.org/TR/rdf12-concepts/#bib-rfc8174)] when, and only when, they appear in all capitals, as shown here.

Are any of the terms in proposed text also in the current text, and if so, which ones?

What is your interpretation of the proposed text that makes you conclude that it changes the interpretation of the current text?

The current text is for normative content and the latter is for non-normative content, and they use different key words. There is nothing in common.

csarven:
The claim of introducing a "three-way decision" is based on a misunderstanding.

RFC 8174:

>When these words are not capitalized, they have their normal
      English meanings and are not affected by this document.

The proposal doesn't restrict non-normative language but brings clarity and consistency.

pchampin:
No consensus in the group that these changes are a good idea. The group may decide to reconsider in the future.

PR #187, opened 2025-04-15
Add version announcement section

rubensworks:
As discussed in https://github.com/w3c/rdf-star-wg/issues/141, this adds a short section to the introduction about version announcement.

I tried to limit the text as much as possible to the common elements across all relevant specs, so that format-specific details can build on top of this.

This could also be a dedicated section outside of the introduction, if that would be preferred.

I did not add myself as editor, since I have not been involved in this spec so far yet, and probably won't be anymore after this change. But please let me know if I should add myself nevertheless.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/187.html" title="Last updated on May 5, 2025, 12:03 PM UTC (5f517ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/187/5e004e3...5f517ba.html" title="Last updated on May 5, 2025, 12:03 PM UTC (5f517ba)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -515,9 +515,9 @@
       or both.</p>
 
     <p>When the version is indicated both in a media-type parameter and in syntax,
-	    they are expected to be the same.
-	    If they differ,
-	    parsers use the version from the media-type parameter and might emit a warning about the mismatch.</p>
+      they are expected to be the same.
+      If they differ,
+      parsers use the version from the media-type parameter and might emit a warning about the mismatch.</p>
 
     <p>To retain compability that is as broad as possible with older parsers,
       only RDF documents that make use of RDF 1.2-specific functionality
--- a/spec/index.html
+++ b/spec/index.html
@@ -525,7 +525,7 @@
       of RDF 1.2-specific functionality it is discouraged to announce a version).</p>
 
     <p>To announce the version in the HTTP responses using the Content-Type header,
-      the <code>version</code> parameter is expected to be used,
+      the server is expected to use the <code>version</code> parameter,
       as illustrated in the following example response.</p>
 
     <pre class="box http">HTTP/1.1 200 OK
--- a/spec/index.html
+++ b/spec/index.html
@@ -364,43 +364,6 @@
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
   </section>
 
-  <section id="section-version-announcement">
-    <h3>RDF Version Announcement</h3>
-
-    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
-      RDF serialization formats are expected to allow a version to be specified,
-      via either the HTTP Content-Type header,
-      a version announcement in a format-specific syntax,
-      or both.</p>
-
-    <p>If the version is announced in both the Content-Type header and the syntax,
-      they are expected to be equal.</p>
-
-    <p>To retain compability that is as broad as possible with older parsers,
-      only RDF documents that make use of RDF 1.2-specific functionality
-      are encouraged to announce their version (i.e., for RDF documents that do not make use
-      of RDF 1.2-specific functionality it is discouraged to announce a version).</p>
-
-    <p>To announce the version in the HTTP responses using the Content-Type header,
-      the <code>version</code> parameter is expected to be used,
-      as illustrated in the following example response.</p>
-
-    <pre class="box http">HTTP/1.1 200 OK
-Content-Type: text/turtle; version=1.2
-Location: http://example.com/document.ttl
-    </pre>
-
-    <p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
-      can be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
-      by including it in the <code>Accept</code> request header,
-      as illustrated in the following example request.</p>
-	
-    <pre class="box http">GET /document.ttl HTTP/1.1
-Host: example.com
-Accept: text/turtle; version=1.2
-    </pre>
-  </section>
-
   <section id="change-over-time">
     <h3>RDF and Change over Time</h3>
 
@@ -540,6 +503,43 @@ Accept: text/turtle; version=1.2
       and different ordering of triples. While these aspects can have great
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
+  </section>
+
+  <section id="section-version-announcement">
+    <h3>RDF Version Announcement</h3>
+
+    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
+      RDF serialization formats are expected to allow a version to be specified,
+      via either the HTTP Content-Type header,
+      a version announcement in a format-specific syntax,
+      or both.</p>
+
+    <p>If the version is announced in both the Content-Type header and the syntax,
+      they are expected to be equal.</p>
+
+    <p>To retain compability that is as broad as possible with older parsers,
+      only RDF documents that make use of RDF 1.2-specific functionality
+      are encouraged to announce their version (i.e., for RDF documents that do not make use
+      of RDF 1.2-specific functionality it is discouraged to announce a version).</p>
+
+    <p>To announce the version in the HTTP responses using the Content-Type header,
+      the <code>version</code> parameter is expected to be used,
+      as illustrated in the following example response.</p>
+
+    <pre class="box http">HTTP/1.1 200 OK
+Content-Type: text/turtle; version=1.2
+Location: http://example.com/document.ttl
+    </pre>
+
+    <p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
+      can be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
+      by including it in the <code>Accept</code> request header,
+      as illustrated in the following example request.</p>
+	
+    <pre class="box http">GET /document.ttl HTTP/1.1
+Host: example.com
+Accept: text/turtle; version=1.2
+    </pre>
   </section>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -365,17 +365,20 @@
   </section>
   
   <section id="section-version-announcement">
-    <h3>RDF version announcement</h3>
+    <h3>RDF Version Announcement</h3>
 	
     <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
-		RDF serialization formats SHOULD allow a version to be specified,
-		either via the HTTP Content-Type header,
+		RDF serialization formats are expected to allow a version to be specified,
+		via either the HTTP Content-Type header,
 		a version announcement in the format-specific syntax,
 		or both.</p>
 
-	<p>If the version is announced in both the Content-Type header and in the syntax, they MUST be equal.</p>
+	<p>If the version is announced in both the Content-Type header and in the syntax, they are expected to be equal.</p>
 	
-	<p>To retain compability that is as broad as possible with older parsers, only RDF documents that make use of RDF 1.2-specific functionality SHOULD announce their version.</p>
+	<p>To retain compability that is as broad as possible with older parsers,
+	  only RDF documents that make use of RDF 1.2-specific functionality
+	  SHOULD announce their version (i.e., RDF documents that do not make use
+	  of RDF 1.2-specific functionality SHOULD NOT announce their version).</p>
 	
 	<p>To announce the version in the HTTP responses using the Content-Type header, the <code>version</code> parameter MUST be used.</p>
 	
@@ -384,8 +387,9 @@ Content-Type: text/turtle; version=1.2
 Location: http://example.com/document.ttl
 	</pre>
 	
-	<p>This <code>version</code> parameter MAY be used in HTTP requests using the Accept header
-		when requesting an RDF document from a server during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]].</p>
+	<p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
+	MAY be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
+	by including it in the <code>Accept</code> request header.</p>
 		
 	<pre class="box http">GET /document.ttl HTTP/1.1
 Host: example.com
@@ -2309,7 +2313,7 @@ Accept: text/turtle; version=1.2
 
   <ul>
 	<li>Added <a href="#section-version-announcement"class="sectionRef"></a>
-		for announcing RDF versions in RDF documents via the HTTP Content-Type header or syntax.</li>
+		for announcing RDF versions in RDF documents via the HTTP Content-Type header and/or syntax.</li>
     <li>Added <a href="#section-classic-full-interop" class="sectionRef"></a>
       for informative mappings between RDF [=Full=] and RDF [=Classic=].</li>
     <li>Added <a href="#section-dataset-quad" class="sectionRef"></a>
--- a/spec/index.html
+++ b/spec/index.html
@@ -363,42 +363,42 @@
 
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
   </section>
-  
+
   <section id="section-version-announcement">
     <h3>RDF Version Announcement</h3>
-	
-    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
-		RDF serialization formats are expected to allow a version to be specified,
-		via either the HTTP Content-Type header,
-		a version announcement in a format-specific syntax,
-		or both.</p>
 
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -363,6 +363,35 @@ <h3>Triple Terms and Reification</h3>
...
+	
+	<p>To announce the version in the HTTP responses using the Content-Type header, the <code>version</code> parameter MUST be used.</p>
+	
+	<pre class="box http">HTTP/1.1 200 OK
+Content-Type: text/turtle; version=1.2
+Location: http://example.com/document.ttl
+	</pre>
+	
+	<p>This <code>version</code> parameter MAY be used in HTTP requests using the Accept header
+		when requesting an RDF document from a server during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]].</p>
```
TallTed, line 388:
```suggestion
	<p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
	MAY be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
	by including it in the <code>Accept</code> request header.</p>
```

spec/index.html
```diff
@@ -2279,6 +2308,8 @@ <h3>Acknowledgments for RDF 1.2</h3>
   <h2>Changes between RDF 1.1 and RDF 1.2</h2>
 
   <ul>
+	<li>Added <a href="#section-version-announcement"class="sectionRef"></a>
+		for announcing RDF versions in RDF documents via the HTTP Content-Type header or syntax.</li>
```
TallTed, line 2312:
```suggestion
		for announcing RDF versions in RDF documents via the HTTP Content-Type header and/or syntax.</li>
```

TallTed (changes requested):

spec/index.html
```diff
@@ -363,6 +363,35 @@ <h3>Triple Terms and Reification</h3>
 
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
   </section>
+  
+  <section id="section-version-announcement">
+    <h3>RDF version announcement</h3>
+	
+    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
+		RDF serialization formats SHOULD allow a version to be specified,
```
gkellogg, line 371: This is an informative section, so we should avoid BCP14 words like SHOULD and MUST, or move them to a normative section.

spec/index.html
```diff
@@ -363,6 +363,35 @@ <h3>Triple Terms and Reification</h3>
...
+		either via the HTTP Content-Type header,
+		a version announcement in the format-specific syntax,
+		or both.</p>
+
+	<p>If the version is announced in both the Content-Type header and in the syntax, they MUST be equal.</p>
```
gkellogg, line 376: This is an informative section, so we should avoid BCP14 words like SHOULD and MUST, or move them to a normative section.

spec/index.html
```diff
@@ -363,6 +363,35 @@ <h3>Triple Terms and Reification</h3>
 
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
   </section>
+  
+  <section id="section-version-announcement">
+    <h3>RDF version announcement</h3>
```
gkellogg, line 368: To be consistent, should be in "Title Case".

```suggestion
    <h3>RDF Version Announcement</h3>
```

gkellogg (changes requested):
As noted, we shouldn't use normative terms such as SHOULD, MUST, or MAY in an informative section.

It may be best to leave conformance wording to the concrete syntaxes, and be more descriptive in Concepts.

spec/index.html
TallTed:
```suggestion
		RDF serialization formats are expected to allow a version to be specified,
```

spec/index.html
TallTed:
```suggestion
	<p>If the version is announced in both the Content-Type header and in the syntax, they are expected to be equal.</p>
```

niklasl:
Since this is about syntax ([IMHO](https://github.com/w3c/rdf-star-wg/issues/141#issuecomment-2794677558), and not about versioning in general), I think it would be best to add this section after [1.9 RDF Documents and Syntaxes](https://www.w3.org/TR/rdf12-concepts/#rdf-documents).

spec/index.html
```diff
@@ -2279,6 +2312,8 @@ <h3>Acknowledgments for RDF 1.2</h3>
   <h2>Changes between RDF 1.1 and RDF 1.2</h2>
 
   <ul>
+	<li>Added <a href="#section-version-announcement"class="sectionRef"></a>
+		for announcing RDF versions in RDF documents via the HTTP Content-Type header and/or syntax.</li>
```
hartig, line 2316: Please use indentation characters that are consistent with the rest of the document.
```suggestion
    <li>Added <a href="#section-version-announcement"class="sectionRef"></a>
      for announcing RDF versions in RDF documents via the HTTP Content-Type header and/or syntax.</li>
```

spec/index.html
```diff
@@ -363,6 +363,39 @@ <h3>Triple Terms and Reification</h3>
...
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
   </section>
+  
+  <section id="section-version-announcement">
+    <h3>RDF Version Announcement</h3>
+	
+    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
+		RDF serialization formats are expected to allow a version to be specified,
+		via either the HTTP Content-Type header,
+		a version announcement in the format-specific syntax,
```
hartig, line 373:
```suggestion
		a version announcement in a format-specific syntax,
```

spec/index.html
```diff
@@ -363,6 +363,39 @@ <h3>Triple Terms and Reification</h3>
...
+		or both.</p>
+
+	<p>If the version is announced in both the Content-Type header and in the syntax, they are expected to be equal.</p>
```
hartig, line 376:
```suggestion
	<p>If the version is announced in both the Content-Type header and the syntax,
	  they are expected to be equal.</p>
```

spec/index.html
```diff
@@ -363,6 +363,39 @@ <h3>Triple Terms and Reification</h3>
...
+	
+	<p>To retain compability that is as broad as possible with older parsers,
+	  only RDF documents that make use of RDF 1.2-specific functionality
+	  are encouraged to announce their version (i.e., RDF documents that do not make use
+	  of RDF 1.2-specific functionality SHOULD NOT announce their version).</p>
```
hartig, line 381:
```suggestion
	  are encouraged to announce their version (i.e., for RDF documents that do not make use
	  of RDF 1.2-specific functionality it is discouraged to announce a version).</p>
```

spec/index.html
```diff
@@ -363,6 +363,39 @@ <h3>Triple Terms and Reification</h3>
...
+	
+	<p>To announce the version in the HTTP responses using the Content-Type header, the <code>version</code> parameter is expected to be used.</p>
```
hartig, line 383:
```suggestion
	<p>To announce the version in the HTTP responses using the Content-Type header,
	  the <code>version</code> parameter is expected to be used,
	  as illustrated in the following example response.</p>
```

spec/index.html
```diff
@@ -363,6 +363,39 @@ <h3>Triple Terms and Reification</h3>
...
+	
+	<pre class="box http">HTTP/1.1 200 OK
+Content-Type: text/turtle; version=1.2
+Location: http://example.com/document.ttl
+	</pre>
+	
+	<p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
+	can be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
+	by including it in the <code>Accept</code> request header.</p>
```
hartig, line 392:
```suggestion
	by including it in the <code>Accept</code> request header,
	as illustrated in the following example request.</p>
```

hartig (changes requested):

hartig:
I second @niklasl's suggestion to move this section after [1.9 RDF Documents and Syntaxes](https://www.w3.org/TR/rdf12-concepts/#rdf-documents).

spec/index.html
csarven, line 376: Implementations should be advised on what to do in the case of a discrepancy between the two values:

```suggestion
	<p>When the version is indicated both in the <code>Content-Type</code> header and in the syntax,
	  they are expected to be the same.
	  If they differ,
	  parsers use the <code>Content-Type</code> version and emit a warning about the mismatch.</p>
```

csarven (changes requested):

spec/index.html
rubensworks: The alternatives would be for parsers to:
1. pick the value from the syntax,
2. ignore, or
3. error.

My feeling is that parsers should this, given that multiple `VERSION` directives could also be defined in Turtle, with different values.

In any case, `1.2` is the only allowed value at the moment, so perhaps this sentence is even not required at this stage.

pchampin (approved):

hartig (approved):

spec/index.html
```diff
@@ -504,6 +504,43 @@ <h3>RDF Documents and Syntaxes</h3>
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
   </section>
+
+  <section id="section-version-announcement">
+    <h3>RDF Version Announcement</h3>
+
+    <p>To allow RDF parsers to error or warn about unsupported RDF versions as early as possible,
+      RDF serialization formats are expected to allow a version to be specified,
+      via either the HTTP Content-Type header,
```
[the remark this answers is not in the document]
csarven: Perhaps a silly question but if not the HTTP headers, what specifications are referring to when you say "media-type parameter"? Or did you mean a higher level library that uses the media-type as input to trigger the desired parser?

spec/index.html
```diff
@@ -504,6 +504,43 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      a version announcement in a format-specific syntax,
```
csarven, line 514: Related to [Pierre-Antoine's comment](https://github.com/w3c/rdf-concepts/pull/187/files#r2046873840) about "media-type parameter" in general:

Wouldn't the inclusion of the version in the syntax be the most reliable and only place that a parser should be concerned with? So, why bother with the version parameter elsewhere? Optimisation? Authoritative?

I stress again that if there are multiple places for a parser to look at as I've mentioned in https://github.com/w3c/rdf-concepts/pull/187#pullrequestreview-2771652770 , the parser should also be advised to pick the most authoritative one among possible options, and probably emit a warning when the version information differ.

spec/index.html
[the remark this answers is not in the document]
pchampin: media-types (and media-type parameters) exist independently of HTTP [[RFC6838](https://www.rfc-editor.org/rfc/rfc6838.html)], and the suggestion here is to use this mechanism to convey version information. I see no reason to restrict it to HTTP.

spec/index.html
pchampin: I sympathize with the concern of having multiple ways to convey the information, but both have their specific advantages:
- the format-specific version indicator is usable in contexts were media-type is not available (e.g. most file systems)
- the media-type parameter can be used during HTTP content-negotiation

spec/index.html
pchampin: This kind of situation exists already on the web: charset can be provided by both HTML and HTTP headers, and I believe that there are known good practices on how to prioritize (although I didn't have time to look them up).

gkellogg: We need to be clear that having parsers emit a warning is not a requirement, but an opportunity. At this point, any use that is not "1.2" can be considered an exception and could generate a warning or an error. The situation would be similar if we did allow a further range of values, and they didn't agree (e.g., media-type parameter is "1.2", but `VERSION` is "1.2-dev").

Regarding the charset discrepancy, I saw a reference to [5.2.2 of HTML401](https://www.w3.org/TR/html401/charset.html#h-5.2.2):

> To sum up, conforming user agents must observe the following priorities when determining a document's character encoding (from highest priority to lowest):

> 1. An HTTP "charset" parameter in a "Content-Type" field.
> 2. A [META](https://www.w3.org/TR/html401/struct/global.html#edef-META) declaration with "http-equiv" set to "Content-Type" and a value set for "charset".
> 3. The [charset](https://www.w3.org/TR/html401/struct/links.html#adef-charset) attribute set on an element that designates an external resource.

This suggests that the version specified in the media-type be used in preference to an inline `VERSION`. It does not say what to do if there is a discrepancy, but I think this can be left up to each parser.

spec/index.html
```diff
@@ -504,6 +504,43 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      or both.</p>
+
+    <p>If the version is announced in both the Content-Type header and the syntax,
+      they are expected to be equal.</p>
```
csarven, line 518:
```suggestion
<p>When the version is indicated both in a media-type parameter and in syntax,
	  they are expected to be the same.
	  If they differ,
	  parsers use the version from the media-type parameter and might emit a warning about the mismatch.</p>
```

spec/index.html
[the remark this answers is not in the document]
csarven: Right but what I meant had to do with in which scenarios besides HTTP the media-type parameter used towards parsing. The version information from the media-type may come from elsewhere, I get that, but I'm asking specifically where. (I'm not disputing the use of media-type parameter instead of Content-Type, which also makes sense to me)

spec/index.html
csarven: Ack. Then, definitely mention priority: https://github.com/w3c/rdf-concepts/pull/187/files#r2047503219

spec/index.html
ektrah: If I upload a Turtle file with a version indicator to to some web server and the server then serves the file with a mismatching version in the media type, isn't it more likely that the web server is wrong?

csarven: Perhaps in a general sense, but that kind of roundtripping interoperability isn't typically checked (AFAIK). What is checked is server-client interoperability, or more broadly, sender-receiver. In this case, the server's representation metadata is authoritative and doesn't reflect the original version (regardless of how it was obtained).

afs (approved):

afs:
Given the WG time https://www.w3.org/2025/04/24-rdf-star-minutes.html#a822
does the "Needs Discussion" label still apply? Or can we remove that, merge this PR and iterate?

See follow-on issue #202 for defining version labels.

spec/index.html
```diff
@@ -504,6 +504,45 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      If they differ,
+      parsers use the version from the media-type parameter and might emit a warning about the mismatch.</p>
+
+    <p>To retain compability that is as broad as possible with older parsers,
+      only RDF documents that make use of RDF 1.2-specific functionality
+      are encouraged to announce their version (i.e., for RDF documents that do not make use
+      of RDF 1.2-specific functionality it is discouraged to announce a version).</p>
+
+    <p>To announce the version in the HTTP responses using the Content-Type header,
+      the <code>version</code> parameter is expected to be used,
```
TallTed, line 528:
```suggestion
      the server is expected to use the <code>version</code> parameter,
```

spec/index.html
```diff
@@ -504,6 +504,45 @@ <h3>RDF Documents and Syntaxes</h3>
...
+
+    <pre class="box http">HTTP/1.1 200 OK
+Content-Type: text/turtle; version=1.2
+Location: http://example.com/document.ttl
+    </pre>
+
+    <p>When requesting an RDF document from an HTTP server, the <code>version</code> parameter
+      can be used during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
+      by including it in the <code>Accept</code> request header,
+      as illustrated in the following example request.</p>
```
TallTed, line 539:
```suggestion
    <p>When requesting an RDF document from an HTTP server, a client 
      can use the <code>version</code> parameter 
      during <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]],
      by including it in the <code>Accept</code> request header,
      as illustrated in the following example request.</p>
```

rubensworks:
This one should be good to go now.

afs:
> Once the Content-Type vs. media-type issues have been reconciled, and we've cleaned up the language of what to do when there are conflicts, I think this is good to go.

@gkellogg - has Content-Type vs. media-type issue been addressed? If not, can it be done separately?

I think we should handle the "conflicts" matter separately (if it needs anything) - with #202 there are multiple labels.

gkellogg:
> > Once the Content-Type vs. media-type issues have been reconciled, and we've cleaned up the language of what to do when there are conflicts, I think this is good to go.
> 
> @gkellogg - has Content-Type vs. media-type issue been addressed? If not, can it be done separately?
> 
> I think we should handle the "conflicts" matter separately (if it needs anything) - with #202 there are multiple labels.

I’ve indicated my approval to merge this PR. I think we have more to discuss on media types and conflicts, but we can do that later. There are still some discussions needed, but this should be merged now.

PR #196, opened 2025-04-23
 Suggestions from January review on §2

pchampin:
@afs' suggestion number 6
https://gist.github.com/afs/f71020019fc33b4ded3e105339c6a655

I generally agree with Andy:
I think that most *syntaxes* will (eventually) support Full RDF 1.2, but it is specific implementations that may impose restrictions.

Conversely, listing those syntaxes under "Full Conformance" may give the wrong impression that *other* syntaxes would be needed by Classic/Basic implementations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/196.html" title="Last updated on Apr 28, 2025, 3:03 PM UTC (1bea0f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/196/d8ff9c8...1bea0f8.html" title="Last updated on Apr 28, 2025, 3:03 PM UTC (1bea0f8)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -525,8 +525,7 @@
     <li><dfn class="no-export lint-ignore" data-lt="full">Full conformance</dfn>
       supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
       with <a>triples</a> that contain <a>triple terms</a>.
-      Concrete syntaxes in which such graphs and datasets can be expressed include
-      [[?RDF12-N-TRIPLES]], [[?RDF12-N-QUADS]], [[?RDF12-TURTLE]], and [[?RDF12-TRIG]].</li>
+      </li>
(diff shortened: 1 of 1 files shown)


pchampin:
> We might just list N-Triples as an RDF full format. I agree that picking “favored” formats is a problem, but N-Triples is foundational.

to restate my comment at the top: I don't believe that the Full/Basic distinction should be thought in terms of *formats*. All implementations (Full ones and Basic ones) will probably use the same formats (N-Triples, Turtle, JSON-LD...), but Basic implementation will use a subset of these formats (probably hinted with the profile= media-type parameter, by the way).

TallTed (approved):

afs (approved):

pfps:
Is this really an enhancement?

gkellogg (approved):

PR #198, opened 2025-04-23
 Suggestions from January review on §3.4

pchampin:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/198.html" title="Last updated on Jul 4, 2025, 1:43 PM UTC (999e9c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/198/6d2d31c...999e9c8.html" title="Last updated on Jul 4, 2025, 1:43 PM UTC (999e9c8)">Diff</a>


gkellogg (approved):

spec/index.html
```diff
@@ -791,7 +791,7 @@ <h3>Literals</h3>
     <p>Literals are used for values such as strings, numbers, and dates.</p>
 
     <p>A <dfn data-local-lt="RDF literal">literal</dfn> in an <a>RDF graph</a> consists of
```
hartig, line 793: Mentioning "an RDF graph" here is incorrect. The notion of a literal should be defined independent of any RDF graph.
```suggestion
    <p>A <dfn data-local-lt="RDF literal">literal</dfn> consists of
```

spec/index.html
```diff
@@ -791,7 +791,7 @@ <h3>Literals</h3>
...
-      two, three, or four elements, as follow:</p>
+      two, three, or four component elements, as follows:</p>
```
hartig, line 794: Just a suggestion ("component elements" sounds a bit weird to me):
```suggestion
      two, three, or four components, as follows:</p>
```

hartig (changes requested):

spec/index.html
pchampin: agreed. I actually found it strange when re-reading it, but I was focusing on the changes suggested during review :)

spec/index.html
TallTed: Small tweak.
```
      two, three, or four components, as follow:</p>
```

TallTed (approved):
Approved with my additional suggested change.

spec/index.html
pchampin: @TallTed the change from "as follow" to "as follows" was suggested by @bobdc, and makes sense to me because

- everywhere else, we use "as follows"
- my understanding, confirmed by [a quick web search](https://duckduckgo.com/?q=%22as+follow%22+%22as+follows%22&t=ffab&ia=web), is that "as follows" is grammatically correct, while "as follow" is not

Any argument in the defense of "as follow"?

afs: "components" is also used as "The three components (s, p, o) of an [RDF triple](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-triple). "component element" is an attempt to distinguish the usage at this point then continue using "element" in the rest of the next. It leaves "element" in the rest of the section alone.

If its one word "component" here the rest of the section needs updating as well. "if the third element is present" etc.

hartig: Thanks for mentioning the rationale for this change @afs ! Yet, I would still prefer to use "components".  "component element" sounds repetitive and clunky.

> If its one word "component" here the rest of the section needs updating as well. "if the third element is present" etc.

There are three places in the whole document where the word "element" is used in the given context:
1. Right after the four bullet points of the [definition of "literal"](https://w3c.github.io/rdf-concepts/spec/#dfn-literal). Here it says: "[..] if the third element is present and the fourth element is not present. [..] if both the third element and fourth elements are present."
    --> I propose to replace these mentions of "element" by the actual names of the elements:  "[..] if the language tag is present and the base direction is not present. [..] if both the language tag and the base direction are present."
2. In the [abstract](https://w3c.github.io/rdf-concepts/spec/#abstract), where it says: "RDF 1.2 also introduces directional language-tagged strings, which contain a base direction element that [..]"
    --> I propose to replace this by: ".. contain a base direction component .."
3. In the [change log](https://w3c.github.io/rdf-concepts/spec/#changes-12) (4-th bullet point), where it says: "Added the base direction element as part of [..]"
    --> Also for this one, I propose to replace it by: ".. the base direction component .."

afs: "component" is OK with those changes.

afs (changes requested):
Requested changes for the component/element discussion.

spec/index.html
TallTed:
```suggestion
      two, three, or four components, as below:</p>
```

hartig: I have pushed a commit (https://github.com/w3c/rdf-concepts/pull/198/commits/bcee7d3d3c922b87f5774fc11459960927850af2) with the changes proposed in [my earlier comment](https://github.com/w3c/rdf-concepts/pull/198#discussion_r2063025942), and also with TallTed's "as below" suggestion.

PR #203, opened 2025-05-05
GH-202: Version Labels

### LINKED ISSUE #202: Version labels for RDF profiles.
Working group discussion: [2025-04-24](https://www.w3.org/2025/04/24-rdf-star-minutes.html#a822)

This change would be added after PR #187 is merged
https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/187.html#section-version-announcement

Define the following labels:

| Label | Syntax | Semantics |
|---------|-----------| ----------- |
| "1.2" | RDF 1.2 syntax | [RDF 1.2 Semantics](https://www.w3.org/TR/rdf12-semantics/) |
| "1.2-basic" | RDF 1.2 syntax without triple terms | [RDF 1.2 Semantics](https://www.w3.org/TR/rdf12-semantics/) |
| "1.1" | RDF 1.1 syntax except for use of a version directive | [RDF 1.1 Semantics](https://www.w3.org/TR/rdf11-semantics/) |

"1.2-basic" can include use of initial text direction.

A label of "1.1" indicates that the data, except for the `VERSION` directive, conforms to RDF 1.1.

The label "1.1" is for RDF 1.2 systems to record that data complies with the RDF 1.1 data model and RDF 1.1 semantics.
Such data can be passed, directly or indirectly, to RDF 1.1 systems. The syntax `VERSION` directive is not be passed on.

afs:
This resolves #202.

The defined labels table is in the normative "conformance" section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/203.html" title="Last updated on May 6, 2025, 9:28 AM UTC (799e74c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/203/9639896...799e74c.html" title="Last updated on May 6, 2025, 9:28 AM UTC (799e74c)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -546,9 +546,8 @@ Host: example.com
 Accept: text/turtle; version=1.2
     </pre>
     <p>
-      Defined <a>version labels</a> to be used with the <code>version</code> parameter 
-      and in <a>concrete RDF syntax</a> are given in the
-      <a href="#defined-version-labels">section below</a>.
+      Section <a href="#defined-version-labels" class="sectionRef"></a> defines <a>version labels</a>
+      to be used with the <code>version</code> parameter and in <a>concrete RDF syntax</a>.
     </p>
   </section>
 </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -545,6 +545,11 @@ Location: http://example.com/document.ttl
 Host: example.com
 Accept: text/turtle; version=1.2
     </pre>
+    <p>
+      Defined <a>version labels</a> to be used with the <code>version</code> parameter 
+      and in <a>concrete RDF syntax</a> are given in the
+      <a href="#defined-version-labels">section below</a>.
+    </p>
   </section>
 </section>
 
@@ -576,6 +581,41 @@ Accept: text/turtle; version=1.2
     another specification.</p>
   <p class="issue" data-number="70">
     Change "Classic Conformance" to "Basic Conformance" and define them as profiles.</p>
+
+  <section id="defined-version-labels">
+    <h3>Version Labels</h3>
+    <p>
+      A <dfn>version label</dfn> is a string that identifies the syntax and semantics conformance
+      for the RDF data.
+    </p>
+    <table id="tab-version-labels" class="simple">
+      <caption>Version Labels</caption>
+      <thead>
+        <tr>
+          <th style="text-align: center">Version Label</th>
+          <th style="text-align: center">Syntax</th>
+          <th style="text-align: center">Semantics</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>"1.2"</td>
+          <td>RDF 1.2 syntax</td>
+          <td><a href="https://www.w3.org/TR/rdf12-semantics/">RDF 1.2 Semantics</a></td>
+        </tr>
+        <tr>
+          <td>"1.2-basic"</td>
+          <td>RDF 1.2 syntax without triple terms</td>
+          <td><a href="https://www.w3.org/TR/rdf12-semantics/">RDF 1.2 Semantics</a></td>
+        </tr>
+        <tr>
+          <td>"1.1"</td>
+          <td>RDF 1.1 syntax except for use of a version directive</td>
+          <td><a href="https://www.w3.org/TR/rdf11-semantics/">RDF 1.1 Semantics</a></td>
+        </tr>
+      </tbody>
+    </table>
+  </section>
 
   <section id="rdf-strings">
     <h3>Strings in RDF</h3>


spec/index.html
```diff
@@ -577,6 +582,41 @@ <h3>RDF Version Announcement</h3>
...
+          <td>RDF 1.2 syntax</td>
+          <td><a href="https://www.w3.org/TR/rdf12-semantics/">RDF 1.2 Semantics</a></td>
+        </tr>
+        <tr>
+          <td>"1.2-basic"</td>
+          <td>RDF 1.2 syntax without triple terms</td>
+          <td><a href="https://www.w3.org/TR/rdf12-semantics/">RDF 1.2 Semantics</a></td>
+        </tr>
+        <tr>
+          <td>"1.1"</td>
```
rubensworks, line 612: It still feels a bit weird to me to introduce a `1.1` version label in the 1.2 spec, while the usage of version directives is illegal in the 1.1 spec.
Is the need for this version label really that large? And can the use cases not be achieved with the `1.2-basic` version label?
Parsers that only support 1.1 will not support version labels anyways, and if they update to 1.2 support, the need for them supporting (only) 1.1 seems small.

spec/index.html
```diff
@@ -577,6 +582,41 @@ <h3>RDF Version Announcement</h3>
...
+      </thead>
+      <tbody>
+        <tr>
+          <td>"1.2"</td>
+          <td>RDF 1.2 syntax</td>
+          <td><a href="https://www.w3.org/TR/rdf12-semantics/">RDF 1.2 Semantics</a></td>
+        </tr>
+        <tr>
+          <td>"1.2-basic"</td>
+          <td>RDF 1.2 syntax without triple terms</td>
```
rubensworks, line 608: Should we add a description (note?) of what parsers are (not) expected to do when they encounter triple terms when the `1.2-basic` version is encountered?

In alignment to the optional nature of version announcement, I would expect that parsers are not required to error if they encounter triple terms under `1.2-basic`, but they may emit a warning.

spec/index.html
afs: It is the "RDF 1.2 without triple terms and without text direction" case (and without `rdf:refies` axiomatic triples).

Two uses cases:
* redistribution
* archiving

Redistribution: A system receiving data labelled "1.1" can use it to know it pass that data onto its own users/clients/downstream which may still be RDF 1.1.

Archiving: Now we have different versions, being able to write files that contain the metadata that the data contents is definitely RDF 1.1 is useful. Having the metadata in the file itself is robust. (Sensible use would be to make the `VERSION` the first line of the file.)

I think now we have introduced the version mechanism, we should provide the possibility to label 1.1 data. Absence of a label leaves the situation unclear. If it does not get used then it does little harm.

spec/index.html
afs: We need a separate discussion about that before we can add something in this section.

@rubensworks - Are you OK with merging this PR now and continuing the wider discussion? (I can raise an issue for it.)

The 1.10 ("RDF Version Announcement") has text about [advised behavior (paragraphs 2 and 3)](https://www.w3.org/TR/rdf12-concepts/#section-version-announcement). Paragraph 3 has a whole document assumption.

Personally, I think we need to be careful with guidance here. It can discourage new uses we haven't thought of.
There isn't deployment practice to record unlike the big note in the IRIs section.

spec/index.html
```diff
@@ -545,6 +545,11 @@ <h3>RDF Version Announcement</h3>
 Host: example.com
 Accept: text/turtle; version=1.2
     </pre>
+    <p>
+      Defined <a>version labels</a> to be used with the <code>version</code> parameter 
+      and in <a>concrete RDF syntax</a> are given in the
+      <a href="#defined-version-labels">section below</a>.
```
hartig, line 551: I had to read this sentence three times to understand what it wants to say, which is due to its use of passive voice and the fact that it begins with the object rather than the subject. I propose to rephrase it as follows.
```suggestion
      Section <a href="#defined-version-labels" class="sectionRef"></a> defines <a>version labels</a>
      to be used with the <code>version</code> parameter and in <a>concrete RDF syntax</a>.
```

spec/index.html
rubensworks: Ok, sounds good 👍

spec/index.html
rubensworks: Ok, I understand the redistribution and archiving cases. Those make sense indeed.

> Absence of a label leaves the situation unclear. If it does not get used then it does little harm.

Indeed. But in that case, the separate discussion section will be important.
Because publishers announcing 1.1 data using `VERSION` will make old 1.1 parsers break, so they should be aware of that.

rubensworks (approved):

hartig (approved):

pchampin (approved):

PR #214, opened 2025-07-06
Clarify section about triple terms and reification

niklasl:
Discussions has indicated that the introduction of triple terms and reification needs some clarification.

Mainly, this proposal:
* Explains triple terms as references to propositions, and clarifying that these as references are not asserted.
* Replaces the problematic formulation:
   > Assertions on the triple term are made using the reifier.

  and moves the theory of that note to the end of the section (and with it the mention of "contradictory").
* Keeps the details about using reifiers in one paragraph of the introduction. That also adds the notion of "triple annotation" alongside "reifying triple", and relates to both in the comment about concrete syntaxes, and from the diagrams.
*  Puts the comment about concrete syntaxes in a distinct paragraph preceding the diagrams.

Finally, it:
* Turns the final comment about nested triple terms into a note, and characterizing it as an advanced practice. (It might even be removed from this section, as there is a corresponding, neutral note in [section 3.1 Triples](https://www.w3.org/TR/rdf12-concepts/#section-triples).)

***
<a href="https://pr-preview.s3.amazonaws.com/niklasl/rdf-concepts/pull/214.html" title="Last updated on Jul 18, 2025, 7:02 PM UTC (0933819)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/214/e98cc0c...niklasl:0933819.html" title="Last updated on Jul 18, 2025, 7:02 PM UTC (0933819)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -369,8 +369,10 @@
       logical relationship; assertions are always made about some
       <a>reifier</a> thereof. There can be multiple, distinct reifiers
       related to the same abstract proposition, such as statements with
-      different sources, or circumstances with different characteristics. Since
-      a proposition so reified does not have to hold, it is possible to make
+      different sources, or circumstances with different characteristics.
+      One reifier may also be used to reify multiple, distinct propositions,
+      expressing different propositional aspects using the same reifier.
+      Since a proposition so reified does not have to hold, it is possible to make
       statements about any kind of statements, including unasserted contradictions.
     </p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -316,8 +316,8 @@
 
     <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
       in another triple. This use is a reference to a <a>proposition</a>. For
-      the triple term to also be asserted, it must appear in a graph as an <a>asserted
-      triple</a>. This allows for statements to be made about statements
+      the triple term to also be asserted, it must [=appear=] in the same graph as both an <a>asserted
+      triple</a> and an unasserted <a>triple term</a>. This allows for statements to be made about statements
       independent of their assertion within an <a>RDF graph</a>.
     </p>
 
@@ -327,7 +327,7 @@
     <p>
       A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
       and the <a>object</a> is a <a>triple term</a>.
-      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and can be the
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and it can be the
       subject or object of other triples.
       When the triple term of a <a>reifying triple</a> also [=appears=] in the
       same graph as an <a>asserted triple</a>, the subset of triples that share
--- a/spec/index.html
+++ b/spec/index.html
@@ -369,8 +369,8 @@
     </figure>
 
     <p class="note">
-      The <a>propositions</a> denoted by <a>triple terms</a> are abstract,
-      logical relationships, and assertions are always made about some
+      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
+      logical relationship; assertions are always made about some
       <a>reifier</a> thereof. There can be multiple, distinct reifiers
       underlying the same abstract proposition, such as statements with
       different sources, or circumstances with different characteristics. Since
--- a/spec/index.html
+++ b/spec/index.html
@@ -384,8 +384,8 @@
         used to reify multiple, distinct propositions, expressing different
         propositional aspects using the same reifier.</p>
       <p>Since a proposition so reified does not have to hold, it is possible
-        to make statements about any kind of statements, including unasserted
-        contradictions.</p>
+        to make statements about any kind of statement, including an unasserted
+        statement that contradicts another statement, whether asserted or not.</p>
     </div>
 
   </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -329,7 +329,7 @@
       and the <a>object</a> is a <a>triple</a> denoted by a <a>triple term</a>.
       The <a>subject</a> of that triple is called a <dfn>reifier</dfn>.
       The <a>IRI</a> or <a>blank node</a> denoting a reifier can be the
-      subject or <a>object</a> of other triples.
+    subject or object of other triples.
       A <a>reifying triple</a> whose object is also an <a>asserted triple</a>
       is called a <dfn>triple annotation</dfn>.
     </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -365,11 +365,13 @@
     </figure>
 
     <p class="note">
-      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
-      logical relationship; assertions are always made about some
-      <a>reifier</a> thereof. There can be multiple, distinct reifiers
-      related to the same abstract proposition, such as statements with
-      different sources, or circumstances with different characteristics.
+      Since <a>triple terms</a> denote abstract, logical <a>propositions</a>,
+      they are only allowed as objects of other triples.
+      Assertions involving them should commonly use <a>reifying triples</a>;
+      the <a>reifier</a> is then used for further description.
+      There can be multiple, distinct reifiers related to the same abstract
+      proposition, such as statements with different sources, or circumstances
+      with different characteristics.
       One reifier may also be used to reify multiple, distinct propositions,
       expressing different propositional aspects using the same reifier.
       Since a proposition so reified does not have to hold, it is possible to make
--- a/spec/index.html
+++ b/spec/index.html
@@ -315,10 +315,7 @@
     <h3>Triple Terms and Reification</h3>
 
     <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
-      in the [=object=] position of another triple.</p>
-
-    <p>This usage of a <a>triple term</a> only functions as a reference to a
-      <a>proposition</a>, and does not in itself imply its truth. For this
+      in another triple. This use is a reference to a <a>proposition</a>. For this
       triple to also be asserted, it must appear in a graph as an <a>asserted
       triple</a>. This allows for statements to be made about statements
       independent of their assertion within an <a>RDF graph</a>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -315,8 +315,8 @@
     <h3>Triple Terms and Reification</h3>
 
     <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
-      in another triple. This use is a reference to a <a>proposition</a>. For this
-      triple to also be asserted, it must appear in a graph as an <a>asserted
+      in another triple. This use is a reference to a <a>proposition</a>. For
+      the triple term to also be asserted, it must appear in a graph as an <a>asserted
       triple</a>. This allows for statements to be made about statements
       independent of their assertion within an <a>RDF graph</a>.
     </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -317,26 +317,29 @@
     <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
       in the [=object=] position of another triple.</p>
 
-    <p>A <a>triple term</a> is not necessarily asserted, allowing
-      statements to be made about other statements that may not be
-      asserted within an <a>RDF graph</a>.
-      This allows statements to be made about relationships that may be contradictory.
-      For a <a>triple term</a> to be asserted,
-      it must also appear in a graph as an <a>asserted triple</a>.</p>
+    <p>This usage of a <a>triple term</a> only functions as a reference to a
+      <a>proposition</a>, and does not in itself imply its truth. For this
+      triple to also be asserted, it must appear in a graph as an <a>asserted
+      triple</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
 
-    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>;
-      such a <a>triple</a> is then a <dfn>reifying triple</dfn>.
-      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
-      Assertions on the <a>triple term</a> are made using the <a>reifier</a>.
-      A concrete syntax may provide a special notation for specifying <a>reifying triples</a>.</p>
+    <p>
+      When a <a>triple term</a> is used as the object of a <a>triple</a> with
+      the predicate <code>rdf:reifies</code>, such a triple is a
+      <dfn>reifying triple</dfn>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>.
+      The <a>IRI</a> or <a>blank node</a> denoting a reifier can be the
+      subject or <a>object</a> of other triples.
+      A <a>reifying triple</a> whose object is also an <a>asserted triple</a>
+      is called a <dfn>triple annotation</dfn>.
+    </p>
 
-    <p class="note">Using the <a>reifier</a> as an indirect way of referencing a <a>triple term</a>
-      allows multiple groups of assertions to be separated from each other,
-      which might be necessary if the same <a>triple term</a> was derived from different sources.
-      Assertions are always made using the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
-      of different triples.
-      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]],
-      may have shortcuts for capturing a <a>triple term</a> with its <a>reifier</a>.</p>
+    <p>
+      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]], may have shortcuts
+      for specifying <a>reifying triples</a> and <a>triple annotations</a> more
+      succinctly.
+    </p>
 
     <p>The following diagram represents a <a>reifying triple</a> of an unasserted, abstract <a>triple term</a>, and a <a>triple</a> relating the <a>reifier</a> to another resource.</p>
 
@@ -351,7 +354,8 @@
     </figure>
 
     <p>Here is a variation on the graph shown in <a href="#fig-triple-term"></a>. This represents a graph
-      where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+      where the <a>triple term</a> corresponds to an <a>asserted triple</a>,
+      qualifying the <a>reifying triple</a> as a <a>triple annotation</a>.
     </p>
 
     <figure id="fig-asserted-triple-term">
@@ -365,7 +369,21 @@
       </figcaption>
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      The <a>propositions</a> denoted by <a>triple terms</a> are abstract,
+      logical relationships, and assertions are always made about some
+      <a>reifier</a> thereof. There can be multiple, distinct reifiers
+      underlying the same abstract proposition, such as statements with
+      different sources, or circumstances with different characteristics. Since
+      a proposition so reified does not have to hold, it is possible to make
+      statements about any kind of statements, including unasserted contradictions.
+    </p>
+
+    <p class="note">
+      While a <a>triple term</a> may also have another <a>triple term</a> as
+      an <a>object</a>, that is considered an advanced practice beyond the
+      regular use of reifiers.</p>
+
   </section>
 
   <section id="change-over-time">
--- a/spec/index.html
+++ b/spec/index.html
@@ -325,9 +325,10 @@
       A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
       and the <a>object</a> is a <a>triple term</a>.
       The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and can be the
-    subject or object of other triples.
-      A <a>reifying triple</a> whose object is also an <a>asserted triple</a>
-      is called a <dfn>triple annotation</dfn>.
+      subject or object of other triples.
+      When the triple term of a <a>reifying triple</a> also [=appears=] in the
+      same graph as an <a>asserted triple</a>, the subset of triples that share
+      the same <a>reifier</a> as subject is called a <dfn>triple annotation</dfn>.
     </p>
 
     <p>
@@ -349,17 +350,17 @@
     </figure>
 
     <p>Here is a variation on the graph shown in <a href="#fig-triple-term"></a>. This represents a graph
-      where the <a>triple term</a> corresponds to an <a>asserted triple</a>,
-      qualifying the <a>reifying triple</a> as a <a>triple annotation</a>.
+      where an <a>asserted triple</a> corresponds to the <a>triple term</a> object of a <a>reifying triple</a>,
+      qualifying the description about this <a>reifier</a> as a <a>triple annotation</a>.
     </p>
 
     <figure id="fig-asserted-triple-term">
       <a href="asserted-triple-term.svg">
         <object data="asserted-triple-term.svg" aria-describedby="fig-asserted-triple-term-alt"
-          aria-label="An RDF graph containing a reifying triple where the triple term corresponds to an asserted triple"></object>
+          aria-label="An RDF graph containing a triple annotation, where the triple term of a reifying triple corresponds to an asserted triple"></object>
       </a>
       <figcaption id="fig-asserted-triple-term-alt">
-        An <a>RDF graph</a> containing a <a>triple annotation</a>, which is a <a>reifying triple</a> whose <a>triple term</a> corresponds to an <a>asserted triple</a>.
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, where the <a>triple term</a> of a <a>reifying triple</a> corresponds to an <a>asserted triple</a>.
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
--- a/spec/index.html
+++ b/spec/index.html
@@ -374,11 +374,6 @@
       statements about any kind of statements, including unasserted contradictions.
     </p>
 
-    <p class="note">
-      While a <a>triple term</a> may also have another <a>triple term</a> as
-      an <a>object</a>, that is considered an advanced practice beyond the
-      regular use of reifiers.</p>
-
   </section>
 
   <section id="change-over-time">
--- a/spec/index.html
+++ b/spec/index.html
@@ -326,7 +326,7 @@
 
     <p>
       A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
-      and the <a>object</a> is a <a>triple</a> denoted by a <a>triple term</a>.
+      and the <a>object</a> is a <a>triple term</a>.
       The <a>subject</a> of that triple is called a <dfn>reifier</dfn>.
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -317,26 +317,29 @@ <h3>Triple Terms and Reification</h3>
...
-      A concrete syntax may provide a special notation for specifying <a>reifying triples</a>.</p>
-
-    <p class="note">Using the <a>reifier</a> as an indirect way of referencing a <a>triple term</a>
-      allows multiple groups of assertions to be separated from each other,
-      which might be necessary if the same <a>triple term</a> was derived from different sources.
-      Assertions are always made using the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
-      of different triples.
-      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]],
-      may have shortcuts for capturing a <a>triple term</a> with its <a>reifier</a>.</p>
+    <p>This usage of a <a>triple term</a> only functions as a reference to a
```
afs, line 320:
```suggestion
    <p>This usage of a <a>triple term</a> is a reference to a
```
or
```suggestion
    <p>This usage of a <a>triple term</a> gives a reference to a
```
"only" seems out of place here.

spec/index.html
```diff
@@ -317,26 +317,29 @@ <h3>Triple Terms and Reification</h3>
...
+      <a>proposition</a>, and does not in itself imply its truth. For this
+      triple to also be asserted, it must appear in a graph as an <a>asserted
```
afs, line 322: Suggestion:
```suggestion
      <a>proposition</a>. For this
      triple to also be asserted, it must appear in a graph as an <a>asserted
```

spec/index.html
```diff
@@ -317,26 +317,29 @@ <h3>Triple Terms and Reification</h3>
...
+      triple</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
+
+    <p>
+      When a <a>triple term</a> is used as the object of a <a>triple</a> with
+      the predicate <code>rdf:reifies</code>, such a triple is a
+      <dfn>reifying triple</dfn>.
```
afs, line 330: Over the edits, this sentence has become a bit convoluted.

Maybe turn it around:

```suggestion
      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
      and the <a>object</a> is a <a>triple</a> denoted by a <a>triple term</a>.
```

spec/index.html
```diff
@@ -360,12 +364,26 @@ <h3>Triple Terms and Reification</h3>
...
+      underlying the same abstract proposition, such as statements with
+      different sources, or circumstances with different characteristics. Since
+      a proposition so reified does not have to hold, it is possible to make
+      statements about any kind of statements, including unasserted contradictions.
+    </p>
+
+    <p class="note">
+      While a <a>triple term</a> may also have another <a>triple term</a> as
+      an <a>object</a>, that is considered an advanced practice beyond the
+      regular use of reifiers.</p>
```
afs, line 385:
```suggestion
      an <a>object</a>, that is considered an advanced practice.</p>
```
or give some substance to "regular use".

spec/index.html
```diff
@@ -360,12 +364,26 @@ <h3>Triple Terms and Reification</h3>
...
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, which is a <a>reifying triple</a> whose <a>triple term</a> corresponds to an <a>asserted triple</a>.
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      The <a>propositions</a> denoted by <a>triple terms</a> are abstract,
+      logical relationships, and assertions are always made about some
+      <a>reifier</a> thereof. There can be multiple, distinct reifiers
```
afs, line 375: Maybe 
```suggestion
      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
      logical relationship; assertions are always made about some
      <a>reifier</a> thereof. There can be multiple, distinct reifiers
```
because it is one triple term for one proposition.

spec/index.html
```diff
@@ -317,26 +317,29 @@ <h3>Triple Terms and Reification</h3>
...
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
+
+    <p>
+      When a <a>triple term</a> is used as the object of a <a>triple</a> with
+      the predicate <code>rdf:reifies</code>, such a triple is a
+      <dfn>reifying triple</dfn>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>.
+      The <a>IRI</a> or <a>blank node</a> denoting a reifier can be the
+      subject or <a>object</a> of other triples.
```
afs, line 333:
```suggestion
    subject or object of other triples.
```
as recently referenced.

afs (commented):
I like the direction of the PR.

spec/index.html
niklasl: Works for me. (Will leave suggestions for a day or so for others to chime in, and then commit them to this PR.)

spec/index.html
niklasl: I agree, much better.

spec/index.html
niklasl: I'll pick this change; unless we agree to remove this paragraph.

franconi: Remove :-)

spec/index.html
```diff
@@ -360,12 +363,26 @@ <h3>Triple Terms and Reification</h3>
...
+      underlying the same abstract proposition, such as statements with
+      different sources, or circumstances with different characteristics. Since
+      a proposition so reified does not have to hold, it is possible to make
+      statements about any kind of statements, including unasserted contradictions.
+    </p>
+
+    <p class="note">
+      While a <a>triple term</a> may also have another <a>triple term</a> as
+      an <a>object</a>, that is considered an advanced practice beyond the
+      regular use of reifiers.</p>
```
niklasl, line 384: The "remove" option (for thumb-up/-down voting):
```suggestion

```

spec/index.html
niklasl: I think this is better. But I'm unsure about the scope we prefer for when referencing again? It seemed to be almost paragraph-by-paragraph in this section (and `object` is only referenced two paragraphs up). I can do this change and clean up some more recently referenced terms in this section if preferred?

spec/index.html
```diff
@@ -360,12 +363,26 @@ <h3>Triple Terms and Reification</h3>
...
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
+      logical relationship; assertions are always made about some
+      <a>reifier</a> thereof. There can be multiple, distinct reifiers
+      underlying the same abstract proposition, such as statements with
+      different sources, or circumstances with different characteristics. Since
+      a proposition so reified does not have to hold, it is possible to make
```
niklasl, line 377: Per the suggestion by @franconi, slightly rephrased:
```suggestion
      different sources, or circumstances with different characteristics.
      One reifier may also reify multiple, distinct propositions, expressing
      different propositional aspects of the same reifier.
      Since a proposition so reified does not have to hold, it is possible to make
```
While I strongly agree with this, in case there is some concern about this as common practice, I'd accept not spelling it out (cf. the possible removal regarding nested triples below). It is still a possible practice of course.

niklasl:
I'd like to combine the first two suggestions by @afs and the first by @franconi, and furthermore join the resulting first and second paragraphs into one; resulting in:
```html
    <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
      in another triple. This usage is a reference to a <a>proposition</a>. For this
      triple to also be asserted, it must appear in a graph as an <a>asserted
      triple</a>. This allows for statements to be made about statements
      independent of their assertion within an <a>RDF graph</a>.
    </p>
```

spec/index.html
niklasl: Sorry, I missed the previous in-paragraph use! I'll commit this suggestion.

(Maybe some more cleanups are needed, but I'll probably stick to the "same paragraph" pattern unless otherwise advised.)

spec/index.html
```diff
@@ -317,26 +317,28 @@ <h3>Triple Terms and Reification</h3>
...
+    <p>This usage of a <a>triple term</a> only functions as a reference to a
+      <a>proposition</a>, and does not in itself imply its truth. For this
+      triple to also be asserted, it must appear in a graph as an <a>asserted
+      triple</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
+
+    <p>
+      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
+      and the <a>object</a> is a <a>triple</a> denoted by a <a>triple term</a>.
```
niklasl, line 329:
```suggestion
      and the <a>object</a> is a <a>triple term</a>.
```

TallTed:
I like @niklasl https://github.com/w3c/rdf-concepts/pull/214#issuecomment-3041941449, with one tweak to change `usage` to `use` (which I feel is appropriate in most if not all cases).

afs (approved):

spec/index.html
```diff
@@ -360,12 +359,21 @@ <h3>Triple Terms and Reification</h3>
...
-        An <a>RDF graph</a> containing a <a>reifying triple</a> where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, which is a <a>reifying triple</a> whose <a>triple term</a> corresponds to an <a>asserted triple</a>.
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
+      logical relationship; assertions are always made about some
```
niklasl, line 369:
```suggestion
      logical relationship; assertions are always made using some
```

spec/index.html
niklasl: I'm inclined to add this. (I updated the suggestion to "use" of reifier, to avoid conflating with its referent.)

pchampin (approved):

spec/index.html
```diff
@@ -315,28 +315,26 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
+
+    <p>
+      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
+      and the <a>object</a> is a <a>triple term</a>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and can be the
+    subject or object of other triples.
+      A <a>reifying triple</a> whose object is also an <a>asserted triple</a>
+      is called a <dfn>triple annotation</dfn>.
```
TallTed, line 330:
```suggestion
      A <a>reifying triple</a> whose object is also an <a>asserted triple</a>
      (elsewhere in the graph) is called a <dfn>triple annotation</dfn>.
```

spec/index.html
TallTed:
```suggestion
      logical relationship; assertions about a <a>proposition</a>
      are always made using some
```

spec/index.html
franconi: Please proceed!

niklasl: Yes; no adverse opinions and lots of eyes here; so committing this. There's still some other suggested adjustments, so I'll suggest leave the PR open for any final reactions. (As I'm not editor here, I won't merge myself, or at least not without an editor asking me to do so.)

spec/index.html
```diff
@@ -315,28 +315,26 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
+      triple to also be asserted, it must appear in a graph as an <a>asserted
+      triple</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
+
+    <p>
+      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
+      and the <a>object</a> is a <a>triple term</a>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and can be the
+    subject or object of other triples.
```
niklasl, line 328:
```suggestion
      subject or object of other triples.
```

spec/index.html
```diff
@@ -360,12 +359,23 @@ <h3>Triple Terms and Reification</h3>
...
-        An <a>RDF graph</a> containing a <a>reifying triple</a> where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, which is a <a>reifying triple</a> whose <a>triple term</a> corresponds to an <a>asserted triple</a>.
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      A <a>proposition</a> denoted by a <a>triple term</a> is an abstract,
+      logical relationship; assertions are always made about some
```
doerthe, line 369: I am not happy with this "abstract, logical relationship" because I think that is confusing for readers. Is a Proposition really a relationship? Basically it is a special kind of resource (not happy with that term either). Maybe just "abstract resource"? But I hope that you have a better idea...

gkellogg: Triples _are_ relationships between resources, so a proposition is denoted by a triple term, it would seem to me that a relationship is some how involved.

niklasl: We do have the time-honored:
> Asserting an RDF triple says that some relationship, indicated by the predicate, holds between the resources

But I agree that the abstract proposition carrying this meaning is an intricate thing to convey succinctly here. Rather than going further into that, I'll attempt a reformulation.

spec/index.html
```diff
@@ -360,12 +359,23 @@ <h3>Triple Terms and Reification</h3>
...
+      <a>reifier</a> thereof. There can be multiple, distinct reifiers
+      related to the same abstract proposition, such as statements with
+      different sources, or circumstances with different characteristics.
```
niklasl, line 372: Attempting to take recent suggestions and comments into account:
```suggestion
      Since <a>triple terms</a> denote abstract, logical <a>propositions</a>,
      assertions involving them require the use of <a>reifying triples</a>. The
      <a>reifier</a> is then used for further description.
      There can be multiple, distinct reifiers related to the same abstract
      proposition, such as statements with different sources, or circumstances
      with different characteristics.
```

TallTed: Are we now saying, not only that reifiers are defined only by their presence as a subject in a triple with `rdf:reifies` as predicate and an explicit triple term as object, but also that an explicit triple term can only be found as the object of an `rdf:reifies` relationship? 

In other words, that non-`rdf:reifies` relationships that might appropriately be expressed in pseudo-turtle as — 
```
:b ex:rship << ex:sub ex:pred ex:obj >> .
``` 
— must instead be expressed as —
```
ex:rei rdf:reifies << ex:sub ex:pred ex:obj >> . 
:b     ex:rship    ex:rei .
```
?

niklasl: Good point. While:
```turtle
:b ex:rship << ex:sub ex:pred ex:obj >> .
```
does expand to:
```turtle
:b     ex:rship    _:rei .
_:rei rdf:reifies <<( ex:sub ex:pred ex:obj )>> . 
```
using a "bare" triple term like:
```turtle
:b ex:rship <<( ex:sub ex:pred ex:obj )>> .
```
is not forbidden (and there may be advanced uses for that, just as for nested triple terms).

So I propose this revision of the above reformulation:
```suggestion
      Since <a>triple terms</a> denote abstract, logical <a>propositions</a>,
      they are only allowed as objects of other triples.
      Assertions involving them should commonly use <a>reifying triples</a>;
      the <a>reifier</a> is then used for further description.
      There can be multiple, distinct reifiers related to the same abstract
      proposition, such as statements with different sources, or circumstances
      with different characteristics.
```
(Still solely focused on reification use, which is the topic of this section, but other than being explicit about the use as objects, avoiding prohibitive language. Cf. #218.)

spec/index.html
```diff
@@ -360,12 +359,23 @@ <h3>Triple Terms and Reification</h3>
...
+      One reifier may also be used to reify multiple, distinct propositions,
+      expressing different propositional aspects using the same reifier.
```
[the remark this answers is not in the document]
niklasl: They should just get pushed down a bit; I'll ensure they're not lost.

spec/index.html
```diff
@@ -360,12 +359,25 @@ <h3>Triple Terms and Reification</h3>
           aria-label="An RDF graph containing a reifying triple where the triple term corresponds to an asserted triple"></object>
       </a>
       <figcaption id="fig-asserted-triple-term-alt">
-        An <a>RDF graph</a> containing a <a>reifying triple</a> where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, which is a <a>reifying triple</a> whose <a>triple term</a> corresponds to an <a>asserted triple</a>.
```
niklasl, line 362: Adapt to updated triple annotation definition:
```suggestion
        An <a>RDF graph</a> containing a <a>triple annotation</a>, where the <a>triple term</a> of a <a>reifying triple</a> corresponds to an <a>asserted triple</a>.
```

TallTed (approved):

spec/index.html
```diff
@@ -315,28 +315,27 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
-    <p class="note">Using the <a>reifier</a> as an indirect way of referencing a <a>triple term</a>
-      allows multiple groups of assertions to be separated from each other,
-      which might be necessary if the same <a>triple term</a> was derived from different sources.
-      Assertions are always made using the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
-      of different triples.
-      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]],
-      may have shortcuts for capturing a <a>triple term</a> with its <a>reifier</a>.</p>
+      in another triple. This use is a reference to a <a>proposition</a>. For this
+      triple to also be asserted, it must appear in a graph as an <a>asserted
+      triple</a>. This allows for statements to be made about statements
```
pchampin, line 320:
```suggestion
      in another triple. This use is a reference to a <a>proposition</a>. For
      the triple term to also be asserted, it must appear in a graph as an <a>asserted
      triple</a>. This allows for statements to be made about statements
```
"this triple" is highly ambiguous in that sentence! ("that triple" would probably be more accurate, but still too ambiguous to my taste).

spec/index.html
```diff
@@ -351,21 +350,35 @@ <h3>Triple Terms and Reification</h3>
...
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
 
-    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
+    <p class="note">
+      Since <a>triple terms</a> denote abstract, logical <a>propositions</a>,
+      they are only allowed as objects of other triples.
+      Assertions involving them should commonly use <a>reifying triples</a>;
+      the <a>reifier</a> is then used for further description.
```
pchampin, line 372: I'm not quite comfortable with these two sentences:
* The causality conveyed by "since" is not obvious, and even controversial (remember that we did not reach consensus on this). Therefore I would not even try to justify this constraint.
* In the 2nd sentence "assertions involving [propositions]" may give the wrong impressions that the reifier also denotes the proposition.


```suggestion
      <a>Triple terms</a> always denote abstract, logical <a>propositions</a>,
      while <a>reifiers</a> may denote a variety of things that are related to,
      or described by, these propositions
      (e.g. a statement or a belief that the proposition holds, an event or a circumstance that makes the proposition true, etc.).
      The vast majority of use-cases brought to the Working Group deal with resources of the latter category;
      it is therefore expected that triple terms are mostly (if not exclusively) used as <a>object</a> of <a>reifying triples</a>,
      and that the reifiers (rather than the triple terms) are used in further descriptions.
```

pchampin (changes requested):
Sorry to chime in late with a request for change. I think the ambiguity in my first comment really needs to be fixed. My 2nd comment is more a matter of opinion, but I would like to discuss it.

spec/index.html
niklasl: I sympathize with this suggestion (and while I maintain my position on the topic, this should certainly be neutrally phrased). Some thoughts:

* Should the Working Group be mentioned as such here?
* Reifiers "described by [these] propositions" might perhaps convey that the propositions related to them as subjects or objects ("weak self-referencing")... While I'd _like_ to get into the details (propositions being induced, reifiers "underlying" them, etc.), I think we should merely say "related to" here (implying via `rdf:reifies`).
* I think we got to "commonly" to stay neutral about other uses, so I'd suggest we keep that (rather than "mostly (if not exclusively)", which I otherwise mostly agree with (if not exclusively...)).

I'll propose a variant based on your suggestion and these thoughts (and change the note to a `<div>` with `<p>`:s, it was already a bit dense).

spec/index.html
```diff
@@ -351,21 +350,35 @@ <h3>Triple Terms and Reification</h3>
...
+      There can be multiple, distinct reifiers related to the same abstract
+      proposition, such as statements with different sources, or circumstances
+      with different characteristics.
+      One reifier may also be used to reify multiple, distinct propositions,
+      expressing different propositional aspects using the same reifier.
+      Since a proposition so reified does not have to hold, it is possible to make
+      statements about any kind of statements, including unasserted contradictions.
+    </p>
```
niklasl, line 380: Based on the suggestion by @pchampin:
```suggestion
    <div class="note">
      <p><a>Triple terms</a> always denote abstract, logical <a>propositions</a>,
        while <a>reifiers</a> may denote a variety of things that are related
        to these propositions (such as a statement or belief that the
        proposition holds, or an event or circumstance that makes the
        proposition true).
        It is therefore expected that triple terms are commonly used as
        <a>objects</a> of <a>reifying triples</a>, and that the reifiers
        (rather than the triple terms) are used in further descriptions.</p>
      <p>
        There can be multiple, distinct reifiers related to the same abstract
        proposition, such as statements with different sources, or
        circumstances with different characteristics. One reifier may also be
        used to reify multiple, distinct propositions, expressing different
        propositional aspects using the same reifier.</p>
      <p>Since a proposition so reified does not have to hold, it is possible
        to make statements about any kind of statements, including unasserted
        contradictions.</p>
    </div>
```

pchampin (approved):

franconi (approved):

spec/index.html
```diff
@@ -365,19 +368,25 @@ <h3>Triple Terms and Reification</h3>
...
+        (rather than the triple terms) are used in further descriptions.</p>
+      <p>
+        There can be multiple, distinct reifiers related to the same abstract
+        proposition, such as statements with different sources, or
+        circumstances with different characteristics. One reifier may also be
+        used to reify multiple, distinct propositions, expressing different
+        propositional aspects using the same reifier.</p>
+      <p>Since a proposition so reified does not have to hold, it is possible
+        to make statements about any kind of statements, including unasserted
+        contradictions.</p>
```
TallTed, line 388:
```suggestion
      <p>Since a proposition so reified does not have to hold, it is possible
        to make statements about any kind of statement, including an unasserted
        statement that contradicts another statement, whether asserted or not.</p>
```

spec/index.html
```diff
@@ -315,31 +315,30 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
 
-    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>;
-      such a <a>triple</a> is then a <dfn>reifying triple</dfn>.
-      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
-      Assertions on the <a>triple term</a> are made using the <a>reifier</a>.
-      A concrete syntax may provide a special notation for specifying <a>reifying triples</a>.</p>
+    <p>
+      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
+      and the <a>object</a> is a <a>triple term</a>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and can be the
```
TallTed, line 330:
```suggestion
      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and it can be the
```

spec/index.html
```diff
@@ -315,31 +315,30 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
-    <p>A <a>triple term</a> is not necessarily asserted, allowing
-      statements to be made about other statements that may not be
-      asserted within an <a>RDF graph</a>.
-      This allows statements to be made about relationships that may be contradictory.
-      For a <a>triple term</a> to be asserted in a graph,
-      it must appear as an <a>asserted triple</a> in that same graph.</p>
+      in another triple. This use is a reference to a <a>proposition</a>. For
+      the triple term to also be asserted, it must appear in a graph as an <a>asserted
+      triple</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
```
TallTed, line 321:
```suggestion
      in another triple. This use is a reference to a <a>proposition</a>. For
      the triple term to also be asserted, it must [=appear=] in the same graph as both an <a>asserted
      triple</a> and an unasserted <a>triple term</a>. This allows for statements to be made about statements
      independent of their assertion within an <a>RDF graph</a>.
```

spec/index.html
```diff
@@ -315,31 +315,30 @@ <h3>RDF Vocabularies and Namespace IRIs</h3>
...
+    </p>
 
     <p>RDF terms that appear in a triple term have the same denotation as when they appear in an asserted triple in the graph.
```
TallTed, line 324:
```suggestion
    <p>RDF terms that [=appear=] in a triple term have the same denotation as when they [=appear=] in an asserted triple in the graph. 
```

TallTed (changes requested):
There's always something...

spec/index.html
niklasl: Or we mark them all?
```suggestion
    <p>[=RDF terms=] that [=appear=] in a [=triple term=] have the same
      <a data-lt="denotes">denotation</a> as when they appear in an
      [=asserted triple=] in the <a data-lt="RDF graph">graph</a>.
      For this reason, we say that triple terms are <i>transparent</i>.</p>
```
(Do we also want to `<dfn>transparent</dfn>`?)

TallTed: I'm of very mixed mind on this. Sometimes linking every occurrence of (one, some, many, all) term(s) makes sense. Sometimes "first occurrence" on a page is the rule (typically for output on paper). Sometimes "first occurrence" in a paragraph (but then what about when paragraphs are short, one-liners? or huge, 30-liners? and what about list entries? among other neighboring cases...).

W3C Style Guide is noncommittal on this, suggesting that "[at least the first mention of a source](https://www.w3.org/guide/manual-of-style/#citation)" should be linked. It says nothing about how much additional linkage is recommended or desirable.

My rule of thumb is that links should happen sufficiently often to be conveniently reached from every instance of the linked term. Often that means linking every instance. This also has the benefit of being explicit that each linked instance has the same meaning, where an unlinked instance *might* be read as having a different meaning.

All of which brings me to *generally* linking every instance.

niklasl: I followed the pattern I see in this document (at least within section 1), which links first occurrence of each paragraph (at most). I thus don't think marking up the same term within the same sentence is useful (to me it raises the question if any non-linked use of a term is less specific). If paragraphs are long, that might warrant repetitive linking.

Issue #216, opened 2025-07-07
Add notion of transparency in 1.5 Triple Terms and Reification


franconi:
It may be useful to add after the text:

> ... For a [triple term](https://w3c.github.io/rdf-concepts/spec/#dfn-triple-term) to be asserted, it must also appear in a graph as an [asserted triple](https://w3c.github.io/rdf-concepts/spec/#dfn-asserted-triple).

the following text:

> The RDF terms appearing in a triple term are interpreted in the same way they would be interpreted when appearing in asserted triples appearing in the graph; we say that triple terms are _transparent_.

in order avoid people believe that triple terms are like literals.

franconi:
Created [PR #219](https://github.com/w3c/rdf-concepts/pull/219).

PR #219, opened 2025-07-11
Add notion of transparency in 1.5 Triple Terms and Reification

franconi:
Following the discussion in [issue #216](https://github.com/w3c/rdf-concepts/issues/216), I have added after the text:

> ... For a [triple term](https://w3c.github.io/rdf-concepts/spec/#dfn-triple-term) to be asserted, it must also appear in a graph as an [asserted triple](https://w3c.github.io/rdf-concepts/spec/#dfn-asserted-triple).

the following text:

> RDF terms that appear in a triple term have the same denotation as when they appear in an asserted triple in the graph. For this reason, we say that triple terms are _transparent_.

in order avoid people believe that triple terms are like literals.

***
<a href="https://pr-preview.s3.amazonaws.com/unibz-krdb/rdf-concepts-franconi-krdb/pull/219.html" title="Last updated on Jul 17, 2025, 2:32 PM UTC (41ae00d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/219/9714dbd...unibz-krdb:41ae00d.html" title="Last updated on Jul 17, 2025, 2:32 PM UTC (41ae00d)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -321,8 +321,8 @@
       statements to be made about other statements that may not be
       asserted within an <a>RDF graph</a>.
       This allows statements to be made about relationships that may be contradictory.
-      For a <a>triple term</a> to be asserted,
-      it must also appear in a graph as an <a>asserted triple</a>.</p>
+      For a <a>triple term</a> to be asserted in a graph,
+      it must appear as an <a>asserted triple</a> in that same graph.</p>
 
     <p>RDF terms that appear in a triple term have the same denotation as when they appear in an asserted triple in the graph. 
       For this reason, we say that triple terms are <i>transparent</i>.</p>


niklasl (approved):
I fully support this. Just note that depending on which is merged first, either this or #214 needs some adjustment (a rebase and likely some manual tweaking). That's OK by me; this is simpler and likely to be merged first. I would be equally OK for this to go into #214 as an added suggestion.

gkellogg:
The language should probably say that a triple term is asserted if it appears as an asserted triple in the _same_ graph. Otherwise, the language might say that a triple term is asserted if _any_ graph contains it as an asserted triple.

spec/index.html
```diff
@@ -324,6 +324,9 @@ <h3>Triple Terms and Reification</h3>
       For a <a>triple term</a> to be asserted,
       it must also appear in a graph as an <a>asserted triple</a>.</p>
```
TallTed, line 325: I think @gkellogg makes [a good point](https://github.com/w3c/rdf-concepts/pull/219#issuecomment-3062916373). To address it, perhaps the following:
```suggestion
       For a <a>triple term</a> to be asserted in a graph,
       it must appear as both
       an <a>asserted triple</a>
       and a <a>triple term</a>
       in that same graph.</p>
```

gkellogg: The proposed language here is a bit redundant, I think. It could just be the following:

```suggestion
      For a <a>triple term</a> to be asserted in a graph,
      it must appear as
      an <a>asserted triple</a>
      in that same graph.</p>
```

TallTed (approved):
Approved with @gkellogg's rephrased sentence.

spec/index.html
pchampin: I like @gkellogg's suggestion better. I also notice that the original text used "also" ("must also appear"), which i would rather keep...

PR #220, opened 2025-07-17
Annotations on assserted triples are based on operational semantics

rat10:
Add clarification that, according to the model-theoretic semantics, annotations on reified triples can't annotate asserted triples, but that an operational semantics establsihed voa best practices, examples, etc may establish such a semantics in practice.

***
<a href="https://pr-preview.s3.amazonaws.com/rat10/rdf-concepts/pull/220.html" title="Last updated on Jul 24, 2025, 11:14 AM UTC (fa4e78d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/220/97140a2...rat10:fa4e78d.html" title="Last updated on Jul 24, 2025, 11:14 AM UTC (fa4e78d)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -271,10 +271,10 @@
     <p>The <a>IRIs</a> in an <a>RDF vocabulary</a> often begin with
       a common substring known as a <dfn>namespace IRI</dfn>.
       Some namespace IRIs are associated by convention with a short name
-      known as a <dfn>namespace prefix</dfn>. Some examples:
+      known as a <dfn>namespace prefix</dfn>.
 
     <table id="tab-vocab-ns" class="simple">
-      <caption>Some example namespace prefixes and IRIs</caption>
+      <caption>Some namespace prefixes and IRIs used in this specification</caption>
       <tr>
         <th>Namespace prefix</th>
         <th>Namespace IRI</th>
@@ -295,16 +295,20 @@
       </tr>
     </table>
 
-    <p>In some serialization formats it is common to abbreviate <a>IRIs</a>
-      that start with <a>namespace IRIs</a> by using a
-      <a>namespace prefix</a> in order to assist readability. For example, the IRI
-      <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral</code>
+    <p>In some serialization formats, it is common
+      to associate some [=namespace IRIs=] with arbitrary [=namespace prefixes=],
+      and to improve readability by abbreviating <a>IRIs</a> that start with one of those <a>namespace IRIs</a>
+      by using the corresponding <a>namespace prefix</a>.
+      For example,
+      based on the prefix mapping in <a href="#tab-vocab-ns">the table above</a>,
+      the IRI <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral</code>
       would be abbreviated as <code>rdf:XMLLiteral</code>.
-      Note however that these abbreviations are <em>not</em> valid IRIs,
-      and must not be used in contexts where IRIs are expected.
-      Namespace IRIs and namespace prefixes are <em>not</em> a formal part of the
-      RDF data model. They are merely a syntactic convenience for
-      abbreviating IRIs.</p>
+      Note however that such abbreviations are <em>not</em> meant to be processed directly as IRIs,
+      and are not to be used in syntactic contexts where IRIs are expected.
+      Note also that [=namespace IRIs=] and [=namespace prefixes=] are not a formal part of the RDF data model.
+      They are merely a syntactic convenience for abbreviating IRIs;
+      for processing, the actual IRIs are reconstructed by replacing each namespace prefix with the corresponding namespace IRI.
+    </p>
 
     <p>The term “<dfn class="lint-ignore">namespace</dfn>” on its own does not have a
       well-defined meaning in the context of RDF, but is sometimes informally
@@ -315,28 +319,33 @@
     <h3>Triple Terms and Reification</h3>
 
     <p>A <a>triple term</a> is an [=RDF triple=] used as an [=RDF term=]
-      in the [=object=] position of another triple.</p>
+      in another triple. This use is a reference to a <a>proposition</a>. For
+      the triple term to also be asserted, it must [=appear=] in the same graph as both an <a>asserted
+      triple</a> and an unasserted <a>triple term</a>. This allows for statements to be made about statements
+      independent of their assertion within an <a>RDF graph</a>.
+    </p>
 
-    <p>A <a>triple term</a> is not necessarily asserted, allowing
-      statements to be made about other statements that may not be
-      asserted within an <a>RDF graph</a>.
-      This allows statements to be made about relationships that may be contradictory.
-      For a <a>triple term</a> to be asserted,
-      it must also appear in a graph as an <a>asserted triple</a>.</p>
+    <p>[=RDF terms=] that [=appear=] in a [=triple term=] have the same
+      <a data-lt="denotes">denotation</a> as when they appear in an
+      [=asserted triple=] in the <a data-lt="RDF graph">graph</a>.
+      For this reason, we say that triple terms are
+      <dfn class="lint-ignore">transparent</dfn>.</p>
 
-    <p>A <a>triple term</a> can be used as the object of a <a>triple</a> with the predicate <code>rdf:reifies</code>;
-      such a <a>triple</a> is then a <dfn>reifying triple</dfn>.
-      The <a>subject</a> of that <a>triple</a> is called a <dfn>reifier</dfn>.
-      Assertions on the <a>triple term</a> are made using the <a>reifier</a>.
-      A concrete syntax may provide a special notation for specifying <a>reifying triples</a>.</p>
+    <p>
+      A <dfn>reifying triple</dfn> is a triple where the <a>predicate</a> is <code>rdf:reifies</code>
+      and the <a>object</a> is a <a>triple term</a>.
+      The <a>subject</a> of that triple is called a <dfn>reifier</dfn>, and it can be the
+      subject or object of other triples.
+      When the triple term of a <a>reifying triple</a> also [=appears=] in the
+      same graph as an <a>asserted triple</a>, the subset of triples that share
+      the same <a>reifier</a> as subject is called a <dfn>triple annotation</dfn>.
+    </p>
 
-    <p class="note">Using the <a>reifier</a> as an indirect way of referencing a <a>triple term</a>
-      allows multiple groups of assertions to be separated from each other,
-      which might be necessary if the same <a>triple term</a> was derived from different sources.
-      Assertions are always made using the <a>reifier</a>, which can be the <a>subject</a> or <a>object</a>
-      of different triples.
-      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]],
-      may have shortcuts for capturing a <a>triple term</a> with its <a>reifier</a>.</p>
+    <p>
+      Concrete syntaxes, such as Turtle [[RDF12-TURTLE]], may have shortcuts
+      for specifying <a>reifying triples</a> and <a>triple annotations</a> more
+      succinctly.
+    </p>
 
     <p>The following diagram represents a <a>reifying triple</a> of an unasserted, abstract <a>triple term</a>, and a <a>triple</a> relating the <a>reifier</a> to another resource.</p>
 
@@ -351,38 +360,61 @@
     </figure>
 
     <p>Here is a variation on the graph shown in <a href="#fig-triple-term"></a>. This represents a graph
-      where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+      where an <a>asserted triple</a> corresponds to the <a>triple term</a> object of a <a>reifying triple</a>,
+      qualifying the description about this <a>reifier</a> as a <a>triple annotation</a>.
     </p>
 
     <figure id="fig-asserted-triple-term">
       <a href="asserted-triple-term.svg">
         <object data="asserted-triple-term.svg" aria-describedby="fig-asserted-triple-term-alt"
-          aria-label="An RDF graph containing a reifying triple where the triple term corresponds to an asserted triple"></object>
+          aria-label="An RDF graph containing a triple annotation, where the triple term of a reifying triple corresponds to an asserted triple"></object>
       </a>
       <figcaption id="fig-asserted-triple-term-alt">
-        An <a>RDF graph</a> containing a <a>reifying triple</a> where the <a>triple term</a> corresponds to an <a>asserted triple</a>.
+        An <a>RDF graph</a> containing a <a>triple annotation</a>, where the <a>triple term</a> of a <a>reifying triple</a> corresponds to an <a>asserted triple</a>.
         The diagram represents the proposition as a fact using the asserted triple, meaning that the relationship holds.
       </figcaption>
     </figure>
 
+    <div class="note">
+      <p><a>Triple terms</a> always denote abstract, logical <a>propositions</a>,
+        while <a>reifiers</a> may denote a variety of things that are related
+        to these propositions (such as a statement or belief that the
+        proposition holds, or an event or circumstance that makes the
+        proposition true).
+        It is therefore expected that triple terms are commonly used as
+        <a>objects</a> of <a>reifying triples</a>, and that the reifiers
+        (rather than the triple terms) are used in further descriptions.</p>
+      <p>
+        There can be multiple, distinct reifiers related to the same abstract
+        proposition, such as statements with different sources, or
+        circumstances with different characteristics. One reifier may also be
+        used to reify multiple, distinct propositions, expressing different
+        propositional aspects using the same reifier.</p>
+      <p>Since a proposition so reified does not have to hold, it is possible
+        to make statements about any kind of statement, including an unasserted
+        statement that contradicts another statement, whether asserted or not.</p>
+    </div>
+
     <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
 
-   <p class="note">
-	This document advises the reader to understand that
-	an assertion on a reified triple 
-	is an assertion on an asserted triple of the same form, 
-	if such a triple occurs in a graph as both a reified triple 
-	and an asserted triple.
-	However, this understanding is not backed by the model-theoretic semantics 
-	as laid out in RDF 1.2 Semantics;
-	i.e., co-reference is not defined via a model-theoretic formalisation of
-	<a href="https://www.w3.org/TR/rdf12-concepts/#dfn-denote">denotation</a>,
-	but relies on an operational semantics of
-	<a href="https://w3c.github.io/rdf-semantics/spec/#dfn-identify">identification</a>, 
-	established via best practices, examples, and convention
-	in this document, RDF 1.2 Primer, and RDF 1.2 Note on triple terms (tbd).
-   </p>
+   <div class="note">
+     <p>This document advises the reader to understand that
+	    an assertion on a reified triple 
+	    is an assertion on an asserted triple of the same form, 
+	    if such a triple occurs in a graph as both a reified triple 
+	    and an asserted triple.
+	    However, this understanding is not backed by the model-theoretic semantics 
+	    as laid out in RDF 1.2 Semantics;
+	    i.e., co-reference is not defined via a model-theoretic formalisation of
+	    <a href="https://www.w3.org/TR/rdf12-concepts/#dfn-denote">denotation</a>,
+	    but relies on an operational semantics of
+	    <a href="https://w3c.github.io/rdf-semantics/spec/#dfn-identify">identification</a>, 
+	    established via best practices, examples, and convention
+	    in this document, RDF 1.2 Primer, and RDF 1.2 Note on triple terms (tbd).</p>
+   </div>
 	  
+    
+    
   </section>
 
   <section id="change-over-time">
@@ -596,11 +628,11 @@ Accept: text/turtle; version=1.2
   <p>This specification establishes two conformance levels:</p>
 
   <ul>
-    <li><dfn class="no-export lint-ignore" data-lt="full">Full conformance</dfn>
+    <li><dfn class="export" data-local-lt="full">Full conformance</dfn>
       supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
       with <a>triples</a> that contain <a>triple terms</a>.
       </li>
-    <li><dfn class="no-export lint-ignore" data-lt="basic">Basic conformance</dfn>
+    <li><dfn class="export" data-local-lt="basic">Basic conformance</dfn>
       only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>
       with <a>triples</a> that do not contain <a>triple terms</a>.</li>
   </ul>
@@ -612,7 +644,7 @@ Accept: text/turtle; version=1.2
   <section id="defined-version-labels">
     <h3>Version Labels</h3>
     <p>
-      A <dfn>version label</dfn> is a string that identifies the syntax and semantics conformance
+      A <dfn class="export">version label</dfn> is a string that identifies the syntax and semantics conformance
       for the RDF data.
     </p>
     <table id="tab-version-labels" class="simple">
@@ -691,8 +723,8 @@ Accept: text/turtle; version=1.2
     <h3>Strings in RDF</h3>
 
     <p>RDF uses Unicode [[Unicode]] as the fundamental representation for string values.
-      Within this, and related specifications, the term <dfn id="dfn-rdf-string">string</dfn>,
-      or <a data-lt="string">RDF string</a>,
+      Within this and related specifications, the term <dfn id="dfn-rdf-string" class="export" data-local-lt="string">RDF string</dfn>,
+      or simply [=string=],
       is used to describe an ordered sequence of zero or more
       <a data-cite="I18N-GLOSSARY#dfn-code-point" class="lint-ignore">Unicode code points</a>
       which are <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>.
@@ -715,7 +747,7 @@ Accept: text/turtle; version=1.2
 <section id="section-rdf-graph">
   <h2>RDF Graphs</h2>
 
-  <p>An <dfn>RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
+  <p>An <dfn class="export">RDF graph</dfn> is a set of <a>RDF triples</a>.</p>
 
   <p>An [=RDF triple=] that is an element of an [=RDF graph=] is also said to be <dfn id="dfn-asserted-triple" class=export data-lt="asserted triple|asserted triples" data-lt-no-plural=""
       >asserted</dfn> in that [=RDF graph=]
@@ -746,7 +778,7 @@ Accept: text/turtle; version=1.2
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
-    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn> of an <a>RDF graph</a>
+    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
       is the set of <a>subjects</a> and <a>objects</a> of the <a>asserted triples</a> of the graph.
       It is possible for a [=predicate=] [=IRI=] to also occur as a [=node=] in
       the same graph.</p>
@@ -773,13 +805,13 @@ Accept: text/turtle; version=1.2
 
     <p><a>IRIs</a>, <a>literals</a>,
       <a>blank nodes</a>, and <a>triple terms</a> are collectively known as
-      <span id="dfn-rdf-terms"><!-- obsolete term--></span><dfn data-lt="rdf term">RDF terms</dfn>.</p>
+      <span id="dfn-rdf-terms"><!-- obsolete term--></span><dfn class="export" data-lt="rdf term">RDF terms</dfn>.</p>
 
     <p>
       [=IRIs=], [=literals=] and [=blank nodes=] are said to be <dfn class="export" data-lt-no-plural>atomic</dfn> [=RDF terms=].
     </p>
 
-    <p><dfn>RDF term equality</dfn>:
+    <p><dfn class="export">RDF term equality</dfn>:
       Two [=RDF terms=] |t| and <var>t'</var> are equal (the same [=RDF term=]) if and only if
       one of the following four conditions holds:</p>
 
@@ -809,7 +841,7 @@ Accept: text/turtle; version=1.2
       [=triple term=] [=appearing=] in that graph.</p>
 
     <p>
-      An [=RDF term=] is said to be <dfn>ground</dfn> if any of the following three conditions holds:
+      An [=RDF term=] is said to be <dfn class="export">ground</dfn> if any of the following three conditions holds:
     </p>
 
     <ul>
@@ -829,7 +861,7 @@ Accept: text/turtle; version=1.2
   <section id="section-IRIs">
     <h3>IRIs</h3>
 
-    <p>An <dfn data-lt="iri"><abbr title="Internationalized Resource Identifier">IRI</abbr></dfn>
+    <p>An <dfn class="export" data-lt="iri"><abbr title="Internationalized Resource Identifier">IRI</abbr></dfn>
       (Internationalized Resource Identifier) within an RDF graph
       is a <a>string</a> that conforms to the syntax
       defined in RFC 3987 [[!RFC3987]].</p>
@@ -939,18 +971,18 @@ Accept: text/turtle; version=1.2
 
     <p>Literals are used for values such as strings, numbers, and dates.</p>
 
-    <p>A <dfn data-local-lt="RDF literal">literal</dfn> consists of
+    <p>A <dfn class="export" data-lt="RDF literal">literal</dfn> consists of
       two, three, or four components, as below:</p>
 
     <ol>
-      <li>A <dfn>lexical form</dfn>, being an <a data-lt="string">RDF string</a>.
-      <li>A <dfn>datatype IRI</dfn>, being an <a>IRI</a>
+      <li>A <dfn class="export">lexical form</dfn>, being an <a data-lt="string">RDF string</a>.</li>
+      <li>A <dfn class="export">datatype IRI</dfn>, being an <a>IRI</a>
         identifying a datatype that determines how the lexical form maps
         to a <a>literal value</a>.</li>
       <li>If and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#langString</code> or
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>, a
-        non-empty <dfn>language tag</dfn> as defined by [[!BCP47]]. The
+        non-empty <dfn class="export">language tag</dfn> as defined by [[!BCP47]]. The
         language tag MUST be well-formed according to
         <a data-cite="bcp47#section-2.2.9">section 2.2.9</a>
         of [[!BCP47]],
@@ -958,15 +990,15 @@ Accept: text/turtle; version=1.2
         Two [[!BCP47]]-complying strings that differ only by case represent the same [=language tag=].</li>
       <li>If and only if the <a>datatype IRI</a> is
         <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString</code>,
-        a <dfn>base direction</dfn> that MUST be one of the following:<ul>
+        a <dfn class="export">base direction</dfn> that MUST be one of the following:<ul>
           <li>`ltr`, indicating that the initial text direction is set to left-to-right</li>
           <li>`rtl`, indicating that the initial text direction is set to right-to-left</li>
         </ul></li>
     </ol>
 
-    <p>A literal is a <dfn>language-tagged string</dfn> if the [=language tag=]
+    <p>A literal is a <dfn class="export">language-tagged string</dfn> if the [=language tag=]
       is present and the [=base direction=] is <em>not</em> present.
-      A literal is a <dfn id="dfn-dir-lang-string">directional language-tagged string</dfn>
+      A literal is a <dfn id="dfn-dir-lang-string" class="export">directional language-tagged string</dfn>
       if both the [=language tag=] and the [=base direction=] are present.
     </p>
 
@@ -1033,7 +1065,7 @@ Accept: text/turtle; version=1.2
     <section>
       <h2>Literal value</h2>
 
-        <p>The <dfn>literal value</dfn> associated with a <a>literal</a> is defined as follows.</p>
+        <p>The <dfn class="export">literal value</dfn> associated with a <a>literal</a> is defined as follows.</p>
 
       <ul>
         <li>If the literal is a <a>language-tagged string</a>,
@@ -1048,7 +1080,7 @@ Accept: text/turtle; version=1.2
               of the <a>datatype</a>, then the literal value is the result of applying
               the <a>lexical-to-value mapping</a> of the datatype to the
               <a>lexical form</a>.</li>
-            <li>Otherwise, the literal is <dfn data-lt-no-plural>ill-typed</dfn> and no literal value can be
+            <li>Otherwise, the literal is <dfn class="export" data-lt-no-plural>ill-typed</dfn> and no literal value can be
                associated with the literal. Such a case produces a semantic
                [=inconsistency=], but it is not <em>syntactically</em> ill-formed.
                Implementations SHOULD accept [=ill-typed=] literals and produce RDF
@@ -1082,7 +1114,7 @@ Accept: text/turtle; version=1.2
   <section id="section-blank-nodes">
     <h3>Blank Nodes</h3>
 
-    <p><dfn data-lt="blank node">Blank nodes</dfn> are disjoint from
+    <p><dfn class="export" data-lt="blank node">Blank nodes</dfn> are disjoint from
       <a>IRIs</a> and <a>literals</a>.  Otherwise,
       the set of possible blank nodes is arbitrary.  RDF makes no reference to
       any internal structure of blank nodes.</p>
@@ -1125,12 +1157,6 @@ Accept: text/turtle; version=1.2
     <p>Triple term equality:
       Since triple terms are [=triples=], equality of triple terms is the same as [=triple equality=].</p>
       
-    <p class="note">Every <a>triple</a> with a <a>triple term</a> as its [=object=] SHOULD
-      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
-      as its <a>predicate</a>.
-      Every <a>triple</a> whose <a>object</a> is not a <a>triple term</a> SHOULD NOT
-      use <code>http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies</code> (<code>rdf:reifies</code>)
-      as its <a>predicate</a>.</p>
   </section>
 
   <section id="graph-isomorphism">
@@ -1246,15 +1272,15 @@ Accept: text/turtle; version=1.2
 <section id="section-dataset">
   <h2>RDF Datasets</h2>
 
-  <p>An <dfn>RDF dataset</dfn> is a collection of
+  <p>An <dfn class="export">RDF dataset</dfn> is a collection of
     <a>RDF graphs</a>, and comprises:</p>
 
   <ul>
-    <li>Exactly one <dfn>default graph</dfn>, being an <a>RDF graph</a>.
+    <li>Exactly one <dfn class="export">default graph</dfn>, being an <a>RDF graph</a>.
       The default graph does not have a name and MAY be empty.</li>
-    <li>Zero or more <span id="dfn-named-graphs"><!-- obsolete term--></span><dfn data-lt="named graph">named graphs</dfn>.
+    <li>Zero or more <span id="dfn-named-graphs"><!-- obsolete term--></span><dfn class="export" data-lt="named graph">named graphs</dfn>.
       Each named graph is a pair consisting of an <a>IRI</a> or a blank node
-      (the <dfn>graph name</dfn>), and an <a>RDF graph</a>.
+      (the <dfn class="export">graph name</dfn>), and an <a>RDF graph</a>.
       Graph names are unique within an RDF dataset.</li>
   </ul>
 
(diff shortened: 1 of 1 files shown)


franconi:
I disagree totally - as I told you several times. 

> This document advises to understand an assertion on a reified triple as an assertion on an asserted triple of the same form, if such a triple occurs in a graph as both a reified triple and an asserted triple. However, this understanding is not backed by the model-theoretic semantics as layed out in RDF 1.2 Semantics. 

As I already replied to you at the last meeting, there is a precise and intuitive formal connection. Given the graph with the asserted triples:
```:s :p :o.```
```:r rdf:reifies <<(:s :p :o)>>.```
then in each model ```I``` of the graph the denotation of the reifier ```:r``` is connected via the denotation of ```:p``` to the corresponding fact in that model, which is an element of ```F(I)```. Recall that facts in ```F(I)``` are exactly the propositions associated to each asserted triple in the graph, namely the propositions which are true in the model.

On this ground, I can not accept this PR.

**Comment by rat10** [comment]
@franconi 

> I disagree totally - as I told you several times.

I find your tone a bit harsh. I repeatedly asked for clarification but from you I only got "The specs are clear enough".

> > This document advises to understand an assertion on a reified triple as an assertion on an asserted triple of the same form, if such a triple occurs in a graph as both a reified triple and an asserted triple. However, this understanding is not backed by the model-theoretic semantics as layed out in RDF 1.2 Semantics.
> 
> As I already replied to you at the last meeting,

Similarly both my memory and the transcript don't mention any explanation.

> there is a precise and intuitive formal connection. Given the graph with the asserted triples: `:s :p :o.` `:r rdf:reifies <<(:s :p :o)>>.` then in each model `I` of the graph the denotation of the reifier `:r` is connected via the denotation of `rdf:reifies` to the corresponding fact in that model, which is an element of `F(I)`. Recall that facts in `F(I)` are exactly the propositions associated to each asserted triple in the graph, namely the propositions which are true in the model.
> 
> On this ground, I can not accept this PR.

Thank you for this explanation, and I agree that it contradicts what my PR says, and I'm fine with that. I really wanted to get clarification on what the semantics means (since I'm not semanticist enough to interpret with certainty what it says).

Would you be willing to turn what you said above into a PR for addition to the semantics document, or should I give it a try? I think it should be part of RDF 1.2 Semantics rather than Concepts because it is concerned with the model-theoretic foundation of that connection.

**Comment by pfps** [comment]
This is not suitable for RDF Semantics either as it is not about semantics.

**Comment by lisp** [comment]
franconi's remarks are very much about semantics.

**Comment by TallTed** [comment]
All — We need to figure this out. We're all smart people, even on our dumbest day, and we're talking about some very complex topics, and somehow talking past each other. I don't believe that anyone is intentionally misunderstanding nor misguiding another, but somehow disagreement is happening. I am hopeful that no-one in the WG will Formally Object to any of the documents when we reach CR or better, and moreso that we will reach actual consensus by transition time.

To that end: it seems inevitable that multiple people outside of our group will read things in the same way as @rat10 (et al), and others as @franconi (et al). I feel that it is incumbent upon all of us to try to read and understand all of our documents, and especially any documents where conflicting opinions persist about their content. This is the only way to arrive at the interoperability of the powerful model we're specifying.

> I repeatedly asked for clarification but from you I only got "The specs are clear enough".

Let us take it as read that the above is a paraphrase of what has been repeatedly said by multiple people to multiple people. I have witnessed such. Repeatedly, with varying actors reading the simple script, in different directions.

To be blut: If someone recognizes that they don't understand something in the documents we've written, and asks an author or editor for clarification or further explanation of any kind — this is an absolute statement that _that spec is **not** clear enough_, and the person(s) being asked for clarification *must* work to bring the questioner to clarity, and then to edit the text of the specs such that the same clarity is delivered to other readers, without their needing to ask for more — because at *that* point the specs will indeed be "clear enough".

So: If @rat10's or @franconi's (or anyone else's) understanding of what we've written appears to differ from your understanding of what's written — please try to find where what we've written might be adjusted to bring their understanding to match yours (or vice versa). This *must be a group effort,* as the disagreements are longstanding, and despite many iterations over the text, they persist, with similar people on either side.

**Comment by franconi** [comment]
> > as reification is the principle change introduced to rdf by 1.2, unless one intends to introduce reification on the order of an entailment regime, its semantics should be described where everything else is described.
> 
> Triple terms are the principle change introduced, not reification. Reification via rdf:reifies is only a mechanism to help prevent users from making the same mistake as was made in the seminal example. Perhaps that is what is needed to be described in Concepts or, probably better, in a Primer or What's New document.

See my comment in PR #144 https://github.com/w3c/rdf-semantics/pull/144

**Comment by niklasl** [comment]
Since there seems to be agreement here that the semantics does specify what is required (that the facts of an interpretation are the propositions which are true in it), I hope there will be less disagreement (if any) on the way triple terms and reification is presented in Concepts. I also really hope that the work now done through #214 will clarify things in general in that section, for as many as possible.

**Comment by rat10** [comment]
> Since there seems to be agreement here that the semantics does specify what is required (that the facts of an interpretation are the propositions which are true in it),

I still can't see how that is all that is required. My question is: how is it established in model-theoretic terms that an annotation on a reifier annotates an asserted triple in the same graph (if the reified triple term and the asserted triple represent the same triple)? And maybe, probably, that is a question best answered not in the Concepts but the Semantics spec, but that's another issue.

> I hope there will be less disagreement (if any) on the way triple terms and reification is presented in Concepts. I also really hope that the work now done through #214 will clarify things in general in that section,

[#214](https://github.com/w3c/rdf-concepts/pull/214) is definitely a good step forward, and to me it seems quite acceptable now! 

> for as many as possible.

However, it doesn't answer the question of how exactly, i.e. in model-theoretic terms, the connection between a reifier of a triple term and an asserted triple of the same form is established. I have now two descriptions, one by @franconi [above](https://github.com/w3c/rdf-concepts/pull/220#issuecomment-3084097269), 

>> Given the graph with the asserted triples:
>> 
>> `:s :p :o.`
>> `:r rdf:reifies <<(:s :p :o)>>.`
>> 
>> then in each model `I` of the graph the denotation of the reifier `:r` is connected via the denotation of `rdf:reifies` to the corresponding fact in that model, which is an element of `F(I)`. Recall that facts in `F(I)` are exactly the propositions associated to each asserted triple in the graph, namely the propositions which are true in the model.

one by @pchampin in a [comment](https://github.com/w3c/rdf-star-wg/issues/169#issuecomment-3057364057) to WG issue #169 (scroll down to "but there is"), 

>> A [proposition](https://www.w3.org/TR/rdf12-semantics/#dfn-set-of-propositions) is an element of the set { RE(x, y, z)｜x is in IR, y is in IP, z is in IR } . Such propositions are [facts](https://www.w3.org/TR/rdf12-semantics/#dfn-set-of-facts) (i.e. true in the interpretation) if furthermore <x, z> is in IEXT(y) .
>>
>> As for the (thing denoted by the) reifier, it is trivially connected to the proposition (as part of IEXTrdf:reifies, per the definition of a reifier in RDF Concepts). Et voila!
>>

They seem to say the same thing, but I'm not convinced of the construction they describe (ie refer to), because in a way that connection between (reified) triple term and asserted triple still seems merely coincidental. Does that have the necessary semantic bite to establish a semantic connection between annotation and asserted triple with enough model-theoretic weight? I'm not sure, and that will take some more thinking on my side. Or maybe someone else can chime in.

And then there's the discussion if these clarifying comments can be reconciled into an additional explaining sentence or paragraph (to help those that are as lost as I am in this respect), and where such a sentence or paragraph should go (I lean towards the semantics specification, but we'll see).

**Comment by pchampin** [comment]
@rat10, your previous comment contains two questions:

> how is it established in model-theoretic terms that an annotation on a reifier annotates an asserted triple in the same graph 

> how exactly, i.e. in model-theoretic terms, the connection between a reifier of a triple term and an asserted triple of the same form is established

you seem to consider that they are two wordings of the same question. In the following, I will try to explain that they are not, that the response you got previously from @franconi and myself respond to the second one, and that the response to the first one is different (which might explain why you get the impression that we are contradicting each other or ourselves, and why other complain about your questions being "moving targets"). Apologies for a long comment.

---

The 2nd question is relatively general: it asks about "the connection" between the reifier and an asserted triple (or more precisely, between their denotee). The connection exists in IEXT(I(rdf:reifiers)).

Let me also point out something in the wording of question 2 that may explain why we keep misunderstanding each other: "an asserted triple of the same form [as the triple term]". Note that those things (the asserted triple and the triple term) are not two distinct entities that coincidentally have the same form: they are the **same** thing (an [RDF triple](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-triple)) used in two different ways (once as a [term](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-term), once as the [element of the graph](https://www.w3.org/TR/rdf12-concepts/#dfn-asserted-triple)).

In a sense, distinguishing them is similar to distinguishing a subject IRI and an object IRI "of the same form". This distinction is irrelevant in RDF Semantics, because they are the same IRI, and so by definition always denote the same thing -- and so anything connected to "one" is connected to "the other".

---

The 1st question is unclear: the terms "annotation" and "annotates" are not currently defined in RDF-Concepts nor in RDF-Semantics (although a definition of "triple annotation" is given in #214). I don't mean to be pedantic, but you will not get precise answers to your questions until your questions are precise enough...

I assume that by "annotation on a reifier", you mean any triple (other than the reifying triple) that use a reifier as its subject (and possibly its object?).

I also assume that by "an asserted triple in the same graph" you mean "that is the same ("has the same form") as the triple term related to the reifier".

Then what you mean by "annotates" is also unclear. Let me take an example. Given the following graph
```
_:r rdf:reifies <<( :s :p :o )>> . # introduce the reifier
_:r :a :b . # the annotation on the reifier
:s :p :o .  # the reified triple is also asserted
```
what would it mean that "the annotation on the reifier annotates the asserted triple"?
If I had to guess, I would go for
```
<<( :s :p :o )>> :a :b . # in generalized RDF
```
but another (more literal) interpretation would be something like
```
<<( _:r :a :b )>> :annotates <<( :s :p :o )>>. # still in generalized RDF
```
Either way, the answer is no: RDF semantics does not provision for any of those inference...

---

As a conclusion, I notice that those questions relate to the semantics, not RDF-Concept, so they rather belong to w3c/rdf-semantics#144, and I support @niklasl 's [suggestion above](https://github.com/w3c/rdf-concepts/pull/220#issuecomment-3085927297).

**Comment by rat10** [comment]
Technical comment: I updated the PR according to TallTed's proposal. As in the meantime the spec was updated I had to resolve a conflict, and commit my updates to the "main branch". No guarantees that I did everything according to best practices. Apologies upfront if I messed things up!

**Comment by pfps** [comment]
Suppose that one actually wanted to make a closer connection between truth in an interpretation and triple terms in Semantics.  The way to go would be to define the set of triples in an interpretation as the range (the mathematical range not the RDFS range) of RE and define IEXT as a subset of the set of triples.  Then an asserted triple reifier of a triple term, x y z, in an interpretation is just a domain element, a, for which RE(a,IS(rdf:reifies),t) is in IEXT and t is in IEXT, for t = RE(I(x),I(y),I(z))).

But all this can already be done, simply by stating the above in terms of the current semantics somewhere (maybe in Semantics, maybe in Concepts, maybe in some other place).  So Semantics already supports the notion of annotations of facts in an interpretation.  

In fact, Semantics already includes part of this in its definition of the set of facts in an interpretation.  So an asserted triple reifier in an interpretation is just the first element of a fact whose predicate is the denotation of rdf:reifies and whose object is also a fact.

**Review by franconi — CHANGES_REQUESTED** [review]
As I said, my opinion is that no change is required in Concepts (possibly fine-tuning #214) or Semantics (even if I will propose an addition in order to explain rdf:reifies as rdf:type is already explained - PR is coming). A forthcoming FAQ/note document based on examples should be the way out.

**Comment by franconi** [comment]
To strengthen the argument by @pchampin, let me rephrase his example using standard vocabularies:
```
@PREFIX prov: <http://www.w3.org/ns/prov#> .
@PREFIX dc: http://purl.org/dc/elements/1.1/>.

<<( :s :p :o ~ :r1 )>> a rdf:Statement; 
  dc:creator :alice ; 
  prov:invalidatedAtTime "2024-09-02T01:31:00Z"^^xsd:dateTime ;
  prov:PrimarySource <another_file.ttl> .
```

**Comment by rat10** [comment]
@pfps said:
> Suppose that one actually wanted to make a closer connection between truth in an interpretation and triple terms in Semantics. 

IMO that would be even mandatory to fully carry out the WG's charter.

> The way to go would be to define the set of triples in an interpretation as the range (the mathematical range not the RDFS range) of RE and define IEXT as a subset of the set of triples. Then an asserted triple reifier of a triple term, x y z, in an interpretation is just a domain element, a, for which RE(a,IS(rdf:reifies),t) is in IEXT and t is in IEXT, for t = RE(I(x),I(y),I(z))).

If I interpret this correctly it means that a reification of a triple term (vulgo: a reifier) is defined as entailing the triple _iff_ that triple is also asserted. That sounds like a clever approach to me.

> But all this can already be done, simply by stating the above in terms of the current semantics somewhere (maybe in Semantics, maybe in Concepts, maybe in some other place). So Semantics already supports the notion of annotations of facts in an interpretation.

And it's good to hear that it can "already be done" - which I take as synonymous to "can be added to the model-theoretic semantics without much extra effort and without breaking anything we already have".

> In fact, Semantics already includes part of this in its definition of the set of facts in an interpretation. So an asserted triple reifier in an interpretation is just the first element of a fact whose predicate is the denotation of rdf:reifies and whose object is also a fact.

Yes, but only _part of this_  ;-) If someone could draft a PR to add the missing part, that would be great.

<hr>


@franconi 
> As I said, my opinion is that no change is required in Concepts (possibly fine-tuning #214) or Semantics (even if I will propose an addition in order to explain rdf:reifies as rdf:type is already explained - PR is coming). A forthcoming FAQ/note document based on examples should be the way out.

Why not try to draft the addition that @pfps suggests? A solid, model-theoretic definition of annotations on facts ("statements about statements") would benefit everybody. The forthcoming FAQ/note document based on examples will still be much needed to instruct users on the basics and subtleties of this mechanism.  

<hr>


@pchampin
> @rat10
> 
> > @pchampin, thank you for the clarifications.
> > **tl;dr** It seems clear by now that the proposed reification mechanism does not provide for "statements about statements", as the WG charter asks for, but rather for something like "statements about (abstract) propositions".
> 
> I lack time to respond to all the points, 

I understand, but maybe later?

> but I strongly disagree with this first point. Reifiers can describe many different things, including statements. The example below contains a statement about a statement.
>
> ```
> <<( :s :p :o ~ :r1 )>> a rdf:Statement; dc:creator :alice; :source <another_file.ttl> .
> ```

The question here is not the interpretation but how to address the asserted triple in the first place. An annotation can be about the asserted triple itself (e.g. source, date of assertion, etc) or about its interpretation (adding detail to what it describes). But before such differentiation can be usefully made (and, notwithstanding the well known identity crisis argument, that is mostly the job of vocabularies) one has to be sure what one is talking about: an asserted triple, or a reified (but still abstract) triple. That is the core of the issue at hand.

> From previous conversation, I suspect that this example will not satisfy you, as you are rather interested about "statements about statements _from the same graph_", but as you insist on the letter of the charter, I would argue that my example above does satisfies the charter.

No, I don't insist on that. It's quite hard to avoid your criticism of not being specific enough, and then also not to be overly specific.

> Furthermore, the example can be easily adapted to talk about statements from the same graph, for example:
> 
> ```
> :s :p :o ~ :r2 {| a rdf:Statement; dc:creator :pchampin; :source <> |}.
> ```
> 
> (using the relative IRI `<>` to indicate that I'm talking about a statement in the very RDF document we are looking at).

Agreed, and indeed useful in some situations, but not what this issue/PR is about
 
> What I suggest to do with an application-specific predicate (`:source`), it seems that you would like to define it in RDF semantics itself.

No, by no means. Neither would I nor is it the topic of this issue/PR.

> [...]


As an aside, your usage of `rdf:Statement` is problematic since those are clearly defined to **not** refer to asserted triples. If we change that, we change all existing uses of RDF 1.0/1.1 reification. Or is your example intended to illustrate exactly that case?

<hr>

@pfps 
> Either "statement" in the charter means instances of rdf:Statement, or is it some under-defined term that is up to the working group to define. 

Neither, IMO. We agree that it can't possibly be meant to refer to `rdf:Statement` which specifically types an RDF 1.0/1.1 reification quad. But the term "statement" is defined in [RDF Concepts](https://www.w3.org/TR/rdf11-concepts/#resources-and-statements) as follows: 
"Asserting an [RDF triple](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-triple) says that some relationship, indicated by the [predicate](https://www.w3.org/TR/rdf11-concepts/#dfn-predicate), holds between the [resources](https://www.w3.org/TR/rdf11-concepts/#dfn-resource) [denoted](https://www.w3.org/TR/rdf11-concepts/#dfn-denote) by the [subject](https://www.w3.org/TR/rdf11-concepts/#dfn-subject) and [object](https://www.w3.org/TR/rdf11-concepts/#dfn-object). This statement corresponding to an RDF triple is known as an RDF statement." 
We can assume that the charter left out the "RDF" part for the sake of conciseness and readability. The charter is also not concerned with specific syntaxes. Therefore the charter's "statements about statements" has to be understood as referring to asserted triples whenever it uses the term "statement".

**Comment by afs** [comment]
"statements" in RDF Concepts 

[1.2 Resources and Statements](https://www.w3.org/TR/rdf12-concepts/#resources-and-statements)

"This statement corresponding to an RDF triple is known as an RDF statement. "

**Comment by pchampin** [comment]
@rat10
> If I interpret this correctly it means that a reification of a triple term (vulgo: a reifier) is defined as entailing the triple

A reifier is a term. It can not entail anything: entailment is a relationship between graphs -- and therefore, no, that's not what @ofps is saying.

My understanding of what @pfps is writing [above](https://github.com/w3c/rdf-concepts/pull/220#issuecomment-3113421890) is: an "asserted triple reifier" is a reifier that reifies a triple that happens to be asserted. This is the point I've been making earlier.

> As an aside, your usage of rdf:Statement is problematic since those are clearly defined to not refer to asserted triples.

That's simply not true. It is perfectly OK to put in the same graph a triple `:s :p :o.` and the reification quad `_:b a rdf:Statement; rdf:subject :s ; rdf:predicate :p; rdf:object :o .`  Again, `_:b` is about the triple `_:s _:p _:o` and `_:s _:p _o` happens to be asserted, so `_:b` is about an asserted triple.
(I use the term "is about" because `_:b` does not refer to the triple itself, it refers to  a "token of the triple", per RDF Semantics D.1, but I believe this is orthogonal to the question).

> No, I don't insist on that [the letter for the charter]

Prior to my comment, you had written 5 times in this thread "statements about statements" in quote, citing the charter. I would consider this as insisting on the letter of the charter. Which, by the way, is your right entirely -- I didn't mean that pejoratively, I just thought it called for an answer -- even though, as you point out, we are getting side-tracked from the topic of this PR.

The PR contains the text:

 > an assertion on a reified triple is an assertion on an asserted triple of the same form, if such a triple occurs in a graph as both a reified triple a and an asserted triple.

if a triple T "occurs in a graph as both a reified triple and an asserted triple, then any assertion on T (a refied triple) is also an assertion on T (an asserted triple)! This is a tautology, and therefore claiming that

> this understanding is not backed by the model-theoretic semantics"

is plain wrong.

**Comment by TallTed** [comment]
@pchampin — Please edit https://github.com/w3c/rdf-concepts/pull/220#issuecomment-3119625467 and correct the typo of `@ofps` to `@pfps`. They will keep getting unsolicited notifications about this thread, but hopefully it will end soon, and no copy-pastes of that comment will non-consensually re-engage that user.

**Comment by Antoine-Zimmermann** [comment]
@pfps,

This piece was discussed yesterday in the call:

> Suppose that one actually wanted to make a closer connection between truth in an interpretation and triple terms in Semantics. The way to go would be to define the set of triples in an interpretation as the range (the mathematical range not the RDFS range) of RE and define IEXT as a subset of the set of triples. Then an asserted triple reifier of a triple term, x y z, in an interpretation is just a domain element, a, for which RE(a,IS(rdf:reifies),t) is in IEXT and t is in IEXT, for t = RE(I(x),I(y),I(z))).
> 
> But all this can already be done, simply by stating the above in terms of the current semantics somewhere (maybe in Semantics, maybe in Concepts, maybe in some other place). So Semantics already supports the notion of annotations of facts in an interpretation.
> 
> In fact, Semantics already includes part of this in its definition of the set of facts in an interpretation. So an asserted triple reifier in an interpretation is just the first element of a fact whose predicate is the denotation of rdf:reifies and whose object is also a fact.
> 
> EDIT: I believe that the right place to discuss assertions on facts is in Schema, as this is part of the intended meaning of rdf:reifies, and in tutorial material, to show how rdf:reifies is supposed to be used.

and this made me think a lot, because it misled me very much. At first, I thought you were bringing a formal argument, so that we had to interpret all this as a sketched mathematical proof of a claim. But interpreting what you say here as a formal proof requires non trivial extrapolation. Yet, @franconi claimed yesterday that this is all trivial.

After putting much thought into it, I believe that you are not particularly trying to make a formal point, but with all due respect, your post is clumsy because it misleads very much. At first, you introduce some formalism suggesting that what @rat10 requests could be done if only we changed the formal semantics. Immediately after, you say that the same thing is doable without any change; yet you only keep the example of how to do it in the revised semantics, without explaining how to do it in the actual RDF 1.2 semantics.

My understanding now of what you are saying is that one can always make statements about an asserted statement by 1) asserting the statement (i.e., adding an RDF triple to the graph that asserts it); 2) assuming that the statement exists in the universe of discourge (i.e., the domain of interpretation IR) and having a syntactic element that's intended to denote it (whether it is a URI, or a triple term mapped to it via the function RE); 3) if using a triple term, make the connection a little tigher by making the triple term exactly the triple that asserts the statement.

E.g., assume we want to state something about the fact that Tim Berners-Lee invented the Web. Let us name $s$ this statement. Then, write a triple that denotes this statement:

    w3:timbl onto:invented w3:www .

By writing this, we assume that the interpretation contains Tim Berners-Lee, contains the Web, contains the relation of inventing something, even though the formal semantics does not enforce in any way that this triple means what we intend it to mean. So we continue:

    <r> rdf:reifier << w3:timbl onto:invented w3:www >> .

Again, nothing formally connects this to the statement, but we can assume that our intention is that RE(I(w3:timbl),I(onto:invented),I(w3:www)) = $s$. Or, if we wanted to make statements about the RDF triple itself, rather than the statement it denotes, we could assume that RE(I(w3:timbl),I(onto:invented),I(w3:www)) = (w3:timbl, onto:invented, w3:www).

It is indeed obvious that we can do this with the current semantics, but we could make the same point with only RDF 1.1. We can mint a URI intended to denote $s$, assert the triple `w3:timbl onto:invented w3:www`, then make the connection tighter by relating the URI to the triple using syntax like:

   <stmt> ex:isTriple "w3:timbl onto:invented w3:www"^^dt:turtle .

and assume that `ex:isTriple` is intended to denote a relation between a statement and a string that serialises an RDF triple that denotes the statement. 

The problem that I see in this is that if "all this can already be done" in RDF 1.1, then why introduce the machinery of triple terms, and stop introducing more that brings something that "cannot already be done"? Wouldn't it be equally satisfying to interpret triple terms as syntactic sugar for RDF reification?

**Comment by pfps** [comment]
Well, yes, in some strong sense RDF 1.1 can already do everything because it has triples.  Even more so, it has vocabulary for statements.  But somehow that vocabulary has been deemed insufficient.

Further, I have found all the discussion around this issue very confusing.  I have never been able to figure out either what the problem is or what the solution should be.  I've even still confused as to whether the desired solution to the problem is supposed to be some fundamental change to the semantics, some ancillary change to the semantics, or just some explanatory text.  So I guess it is not surprising that my comments might be considered confusing.

My goal in the referenced post was two-fold.  First, to describe how the 1.2 semantics might be changed in a way that might solve what might be the problem.  (There are lots of ways that this might be done and I only described one of these.)  Second, to state that the current 1.2 semantics already has machinery that does essentially the same as the semantics change.  I didn't make this a formal argument.  So why would I do this?  My goal was to show that the current 1.2 semantics gets you to essentially the same place as you would get to with the potentially more-explicit machinery.  


The post above appears to be complaining that there is no, or maybe insufficient, connection between:

:a :b :c .

and

<<( :a :b :c )>>

It is certainly true that one might want to have a logical formalism where there is closer connection.  But what, formally, is this connection supposed to be?  And without adding A LOT to RDF?

One might also want to have more description about how triple terms are to be used to make "annotations on statements".  But what?  And what part of this needs to be in Semantics, a document that is not likely to be read by RDF users who just want to do things like:

<< :NotreDame :reconstructed "2024"^^xsd:gYear >> :accordingTo :NYT .

**Comment by pfps** [comment]
I've never been of the opinion that the task of the WG was anything more than adding triple terms (quoted triples) to RDF.  As an aside, this doesn't even cover rdf:reifies so the WG is over-delivering.  It is my view that this machinery is adequate to support (in RDF style) all that is covered in the seminal paper, the CG final report, the WG charter, and everything in the current WG documents.

Is there more that could be done in this area?  Absolutely!  Should it be part of RDF?  For almost all of it, absolutely not!  For any remaining bits, there is now (and, by the way, there already was in RDF 1.1) adequate machinery to do so, but these bits are intended meaning and custom and practice and so not part of the RDF 1.2 semantics.

**Comment by pfps** [comment]
Ooops, how do I revert my inadvertent edit to domel's post?

**Comment by pfps** [comment]
>     1. Do you agree that a triple being asserted (i.e. a member of the graph set) and the same triple used as a term (e.g. as the object of a reifying triple) are the same triple? Or do you think otherwise, for instance that the two uses of identical triples in different roles somehow should turn that triple into two different (yet identical) triples?
>        Given the current definitions, it is the same triple, denoting the same proposition, one used for assertion (assigning truth value), one for reference (reification).

My belief is that the whole idea behind quoted triples (triple terms), is that there is only one quoted triple (triple term) for any triple, and that the quoted triple (triple term) is the triple, just in a different context (part of another triple instead of directly an element of the graph).

**Comment by afs** [comment]
> Ooops, how do I revert my inadvertent edit to domel's post?

You can copy versions out of the edit history.

**Comment by pchampin** [comment]
@rat10 

> * Now **what I would like to have** is an addition to RDF Semantics that formalizes that co-occurrence of reified hypothetical triple and asserted triple is to be understood as entailment.
>       E.g. _if_ a graph contains both `s p o.` and `r rdf:reifies <<s p o>>.` _then_ `<<s p o>>` can be understood to _entail_ `s p o.`.

Assuming that `<<s p o>>` above is in fact `<<(s p o)>>` (i.e. the triple term), I would say that you already have *almost* what you ask above:
* if a graph contains `s p o`, then it entails `s p o` (trivially)
* if, additionally, the graph contains a reifying triple about `<<(s p o )>>`, then the graph still entails `s p o`

You almost *have it*, because under the conditions you put, `s p o` is indeed entailed.
You *almost* have it, because you can not really say that "`<<(s p o)>>` can be understood to entail `s p o`" or, in other words, that the presence of the triple term `<<(s p o)>>` in the graph is the reason for `s p o` to be entailed by that graph.

Why can you not say that? First, because if we take out of the graph all mentions of `<<(s p o)>>` (and only that), then the new graph still entails `s p o`. Second, because if we take out the asserted triple `s p o` and keep the mentions of the triple term, the new graph does not (necessarily) entail `s p o` (but that should not bother you, as the presence of `s p o` is a pre-condition in your own proposal).

And again, if `s p o` is supposed to be in the graph, then why should we pretend that the graph entails `s p o` for another reason? And what would that bring us? How would it change, if at all, the way entailment currently work?

----

@lisp 

> i am less concerned with assigning a "truth value to a proposition" than with associating annotations with an asserted triple.

Given that an asserted triple is nothing more than a triple that happens to be asserted (see the [definition](https://pr-preview.s3.amazonaws.com/rat10/rdf-concepts/pull/220.html#dfn-asserted-triple)), then what we have totally allows you to associate annotations with assereted triples: make an assertion A about a triple T; assert T; boom, A is now about an asserted triple.

> that is i would value an rdf for which the abstract syntax is extended to permit (...)

in your example, you are talking about entailments in named graphs... but named graphs are not, and never were, covered by the semantics (nor is it in our charter to define a semantics for named graphs). So how is this relevant to the current discussion?

**Comment by pchampin** [comment]
> It is no problem (and that it a feature, not a bug) if two triples independently from each other entail the same third triple, and if that third triple is already in the graph.

I totally agree, but the situation you want is very different: 
* "two triples independently from each other entail the same third triple" → A entails C, B entails C
* what you want above → the co-occurrence of A and B entails A

> It seems like you don't have strong arguments against my proposal,

Let me rephrase:
* In `:c1 rdfs:subClassOf :c2. :c2 rdfs:subClassOf :c3.`, I agree that the co-occurrence of the two triples entails `:c1 rdfs:subClassOf :c3.`, because
  - any graph containing the two initial triples always entails that conclusion, and
  - if you remove any of the two initial triples, the conclusion does not (always) hold.

* In `:s :p :o. _:r rdf:reifies <<(:s :p :o)>>.`, I disagree that it is the co-occurrence of the to triples that entails `:s :p :o.`:
  - granted, any graph containing the two always entails that conclusion, but
  - if you remove the second triple, then the conclusion still holds, so the co-occurence is not required.

* Would you claim that in `:s :p :o. :a :b :c.`, the co-occurence of the two triples is the reason that `:s :p :o.` holds? I suppose not; yet, `:a :b :c.` has just as much influence on that entailment that `_:r rdf:reifies <<(:s :p :o)>>.` in the previous example. 



> * To me this seems like a semantically more solid link between reified triple and asserted triple. Entailment is _the_ way to solidly define a dependency between two triples in RDF. 

That's the thing: there is no dependency between the triples: any of them can be true regardless of the other one being true or false!

> Anything else, e.g. the current semantics, is just an approximation and _can_ be interpreted in different ways. The prose tries to lead into another direction, but the normative force of prose is limited.

Can you point to what part of the prose leads to believe that there is a dependency between a reifying triple and an assertion of the triple being reified?

> * It keeps the semantics where it should be: with the property (in this case `rdf:reifies`). IIUC the current semantics bakes so much meaning into the triple term itself that it is impossible to define properties with weaker semantics, e.g. properties that **reference the triple term irrespective of and explicitly agnostic to the corresponding triple actually being asserted**.

Yet, that (the part I emphasized) is exactly what rdf:reifies is doing...

**Comment by pchampin** [comment]
> The specs claim that an annotation on a reifier referencing a triple term `<<(spo)>>` also annotates an asserted triple `spo.`. And that connection is only claimed in prose, but not defined formally in model-theoretic semantics.

Indeed, the spec defines a [triple annotation](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-annotation) as a set of triples whose subject is a common reifier. This notion only defined at the syntax level, as a generalization of the annotation syntax provided by Turtle.

I don't see any semantic implication in that term (the term "annotation" is pretty genetic in my opinion) nor in the prose around it.

> > > * It keeps the semantics where it should be: with the property (in this case `rdf:reifies`). IIUC the current semantics bakes so much meaning into the triple term itself that it is impossible to define properties with weaker semantics, e.g. properties that **reference the triple term irrespective of and explicitly agnostic to the corresponding triple actually being asserted**.
> > 
> > 
> > Yet, that (the part I emphasized) is exactly what rdf:reifies is doing...
> 
> Yes, but the definition of triple terms in RDF Concepts does more, and that's the problem I'm describing. It overrides weaker semantics in any property referring to them. That is an overreach of RDF Concepts, in itself unnecessary (the same effect could be achieved with a more explicit definition of `rdf:reifies`) and blocking other approaches (like a duality of e.g. `rdfs:states` and `rdf:mentions`). RDF Concepts should restrain itself to define triple terms as references to propositions. That's all it needs to do.

... and that's all it does. Again, what part of RDF Concept do you consider "overrides" the semantic of any property referring to triple terms / propositions?

**Comment by rat10** [comment]
@pchampin I have to say that I have the feeling that you are trolling me.

> > The specs claim that an annotation on a reifier referencing a triple term `<<(spo)>>` also annotates an asserted triple `spo.`. And that connection is only claimed in prose, but not defined formally in model-theoretic semantics.
> 
> Indeed, the spec defines a [triple annotation](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-annotation) as a set of triples whose subject is a common reifier. This notion only defined at the syntax level, as a generalization of the annotation syntax provided by Turtle.
> 
> I don't see any semantic implication in that term (the term "annotation" is pretty genetic in my opinion) nor in the prose around it.

From when on is a term solidly enough specified to define a semantic implication in your opinion?

> > > > * It keeps the semantics where it should be: with the property (in this case `rdf:reifies`). IIUC the current semantics bakes so much meaning into the triple term itself that it is impossible to define properties with weaker semantics, e.g. properties that **reference the triple term irrespective of and explicitly agnostic to the corresponding triple actually being asserted**.
> > > 
> > > 
> > > Yet, that (the part I emphasized) is exactly what rdf:reifies is doing...
> > 
> > 
> > Yes, but the definition of triple terms in RDF Concepts does more, and that's the problem I'm describing. It overrides weaker semantics in any property referring to them. That is an overreach of RDF Concepts, in itself unnecessary (the same effect could be achieved with a more explicit definition of `rdf:reifies`) and blocking other approaches (like a duality of e.g. `rdfs:states` and `rdf:mentions`). RDF Concepts should restrain itself to define triple terms as references to propositions. That's all it needs to do.
> 
> ... and that's all it does. Again, what part of RDF Concept do you consider "overrides" the semantic of any property referring to triple terms / propositions?

So it is not specified how to annotate asserted triples? Not even with operational semantics?

**Comment by pchampin** [comment]
> @pchampin I have to say that I have the feeling that you are trolling me.

I apologize for that, that is not my intention.

> > > The specs claim that an annotation on a reifier referencing a triple term `<<(spo)>>` also annotates an asserted triple `spo.`. And that connection is only claimed in prose, but not defined formally in model-theoretic semantics.
> > 
> > 
> > Indeed, the spec defines a [triple annotation](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-annotation) as the subset of triples whose subject is a common reifier. This notion only defined at the syntax level, as a generalization of the annotation syntax provided by Turtle.
> > I don't see any semantic implication in that term (the term "annotation" is pretty genetic in my opinion) nor in the prose around it.
> 
> From when on is a term solidly enough specified to define a semantic implication in your opinion?

Let me clarify what I meant. It seems that the term "annotation" has some connotations for you that imply some semantics consequences, and therefore, you (understandably) would like to see them formally reflected in RDF semantics. When I wrote "I don't see any semantic implication", I meant that I didn't personally share this perception of the term "annotation". I'm therefore happy to see it used in RDF Concepts only (like other terms, such as "reifying triple", "simple literal" or" blank node identifier"...). I'm also under the impression that the majority of WG participants have the same perception.

> > > > > * It keeps the semantics where it should be: with the property (in this case `rdf:reifies`). IIUC the current semantics bakes so much meaning into the triple term itself that it is impossible to define properties with weaker semantics, e.g. properties that **reference the triple term irrespective of and explicitly agnostic to the corresponding triple actually being asserted**.
> > > > 
> > > > 
> > > > Yet, that (the part I emphasized) is exactly what rdf:reifies is doing...
> > > 
> > > 
> > > Yes, but the definition of triple terms in RDF Concepts does more, and that's the problem I'm describing. It overrides weaker semantics in any property referring to them. That is an overreach of RDF Concepts, in itself unnecessary (the same effect could be achieved with a more explicit definition of `rdf:reifies`) and blocking other approaches (like a duality of e.g. `rdfs:states` and `rdf:mentions`). RDF Concepts should restrain itself to define triple terms as references to propositions. That's all it needs to do.
> > 
> > 
> > ... and that's all it does. Again, what part of RDF Concept do you consider "overrides" the semantic of any property referring to triple terms / propositions?

My question still holds...

> So it is not specified how to annotate asserted triples? Not even with operational semantics?

"Triple annotation" is defined . Arguably, one could conjugate that as "annotate a triple", and by definition, that triple is asserted. So yes, RDF Concepts specifies how to annotate asserted triples. True, this triple annotation ("a subset of triples about a common reifier") receives no special treatment in RDF semantics, but its semantics is defined just as well as the semantics of any other subset of triples...

If you take the example of triple annotation in [Fig 3](https://www.w3.org/TR/rdf12-concepts/#fig-asserted-triple-term) of RDF Concepts: the triple annotation is made of 2 triples. RDF Semantics tells us that:

* the proposition π denoted by `:Alice :name "Alice".` is related to the thing ρ denoted by `:rei-1` (via the property denoted by `rdf:reifies`)
* the thing β denoted by `:Bob` is related to the thing ρ denoted by `:rei-1` (via the property denoted by `:accordingTo`)
* therefore,  there is a "path" between the proposition π and the thing β, and that path goes through thing ρ.

That's basically what you can say about the semantics of annotations, just applying the text of RDF Concepts and RDF Semantics as it is. And at no point did I do anything special because one of the terms denoted a proposition rather than an arbitrary resource, nor because one of the properties was I(`rdf:reifies`). This is why I (and others) do not see value in adding text to describe this as a special case that it is not.

**Comment by pfps** [comment]
The title of this issue still looks as if there is a need to change the semantics of RDF.  So until this issue is closed or amended to indicate that there is no effect on Semantics the final checks on Semantics should not be done.

**Comment by pfps** [comment]
A Modest Proposal to solve all the problems with rdf:reifies, reification, and so on.

Given all the perceived problems surrounding reifiers and rdf:reifies I propose that the WG vote to remove rdf:reifies from RDF 1.2 and remove all discussion of reification and related concepts.  This will open up triple terms to any sort of exploitation by users of RDF 1.2 without any restrictions or guardrails coming from the working group.

I further propose that the vote be a ranked-choice vote between:
1/ Keeping rdf:reifies and reification as they are.
2/ Removing rdf:reifies and reification.
3/ Making changes to reification.

**Comment by niklasl** [comment]
Apart from undoing at least one year of our work&mdash;in Concepts, Semantics, SPARQL, most if not all all syntaxes, several votes, including a formal vote and a chair decision&mdash;that still wouldn't solve this particular issue, since use of triples as terms is not assertion.

Granted, it might reveal, as some previous discussions and votes have perhaps indicated, if there is a wider inability and/or unwillingness to distinguish statements (tokens with date, source, etc., which can be indirectly reified in RDF 1.1) from propositions (denoted by triples in RDF interpretations, which in current 1.2 can be used in a direct form of reification). And if this is actually such a big conceptual problem, the even more modest thing to do would be to undo the introduction of this distinction. That would undo some more of the work.

File: spec/index.html [hunk 1]
```diff
@@ -395,6 +395,26 @@ <h3>Triple Terms and Reification</h3>
         statement that contradicts another statement, whether asserted or not.</p>
     </div>
 
+    <p>Note that a <a>triple term</a> may also have another <a>triple term</a> as an <a>object</a>.</p>
```
TallTed, line 398:
```suggestion
    <p>Note that a <a>triple term</a> may itself have another <a>triple term</a> as an <a>object</a>.</p>
```

Antoine-Zimmermann:
> A Modest Proposal to solve all the problems with rdf:reifies, reification, and so on.
> 
> Given all the perceived problems surrounding reifiers and rdf:reifies I propose that the WG vote to remove rdf:reifies from RDF 1.2 and remove all discussion of reification and related concepts. This will open up triple terms to any sort of exploitation by users of RDF 1.2 without any restrictions or guardrails coming from the working group.
> 
> I further propose that the vote be a ranked-choice vote between: 1/ Keeping rdf:reifies and reification as they are. 2/ Removing rdf:reifies and reification. 3/ Making changes to reification.

I very much support this proposal. I'd vote for number 2. Since it was introduced, I've always thought that `rdf:reifies` is pointless and potentially harmful.

niklasl:
> > A Modest Proposal to solve all the problems with rdf:reifies, reification, and so on.
> > Given all the perceived problems surrounding reifiers and rdf:reifies I propose that the WG vote to remove rdf:reifies from RDF 1.2 and remove all discussion of reification and related concepts. This will open up triple terms to any sort of exploitation by users of RDF 1.2 without any restrictions or guardrails coming from the working group.
> > I further propose that the vote be a ranked-choice vote between: 1/ Keeping rdf:reifies and reification as they are. 2/ Removing rdf:reifies and reification. 3/ Making changes to reification.
> 
> I very much support this proposal. I'd vote for number 2. Since it was introduced, I've always thought that `rdf:reifies` is pointless and potentially harmful.

@Antoine-Zimmermann I've come to understand that you have this position, but I haven't yet understood if that also means that you would support uses of triple terms as subjects to denote propositions with dates and/or having sources? (In the same way as 1.1 reification using `rdf:Statement` instances.)

lisp:
> I don't see any restriction or guardrail on triple terms usage coming from the working group related to the presence of rdf:reifies in the spec. Users can exploit triple terms without any restriction or guardrail also in presence of rdf:reifies. What makes you think so?

is there no longer a restriction on placing them in the subject position?

franconi:
> > I don't see any restriction or guardrail on triple terms usage coming from the working group related to the presence of rdf:reifies in the spec. Users can exploit triple terms without any restriction or guardrail also in presence of rdf:reifies. What makes you think so?
> 
> is there no longer a restriction on placing them in the subject position?

The restriction on placing triple terms in the subject position does not depend on the presence of rdf:reifies.

afs:
> A Modest Proposal

Some context:

[A Modest Proposal](https://en.wikipedia.org/wiki/A_Modest_Proposal) is a work of satire.

franconi:
@TallTed, there are a number of serious problems in your encoding:

1. your encoding is lossy: there is no way to understand from the RDF 1.1 representation which are the two triples being reified. The structure and the identity of the two original triples is lost, and there is no SPARQL query which can reconstruct the two original triples. In RDF 1.2 we have introduced `rdf:reifies` exactly to deal with this case: you may remember the discussion and the final vote about the four options a couple of years ago. 
<br>To better see this, I can printout the triples in the 1.1 graph as follows (order of triples doesn't matter in an RDF graph):

 ```
:w2 rdf:type       rdf:statement ; 
     rdf:subject    :liz          ; 
     rdf:predicate  :married      ; 
     rdf:object     :las-vegas    .  
:w2 rdf:type       rdf:statement ;  
     rdf:subject    :richard      ;   
     rdf:predicate  :married-in   ;
     rdf:object     :richard      .
```

2. `:w2` is both a `rdf:statement` and a `:Marriage`, making this resource a semantic horror. In RDF 1.2 we distinguish between propositions (instances of `rdfs:Proposition`) and reifiers.

And, by the way, I don't see any _poor modelling_ in my original example.

franconi:
@Antoine-Zimmermann:

> Since it was introduced, I've always thought that `rdf:reifies` is pointless and potentially harmful.

It is important that you tell us your doubts. Why **pointless** and why **harmful**?
I believe that my example above shows that it is not pointless.

rat10:
I could live with definitions that are more concrete wrt the nature of `rdf:reifies`, and that leave the door open for other uses of triple terms:

- for RDF Semantics to provide the necessary means to formally define the semantics of eventual properties `rdfsx:states` (which _entails_ the triple term) and `rdfx:mentions` (which does not annotate any such _asserted_ triple, even if it exists in the same graph)

- for RDF concepts to clearly specify that it’s not the triple term itself, but the reference to it via `rdf:reifies` which makes an annotation refer to an asserted triple if such a triple exists in the graph (because otherwise there can be no `rdfx:mentions` based on triple terms)

- for RDF Schema to provide a proper definition of purpose, domain and range of `rdf:reifies`, i.e.  what it refers to (triple terms, representing propositions), what that reference means (a reference to an instance/token of such a triple, asserted or not), what it doesn’t mean (any fixing of what can be said about such a reifier, any fixing if the reifier is used to refer to the act of stating or to a specific characteristic of the proposition it refers to, etc)

franconi:
> > I don't believe that you can convert the following example to the old reification:
> > ```
> > << :liz :married :richard  ~  :w2 >> a :Marriage ; :starts 1975 .
> > << :richard :married-in :las-vegas  ~  :w2 >> :best-man :jim-benton .
> > ```
> > You need `rdf:reifies` to reify two different triples to the same reifier.
> 
> I would assume that it's pretty easy to overcome this restriction: just list a few reifications, and annotate the list (and the little uncertainty if the annotation refers to the list _itself_ or _what it lists_ is just standard RDF identification semantics trouble). More triples, but not impossible.

Please do it with my example.
And you should not use any property beyond the `rdf:statement`, `rdf:subject`, `rdf:predicate`, `rdf:object` combo, without using any other property whose purpose is to replicate the structure of `rdf:reifies`. The presence of an additional property to create the structure differentiating propositions with reifiers is unavoidable.

niklasl:
@rat10 Where have I said that "an assertion on a reifier" is "an assertion on an asserted triple"? If by "assertion on" you mean the subject of an asserted triple, I do not at all claim that the asserted triple is or should be the subject of a triple annotation. On the contrary, the subject is a reifier. That, by definition, `rdf:reifies` the triple term, the latter which may very well also be asserted. That is all in the graph (not the interpretation thereof, which is usually what we "see" in the depictions, even if "ce n'est pas un pipe"). Its _meaning_ is that the proposition holds, and is also reified by something.

rat10:
@pfps 

> > I wasn't referring to figures. It is the impression that I take away from the text and also from e.g. a lot of @niklasl's comments. So you are saying that there is no such relationship, and that an assertion on a reifier is strictly only an assertion on that reifier, not on the triple (term) it reifies, and not on any such (asserted) triple in the same graph?
> 
> An assertion on a refier is in no way an assertion _on_ the triple term that it reifies. In fact, the whole idea of reifiers is to make this _not_ be the case.

Yes, of course, I was sloppy in not adding "... an instance/token/occurrence of..." 

> Instead a reifier is a way to collect multiple assertions about some thing related to the triple term 

But how exactly "related" is the interesting question, is the crucial detail. And how does that relation correspond to other relations that e.g. an asserted triple may have to the same triple term?


@niklasl 
> @rat10 Where have I said that "an assertion on a reifier" is "an assertion on an asserted triple"? 

You seem to say, e.g. in your Lotico talk, and also RDF 1.2 Concepts Section 1.5 seems to say that if a reifier refies a triple term and the same triple also occurs as asserted triple in the same graph, then the assertion on the reifier is about that asserted triple - simply because they refer to the same proposition.

> If by "assertion on" you mean the subject of an asserted triple,  I do not at all claim that the asserted triple is or should be the subject of a triple annotation. 

I'm not talking about a triple in subject position of another triple. Check my comment again. I'm careful to speak about reference: "if the triple is both referenced as a triple term via a reifier and asserted in the graph, then an annotation on the reifier annotates the asserted triple". 

> On the contrary, the subject is a reifier. That, by definition, `rdf:reifies` the triple term, the latter which may very well also be asserted. That is all in the graph (not the interpretation thereof, which is usually what we "see" in the depictions, even if "ce n'est pas un pipe"). Its _meaning_ is that the proposition holds, and is also reified by something.

@pfps, @niklasl
I'm not sure if we start to understand each other, but I am sure that we wouldn't have all these difficulties of establishing a common understanding if reifying the triple term would entail it, plain and simple (like the `rdfs:states`property proposed, and like RDF* did).

pfps:
> @pfps, @niklasl I'm not sure if we start to understand each other, but I am sure that we wouldn't have all these difficulties of establishing a common understanding if reifying the triple term would entail it, plain and simple (like the `rdfs:states`property proposed, and like RDF* did).

i believe that this change would make no difference whatsoever to the reality of the situation.  There would still be lots of issues related to "stating", "asserting", etc.  The only difference would be that there would be no possibility of having a reification in a graph of a triple that is not also asserted in the graph.

My view is that a lot of the problems the WG is encountering in this area is due to sloppy language.  We all use sloppy language, because that's what makes communication efficient.  But sometimes one has to be precise.  In RDF 1.2 there are triples, and that's it.  Phrases like "the asserted triple" or "the unasserted triple" or "asserting a triple" have to be used with care and avoided when one is trying to be precise, instead saying things like "the triple ..., which is asserted in this graph," or "the triple ..., which is not asserted in this graph,".

rat10:
> @rat10, your use of `rdf:value` corresponds to `rdf:reifies`.

`rdf:reifies` is indeed so vaguely specified that it seems to have no more semantics than `rdf:value`, aside from the range definition.

rat10:
> > ... and also RDF 1.2 Concepts Section 1.5 seems to say ...
> 
> Could say which text exactly? (Chrome's "Copy link to highlight" or copy the key text here.)

"This represents a graph where an [asserted triple](https://w3c.github.io/rdf-concepts/spec/#dfn-asserted-triple) corresponds to the [triple term](https://w3c.github.io/rdf-concepts/spec/#dfn-triple-term) object of a [reifying triple](https://w3c.github.io/rdf-concepts/spec/#dfn-reifying-triple). In this case, the subset of triples including the [reifier](https://w3c.github.io/rdf-concepts/spec/#dfn-reifier) as subject—as illustrated in these examples— is called a triple annotation." 
[link to cite in original context](https://w3c.github.io/rdf-concepts/spec/#:~:text=This%20represents%20a%20graph%20where%20an%20asserted%20triple%20corresponds%20to%20the%20triple%20term%20object%20of%20a%20reifying%20triple.%20In%20this%20case%2C%20the%20subset%20of%20triples%20including%20the%20reifier%20as%20subject%E2%80%94as%20illustrated%20in%20these%20examples%E2%80%94%20is%20called%20a%20triple%20annotation.)

afs:
[Discussion item 2025-09-25](https://www.w3.org/2025/09/25-rdf-star-minutes.html#2aab)

Issue #221, opened 2025-07-21
dfn changes breaking external links


afs:
```
https://www.w3.org/TR/rdf12-concepts/#dfn-literal
https://www.w3.org/TR/rdf12-concepts/#dfn-node
```
used, externally, are flagged as breaking documents, e.g. SHACL.

I think the changes that happened are:

`<dfn class="export" data-lt="RDF literal">literal</dfn>`  
was  
`<dfn data-local-lt="RDF literal">literal</dfn>`.

If the intention is that the preferred terminology is "rdf literal" then `<span>` alias would be in order (does repec itself support alt names?).

`<span id="dfn-nodes"><!-- obsolete term--></span><dfn class="export" data-lt="RDF node">nodes</dfn>`  
was  
`<span id="dfn-nodes"><!-- obsolete term--></span><dfn data-lt="node">nodes</dfn>`.

I'm less clear what the intention is here.

The `#dfn-node` (singular) is in RDF 1.1. But the concept at this point is "a set of nodes".

Is "dfn-nodes" obsolete? "dfn-nodes" does not seem to be in the unversioned TR. It is just in `rdf-xml` 1.2 and I didn't find it as a link in the unversioned TR of that either.

gkellogg:
IIRC, we've always exported `#dfn-rdf-literal` and `#dfn-rdf-node`, and those are still anchors in the published WD. I believe it was `dfn-node` which is obsolete.

I think @pchampin may have been working on this recently, but the errors don't seem to be in RDF Concepts, but in the references from other specs, which seem to have been replicated (at least for `dfn-rdf-literal`).

gkellogg:
In Turtle, there are definitely a lot of references to `dfn-literal`, and the spans would have been intended to maintain those anchors while publishing the "official" `dfn-rdf-literal` anchor.

While we should be sure that we export the terms we want from Concepts, we should also maintain as unexpected anchors those that are obsolete.

afs:
> but the errors don't seem to be in RDF Concepts, but in the references from other specs,
What's the local in `data-local-lt="RDF literal"`?

RDF 1.2 Concepts changed.

SHACL broke, as will SPARQL when the bot gets round to it.

RDF 1.1 has
  https://www.w3.org/TR/rdf11-concepts/#dfn-literal
  https://www.w3.org/TR/rdf11-concepts/#dfn-node

but not `#dfn-rdf-literal` or `#rdf-node`.

[...]

PR #215, opened 2025-07-07
export normative terms

pchampin:
Address #152

I broke my own rule of systematically exporting normative definitions...
I strongly felt that the following terms were defined for internal purposes only,
more precisely to help define other terms, so I chose to not export them.

- IRI equality
- Blank node equality
- Literal equality
- Triple equality
- isomorphic RDF-term mapping

I experimented with scoped definition [1]
for genetic terms such as 'atomic', 'node', 'subject'...
but I found that the ergonomic toll was not worth it.
In particular, there is no hint in the HTML
that the definition requires some additional context to be used.

[1] https://respec.org/docs/#scoped-concepts

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/215.html" title="Last updated on Jul 8, 2025, 10:41 AM UTC (bb6d4dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/215/9714dbd...bb6d4dd.html" title="Last updated on Jul 8, 2025, 10:41 AM UTC (bb6d4dd)">Diff</a>

PR #222, opened 2025-07-23
restore broken dfn anchors

pchampin:
these IDs were modified by #215,
causing problems in other specs (see #221).

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -608,11 +608,11 @@ Accept: text/turtle; version=1.2
   <p>This specification establishes two conformance levels:</p>
 
   <ul>
-    <li><dfn class="export" data-local-lt="full">Full conformance</dfn>
+    <li><dfn id="dfn-full" class="export" data-local-lt="full">Full conformance</dfn>
       supports <a data-lt="RDF graph">graphs</a> and <a data-lt="RDF dataset">datasets</a>
       with <a>triples</a> that contain <a>triple terms</a>.
       </li>
-    <li><dfn class="export" data-local-lt="basic">Basic conformance</dfn>
+    <li><dfn id="dfn-basic" class="export" data-local-lt="basic">Basic conformance</dfn>
       only supports <a data-lt="RDF graph">graphs</a> or <a data-lt="RDF dataset">datasets</a>
       with <a>triples</a> that do not contain <a>triple terms</a>.</li>
   </ul>
@@ -758,7 +758,7 @@ Accept: text/turtle; version=1.2
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
-    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
+    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn id="dfn-node" class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
       is the set of <a>subjects</a> and <a>objects</a> of the <a>asserted triples</a> of the graph.
       It is possible for a [=predicate=] [=IRI=] to also occur as a [=node=] in
       the same graph.</p>


spec/index.html
```diff
@@ -758,7 +758,7 @@ <h3>Triples</h3>
 
     <p>The three components (|s|, |p|, |o|) of an [=RDF triple=] are respectively called the <dfn class=export >subject</dfn>, <dfn class=export >predicate</dfn> and <dfn class=export >object</dfn> of the triple.</p>
 
-    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
+    <p>The set of <span id="dfn-nodes"><!-- obsolete term--></span><dfn id="dfn-node" class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
```
afs, line 761:
```suggestion
    <p>The set of <span id="dfn-rdf-node"></span> <span id="dfn-nodes"></span><dfn id="dfn-node" class="export" data-lt="RDF node">nodes</dfn> of an <a>RDF graph</a>
```
to match #215.

There may be a better way to write this in reSpec.

gkellogg: This looks like it sets two anchors, both “dfn-node”. I think the span should be dfn-node, but the dfn be dfn-rdf-node. When searching for a term, at least theoretically, it’s the dfn version (when exported). Anything else remains for historic references.

afs: The suggestion change is adding`dfn-rdf-node` that is, for a few days only, used in SHACL 1.2 Core.

3 different anchors : `dfn-nodes`, `dfn-rdf-node`, `dfn-node`.

RDF 1.1 : `#dfn-node`
RDF 1.2 current, post 215 : `#dfn-nodes` 
RDF 1.2 current, post 215 : `#dfn-rdf-node``

We could make a breaking change of new items introduced recently as long as `#dfn-nodes` works for RDF 1.1 TR/rdf-concepts.

pchampin: I was working under the assumption that the anchors that existed only for a few days after #215 was merged could safely be dropped, but the SHACL example proved me wrong, so I merged Andy's proposal to be on the safe side...

@afs do you think that we also need to re-introduced `#dfn-full-conformance` and `#dfn-basic-conformance`?

afs: SHACL core has updated to `dfn-rdf-literal`, `dfn-rdf-node` (singular). https://github.com/w3c/data-shapes/pull/437

> about dfn-node or dfn-literal

`dfn-node` (singular) and `dfn-literal` are in RDF 1.1 Concepts.

So we can keep exactly those as `<span>`, and remove the `dfn-nodes` before it leaks.

> @afs do you think that we also need to re-introduced `#dfn-full-conformance` and `#dfn-basic-conformance`?

I don't know of any reason to. These are new concepts.

I'm sure exactly how data-lt works so writingit out, is this what we want?

```<span id="dfn-rdf-node"></span><dfn id="dfn-rdf-node" class="export" data-lt="RDF node">nodes</dfn>```
  
```<span id="dfn-literal"></span><dfn id="dfn-rdf-literal" class="export" data-lt="RDF literal">literal</dfn>```

TallTed: I think it's important to keep the `<!-- obsolete term-->` in the otherwise empty `<span>`, to remind future editors (including ourselves) why that empty `<span>` is there, such that it not be removed without otherwise addressing the reason for its presence.

I *think* I've absorbed how these work well enough to provide the following guidance. If I'm in error anywhere, please forgive and correct me.

The empty `<span>` gets the `@id` previously based on the old term — e.g., `id="dfn-literal"` — and the full `<dfn>` that follows gets an `@id` based on the new term — e.g., `id="dfn-rdf-literal"`. This keeps `@href`s from existing documents that target the old `@id` from breaking.

The `@data-lt` is required when the visible text (e.g., `nodes`) doesn't exactly match the term we mean to define (e.g., `data-lt="RDF node"`) occurrences of which elsewhere can be marked to link to this definition with plain wrappers of `<a> </a>`.

The `@data-lt` value *can* include multiple strings (separated by `|`) so multiple "source" texts (which need not include the visible text, nor most regular plurals/singular variants, because ReSpec already handles these; irregular plurals/singular variants are sometimes necessary) can be linked to here.

gkellogg (approved):

hartig (approved):

PR #225, opened 2025-08-02
Move 'Initial Text Direction' section to being a subsection of "3.4 Literals"

### LINKED ISSUE #189: Section 3.8 Initial Text Direction should be a subsection of 3.4 Literals
everything is in the title

currently, we have this [Section 3.8 Initial Text Direction](http://localhost:8000/spec/#section-text-direction) oddly sitting between 3.7 Graph Comparison an 3.9 Replacing Blank Nodes with IRIs ...

afs:
This closes #189
See also #224

Move the "Initial text direction" section to become section 3.4.3 under "3.4 Literals".

The only changes other than position are:
* Change other subsections in 3.,4 to be  `<h4>` - which makes no difference to reSpec and appearance.
* Change the case of the section title to "Initial text direction" because other 3.4.x are written that way.

There is no other changes to the text or to HTML layout.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1002,7 +1002,7 @@ Accept: text/turtle; version=1.2
     </ul>
 
     <section>
-      <h2>Representation of literals</h2>
+      <h4>Representation of literals</h4>
 
       <p>Some concrete syntaxes MAY support
         <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
@@ -1043,7 +1043,7 @@ Accept: text/turtle; version=1.2
     </section>
 
     <section>
-      <h2>Literal value</h2>
+      <h4>Literal value</h4>
 
         <p>The <dfn class="export">literal value</dfn> associated with a <a>literal</a> is defined as follows.</p>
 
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1089,6 +1089,40 @@ <h2>Literal value</h2>
         same literal <a>RDF term</a> because their
         <a>lexical forms</a> differ.</p>
     </section>
+
+    <section id="section-text-direction" class="informative">
+      <h4>Initial text direction</h4>
```
gkellogg, line 1094: When used as a title, I believe common usage requires it be. “Initial Text Directiom”, even if it is “Initial text direction” in the body.

gkellogg (changes requested):

gkellogg:
It actually looks like we’re pretty inconsistent in using Title Case, which we may want to address separately.

afs:
The case-fix is is showing in [the git diff](https://github.com/w3c/rdf-concepts/pull/225.diff).

niklasl (approved):

hartig (approved):

PR #230, opened 2025-08-07
proposed changes to use 'abstract syntax' where it is appropriate.

lisp:
this replaces those instances of "data model" which should be "abstract syntax" and comments out the related issue passage.
all of the existing appearances of "abstract syntax" are appropriate.

***
<a href="https://pr-preview.s3.amazonaws.com/datagraph/rdf-concepts/pull/230.html" title="Last updated on Aug 10, 2025, 10:30 AM UTC (1cd0859)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/230/c976b42...datagraph:1cd0859.html" title="Last updated on Aug 10, 2025, 10:30 AM UTC (1cd0859)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="color-scheme" content="light dark">
-  <title>RDF 1.2 Concepts and Abstract Syntax</title>
+  <title>RDF 1.2 Concepts and Abstract Data Model</title>
   <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
   <script src="./common/local-biblio.js" class="remove"></script>
   <script src="./common/fixup.js" class="remove"></script>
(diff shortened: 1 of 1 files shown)


TallTed:
> this replaces those instances of "data model" which should be "abstract syntax" and comments out the related issue passage.
> 
> all of the existing appearances of "abstract syntax" are appropriate.

It is difficult to fully review this PR, as the unchanged instances of "data model" and "abstract syntax" are invisible (having not been changed). Maybe some HTML comment could be inserted alongside the unchanged strings, such that we can review them all, and then deleted before merge (and harmless if missed by such deletion, being HTML comments that don't render)?

afs:
> I don't think we can abandon "data model",

I agree : most of these changes are to RDF 1.1 text including the abstract:

> This document defines an abstract syntax (a data model) in the Abstract

so the terminology is already in use.

lisp:
> It is difficult to fully review this PR, as the unchanged instances of "data model" and "abstract syntax" are invisible (having not been changed). 

as the criteria for acceptance should be whether the phrase is correct independent of whether it was original or is the result of the revision,  it should be sufficient to read the text as it stands.

lisp:
in any case, as the changes conflict with audience expectations.
they were proposed as the simplest way to avoid the evident contradiction between "abstract syntax" and "data model", but  i will withdraw them.

PR #234, opened 2025-08-15
Separate label for Durations

afs:
Give durations their own area.

Durations are not "Recurring and partial dates".

(Existing HTML table layout retained.)


domel:
Integrate with RDF Schema 3.6 XML Schema datatypes

domel:
@afs Yes, that was a comment mostly to me. RDF Schema follows the old classification. When you change it here, I should do the same in the RDFS spec.

gkellogg (approved):

PR #238, opened 2025-09-07
Further explain abstract data model and abstract syntax

niklasl:
Following #232, this adds to the introductions of the abstract data model and syntaxes, respectively, to further explain their respective meanings.

In the subsection about the abstract data model:
* Notes that this is a mathematical structure.
* Adds a diagram showing the same term used in two triples as a single node.
* Adds text about this being the basis of syntax and interpretation, and noting the difference between graph structure and conceptual modelling.

In the subsection about documents and syntaxes:
* Adds a definition of the abstract syntax (not as formal grammar, but as two similarly structured tables).

(There are also a few related minor adjustments.)

While merely adding content, this also seeks more balanced proportions between the introductory subsections. Some of the points may be obvious, but the goal is to clarify these notions to a wider audience with varying backgrounds.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -149,14 +149,47 @@
         <object data="rdf-graph.svg"
           aria-label="An RDF graph with two nodes (Subject and Object) and an arc (Predicate) connecting them"></object>
       </a>
-      <figcaption>An RDF graph with two nodes (Subject and Object) and an arc (Predicate) connecting them</figcaption>
+      <figcaption>An RDF graph with two nodes (Subject and Object) and an arc (Predicate) connecting them.</figcaption>
     </figure>
 
     <p>There can be four kinds of <a>nodes</a> in an
       <a>RDF graph</a>: <a>IRIs</a>, <a>literals</a>,
       <a>blank nodes</a>, and <a>triple terms</a>.</p>
-    <!--
-    <p class="issue" data-number="129">There is a mixture of "Abstract Syntax" and "Data Model". We should have a consistent way to say "Abstract Syntax" vs "Data Model". One way is to use "Abstract Syntax" as the basis of semantics and usually say "Data Model" in Concepts otherwise.</p> -->
+
+    <p>This is a mathematical structure, and when one term appears in multiple
+      triples, these are simply multiple occurrences of that same term. For
+      example, in a graph containing two triples (here expressed in common
+      <a href="https://en.wikipedia.org/wiki/Set_(mathematics)">set</a> and
+      <a href="https://en.wikipedia.org/wiki/Tuple">tuple</a> notation, and
+      using abstract names as distinct terms):</p>
+    <pre class="box">
+    { (R1, P1, R3),
+      (R2, P2, R3) }
+    </pre>
+    <p>the term R3 is the same single term used twice, and there are five terms
+      in total. This is more readily shown in two-dimensional graph diagrams,
+      where one single node can simply be connected from or to multiple other
+      nodes using labelled arcs.</p>
+    <figure id="fig-rdf-graph-arcs">
+      <a href="rdf-graph-arcs.svg">
+        <object data="rdf-graph-arcs.svg"
+          aria-label="An RDF graph with thee nodes (R1, R2, and R3) and two arcs (P1 and P2) connecting R1 and R2 to R3"></object>
+      </a>
+      <figcaption>An RDF graph with thee nodes (R1, R2, and R3) and two arcs (P1 and P2) connecting R1 and R2 to R3.</figcaption>
+    </figure>
+
+    <p>This abstract data model can be encoded in different ways while
+      preserving the same stucture, as described in [[[#rdf-documents]]].</p>
+
+    <aside class="note">
+      The graph structure of the abstract data model is not a conceptual model.
+      It is a symbolic, structural basis for such modelling, enabling
+      definition and use of <a data-lt="vocabularies">RDF vocabularies</a>, and
+      semantic
+      <dfn data-cite="RDF12-SEMANTICS#dfn-interpretation">interpretation</dfn>
+      [[RDF12-SEMANTICS]].
+      Practical examples are given in the [[[RDF12-PRIMER]]] [[RDF12-PRIMER]].
+    </aside>
   </section>
 
   <section id="resources-and-statements">
@@ -255,7 +288,7 @@
       unique identifiers in a graph data model that describes resources.
       However, those interactions are critical to the concept of
       [[[LINKED-DATA]]], [[LINKED-DATA]],
-      which uses the RDF abstract syntax and serialization formats.</p>
+      which uses the RDF <a>abstract syntax</a> and serialization formats.</p>
   </section>
 
   <section id="vocabularies">
@@ -536,6 +569,74 @@
       and different ordering of triples. While these aspects can have great
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
+
+    <p>The basis for <a>concrete RDF syntaxes</a> is the structure of the
+      <a href="#data-model">abstract data model</a>,
+      called the <dfn>abstract syntax</dfn>;
+      summarized in the following two tables.</p>
+    <table id="tab-graph-abstract-syntax" class="simple">
+      <caption>The abstract syntax of RDF graphs</caption>
+      <tr>
+        <th>Production</th>
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td><a>RDF graph</a></td>
+        <td><em>a set of zero or more</em> <a>triples</a></td>
+      </tr>
+      <tr>
+        <td>triple</td>
+        <td><em>a 3-tuple of</em>
+          <a>subject</a>, <a>predicate</a>, <a>object</a></td>
+      </tr>
+      <tr>
+        <td>subject</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a></td>
+      </tr>
+      <tr>
+        <td>predicate</td>
+        <td><a>IRI</a></td>
+      </tr>
+      <tr>
+        <td>object</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a>
+          <em>or</em> <a>Literal</a>
+          <em>or</em> <a>triple term</a></td>
+      </tr>
+      <tr>
+        <td>triple term</td>
+        <td><a>triple</a></td>
+      </tr>
+    </table>
+    <br>
+    <table id="tab-dataset-abstract-syntax" class="simple">
+      <caption>The abstract syntax of RDF datasets</caption>
+      <tr>
+        <th>Production</th>
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td><a>RDF dataset</a></td>
+        <td><em>a pair of one</em> <a>default graph</a>,
+          <em>and a set of zero or more</em> <a>named graphs</a></td>
+      </tr>
+      <tr>
+        <td>default graph</td>
+        <td><a>RDF graph</a></td>
+      </tr>
+      <tr>
+        <td>named graph</td>
+        <td><em>a pair of</em>
+          <a>graph name</a>, <a>RDF graph</a></td>
+      </tr>
+      <tr>
+        <td>graph name</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a></td>
+      </tr>
+    </table>
   </section>
 
   <section id="section-version-announcement">
@@ -847,8 +948,8 @@ Accept: text/turtle; version=1.2
       is a <a>string</a> that conforms to the syntax
       defined in RFC 3987 [[!RFC3987]].</p>
 
-    <p>An IRI in the RDF abstract syntax
-      <!-- this passeage does not discuss abstract syntax. it concerns concrete systax -->
+    <p>For an IRI in a <a data-lt="concrete RDF syntax">concrete syntax</a> to conform to the
+      <a>abstract syntax</a>, it
       MUST be <a data-cite="RFC3986#section-5">resolved</a> per [[RFC3986]] and
       MUST NOT be a <a data-cite="RFC3986#section-4.2">relative reference</a>.
       An IRI MAY contain a <a data-cite="RFC3986#section-3.5">fragment identifier</a>.
@@ -1009,7 +1110,7 @@ Accept: text/turtle; version=1.2
       <p>Some concrete syntaxes MAY support
         <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
         <a>lexical form</a> without any <a>datatype IRI</a>, <a>language tag</a>, or <a>base direction</a>.
-        Simple literals are syntactic sugar for abstract syntax
+        Simple literals are syntactic sugar for <a>abstract syntax</a>
         <a>literals</a>
         with the <a>datatype IRI</a>
         <code>http://www.w3.org/2001/XMLSchema#string</code>
--- a/spec/rdf-graph-arcs.svg
+++ b/spec/rdf-graph-arcs.svg
@@ -1,0 +1,35 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 464 180">
+
+  <style>@import url('figure.css')</style>
+
+  <defs>
+    <marker id="arrowhead" orient="auto-start-reverse" markerWidth="10" markerHeight="10"
+            refX="0" refY="5">
+      <path d="M0,0 V10 L10,5 Z" fill="context-stroke" />
+    </marker>
+  </defs>
+
+  <g transform="translate(104 40)">
+    <g>
+      <ellipse rx="70" ry="26"/>
+      <text class="middle" y="5" font-size="16">R1</text>
+    </g>
+    <g transform="translate(70 0)">
+      <text class="middle" x="52" y="-5" font-size="16">P1</text>
+      <path d="M0,0 103,35" class="arrow"/>
+    </g>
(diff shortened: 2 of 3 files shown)


spec/index.html
```diff
@@ -149,14 +149,47 @@ <h3>Graph-based Abstract Data Model</h3>
...
+      <figcaption>An RDF graph with two nodes (Subject and Object) and an arc (Predicate) connecting them.</figcaption>
     </figure>
 
     <p>There can be four kinds of <a>nodes</a> in an
       <a>RDF graph</a>: <a>IRIs</a>, <a>literals</a>,
       <a>blank nodes</a>, and <a>triple terms</a>.</p>
-    <!--
-    <p class="issue" data-number="129">There is a mixture of "Abstract Syntax" and "Data Model". We should have a consistent way to say "Abstract Syntax" vs "Data Model". One way is to use "Abstract Syntax" as the basis of semantics and usually say "Data Model" in Concepts otherwise.</p> -->
+
+    <p>This is a mathematical structure, and when one term appears in multiple
```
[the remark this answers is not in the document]
niklasl: Fair point; and I will not argue for keeping it unless others do (or some variation thereof). Just noting that I formulated it so since [later on](https://w3c.github.io/rdf-concepts/spec/#change-over-time), the document contains:

> Since RDF graphs are defined as mathematical sets, 

which I wanted to highlight early on.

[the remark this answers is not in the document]
niklasl: Good point.

Just to be clear, given (more of that wikipedia text):
> [graphs](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)), which are [mathematical structures](https://en.wikipedia.org/wiki/Mathematical_structures) used to model pairwise relations between objects

that is not the stricter RDF graph notion in the abstract data model; right? That is, the graph notion implied in that text could even be taken _as_ a model/interpretation (beyond the Herbrand kind)? That would indeed defeat what I'm seeking here. So I readily accept your suggestion.

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      <tr>
+        <th>Production</th>
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td><a>RDF graph</a></td>
+        <td><em>a set of zero or more</em> <a>triples</a></td>
+      </tr>
+      <tr>
+        <td>triple</td>
```
hartig, line 588: Is there a reason why you mark up "RDF graph" as a link but the rest of the entries in this column of the table are not marked up as links? That looks a bit odd.

(Same comment for the second table as well.)

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+        <td><em>a 3-tuple of</em>
+          <a>subject</a>, <a>predicate</a>, <a>object</a></td>
+      </tr>
+      <tr>
+        <td>subject</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a></td>
```
hartig, line 595:
```suggestion
        <td><em>either an</em> <a>IRI</a>
          <em>or a</em> <a>Blank node</a></td>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      </tr>
+      <tr>
+        <td>predicate</td>
+        <td><a>IRI</a></td>
```
hartig, line 599:
```suggestion
        <td><em>an</em> <a>IRI</a></td>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      </tr>
+      <tr>
+        <td>object</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a>
+          <em>or</em> <a>Literal</a>
+          <em>or</em> <a>triple</a></td>
```
hartig, line 606:
```suggestion
        <td><em>either an</em> <a>IRI</a>
          <em>or a</em> <a>Blank node</a>
          <em>or a</em> <a>Literal</a>
          <em>or a</em> <a>triple</a></td>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td><a>RDF graph</a></td>
+        <td><em>a set of zero or more</em> <a>triples</a></td>
+      </tr>
+      <tr>
+        <td>triple</td>
+        <td><em>a 3-tuple of</em>
+          <a>subject</a>, <a>predicate</a>, <a>object</a></td>
```
hartig, line 590:
```suggestion
          <em>a</em> <a>subject</a>, <em>a</em> <a>predicate</a>, <em>and an</em> <a>object</a></td>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+    <br>
+    <table id="tab-dataset-abstract-syntax" class="simple">
+      <caption>The abstract syntax of RDF datasets</caption>
+      <tr>
+        <th>Production</th>
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td><a>RDF dataset</a></td>
+        <td><em>a pair of one</em> <a>default graph</a>,
```
hartig, line 618:
```suggestion
        <td><em>a pair of a</em> <a>default graph</a>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+      </tr>
+      <tr>
+        <td>named graph</td>
+        <td><em>a pair of</em>
+          <a>graph name</a>, <a>RDF graph</a></td>
+      </tr>
+      <tr>
+        <td>graph name</td>
+        <td><em>either</em> <a>IRI</a>
+          <em>or</em> <a>Blank node</a></td>
```
hartig, line 633:
```suggestion
        <td>default graph</td>
        <td><em>an</em> <a>RDF graph</a></td>
      </tr>
      <tr>
        <td>named graph</td>
        <td><em>a pair of</em>
          <em>a</em> <a>graph name</a> <em>and an</em> <a>RDF graph</a></td>
      </tr>
      <tr>
        <td>graph name</td>
        <td><em>either an</em> <a>IRI</a>
          <em>or a</em> <a>Blank node</a></td>
```

spec/index.html
niklasl: Good point! I wanted to link to what is being defined; but having all productions linked to their definitions was a bit too much, so I just left the leading ones (the ones whose structures are being defined). But those can also be linked in the table caption, which I think looks better. Done in 7d2d97fa3015f2d984a22a0f882c1c1e5dfa522c.

spec/index.html
```diff
@@ -847,8 +948,8 @@ <h3>IRIs</h3>
       is a <a>string</a> that conforms to the syntax
       defined in RFC 3987 [[!RFC3987]].</p>
 
-    <p>An IRI in the RDF abstract syntax
-      <!-- this passeage does not discuss abstract syntax. it concerns concrete systax -->
+    <p>For an IRI in a <a data-lt="concrete RDF syntax">concrete syntax</a> to conform to the
+      <a>abstract syntax</a>, it
```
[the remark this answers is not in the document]
niklasl: Minor variation&mdash;I'd prefer reverting to keeping "abstract syntax" (but linked to its definition) and just remove the comment:
```suggestion
    <p>An IRI in the RDF <a>abstract syntax</a>
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td>RDF graph</td>
+        <td><em>a set of zero or more</em> <a>triples</a></td>
+      </tr>
+      <tr>
+        <td>triple</td>
+        <td><em>a 3-tuple of</em>
+          <em>a</em> <a>subject</a>, <em>a</em> <a>predicate</a>, <em>and an</em> <a>object</a></td>
```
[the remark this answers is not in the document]
domel: I think so.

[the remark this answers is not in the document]
domel: @hartig Totally agree. A small side note: the `<i>` element is no longer considered the “italicized text” tag as it once was. In HTML5, it can still be used, but its semantics are different: `<i>` is meant for text that is set apart from the surrounding prose (for instance, technical terms, foreign words, or taxonomic designations) without adding special importance.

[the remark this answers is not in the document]
niklasl: My intent was to set it apart, to make this as close to a formal notation as possible while remaining in prose (rather than using `|` and `*` which would anyway have to be explained or be expressed in formal ABNF&mdash;which I don't think is that meaningful since it is not a concrete syntax).

I thought about using `<i>`, and, since its semantics you mentioned @domel does reflect my intent, I changed to this for the operators/types only in 49a5b47af52c3613b8921bcc32e521e788ab9eed. It's not used anywhere else in this document though, whereas e.g. `<em><strong>true</strong></em>` is used in as a value space value.

I think it gives an appropriate enough structure; but if you strongly prefer to remove it I'll do so.

hartig (approved):

TallTed:
https://github.com/w3c/rdf-concepts/issues/242 is related to this.

I got notification of the commit https://github.com/w3c/rdf-concepts/pull/238/commits/8b566d1aea5b534baee785aba7a7167365714e73 before this PR.

spec/index.html
```diff
@@ -149,14 +149,47 @@ <h3>Graph-based Abstract Data Model</h3>
...
+    <p>the term R3 is the same single term used twice, and there are five terms
+      in total. This is more readily shown in two-dimensional graph diagrams,
+      where one single node can simply be connected from or to multiple other
+      nodes using labelled arcs.</p>
+    <figure id="fig-rdf-graph-arcs">
+      <a href="rdf-graph-arcs.svg">
+        <object data="rdf-graph-arcs.svg"
+          aria-label="An RDF graph with thee nodes (R1, R2, and R3) and two arcs (P1 and P2) connecting R1 and R2 to R3"></object>
+      </a>
+      <figcaption>An RDF graph with thee nodes (R1, R2, and R3) and two arcs (P1 and P2) connecting R1 and R2 to R3.</figcaption>
```
TallTed, line 178:
```suggestion
          aria-label="An RDF graph with three nodes (R1, R2, and R3) and two arcs (P1 and P2), the arcs respectively connecting R1 and R2 to R3"></object>
      </a>
      <figcaption>An RDF graph with three nodes (R1, R2, and R3) and two arcs (P1 and P2), the arcs respectively connecting R1 and R2 to R3.</figcaption>
```

spec/index.html
```diff
@@ -255,7 +288,7 @@ <h3>The Referent of an IRI</h3>
       unique identifiers in a graph data model that describes resources.
       However, those interactions are critical to the concept of
       [[[LINKED-DATA]]], [[LINKED-DATA]],
-      which uses the RDF abstract syntax and serialization formats.</p>
+      which uses the RDF <a>abstract syntax</a> and serialization formats.</p>
```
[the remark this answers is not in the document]
TallTed: I do wonder that there is no `<a>` wrapper on `serialization formats`. This seems necessary to put `serialization formats` on equal footing with `the RDF <a>abstract syntax</a>`.

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
       and different ordering of triples. While these aspects can have great
       effect on the convenience of working with the <a>RDF document</a>,
       they are not significant for its meaning.</p>
+
+    <p>The basis for <a>concrete RDF syntaxes</a> is the structure of the
```
TallTed, line 573:
```suggestion
    <p>The basis for <a>concrete RDF syntaxes</a> (also called RDF serialization formats)
      is the structure of the
```

spec/index.html
```diff
@@ -536,6 +569,70 @@ <h3>RDF Documents and Syntaxes</h3>
...
+        <th>Defined as</th>
+      </tr>
+      <tr>
+        <td>RDF graph</td>
+        <td>a <i>set</i> of <i>zero or more</i> <a>triples</a></td>
+      </tr>
+      <tr>
+        <td>triple</td>
+        <td>a <i>3-tuple</i> of
+          a <a>subject</a>, a <a>predicate</a>, and an <a>object</a></td>
```
TallTed, line 590: It might be worthwhile to say somewhere that `an alternate labeling of this <i>3-tuple</i> is as an <a>entity</a>, an <a>attribute</a>, and a <a>value</a>`, and that `a <a>predicate</a> or <a>attribute</a> can also be referred to as a <a>property</a>`

spec/rdf-graph-arcs.svg
TallTed, line 1: Is it just me, or are these nodes just solid black ovals?

spec/index.html
```diff
@@ -1009,7 +1105,7 @@ <h4>Representation of Literals</h4>
       <p>Some concrete syntaxes MAY support
         <dfn data-lt="simple literal" class="export">simple literals</dfn> consisting of only a
         <a>lexical form</a> without any <a>datatype IRI</a>, <a>language tag</a>, or <a>base direction</a>.
-        Simple literals are syntactic sugar for abstract syntax
+        Simple literals are syntactic sugar for <a>abstract syntax</a>
```
TallTed, line 1108:
```suggestion
        Except where specified otherwise by a concrete syntax
        (also known as a serialization format) specification,
        as with the syntactic sugar of `a`, `true`, and `false` in Turtle,
        simple literals are syntactic sugar for <a>abstract syntax</a>
```

TallTed (changes requested):
Not yet approved. Depends on the handling of my comments above.

spec/rdf-graph-arcs.svg
niklasl: In the github diff they are not displayed correctly; since they share a CSS and the github diff preview doesn't properly use that. See the [githack preview](https://raw.githack.com/niklasl/rdf-concepts/graph-data-model-description/spec/index.html#data-model) for properly served (and thus rendered) versions. (Or serve this locally.)

spec/index.html
niklasl: Ouch; good catch!

TallTed:
I think #242 was overtaken by events. I'll leave it open for now, and probably close it based on resolution of all of the above.

spec/rdf-graph.svg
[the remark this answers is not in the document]
afs: Use the rawgit link.

spec/index.html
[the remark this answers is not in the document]
afs: "serialization formats" isn't a `<dfn>` (it is generally understand language).

spec/index.html
afs: Section 3.1 calls them subject-predicate-object, _not_ entity-attribute-value. 

Attribute-value is particular has other meanings e.g. XML.
"value" is used in many ways.

> <a>entity</a>

Where is entity `<dfn>`?

`<a>entity</a>` would be a link to `#dfn-entity` (handled by reSpec).

spec/index.html
afs: I don't know how to read this.

simple literals  - "like this" -  are strings as mentioned at 1107. Lexical forms have quoted round them.

Turtle `a` or `true` aren't a lexical forms. They are keywords.

spec/index.html
afs: `<a>subject</a>` `<a>predicate</a>` and `<a>object</a>` link to the formal definitions in section 3.

"entity-attribute-value" is not formal terminology. 

"Value" is unhelpful due to its multiple meanings (maybe you are getting cross-over from SHACL" where "value" is used for object but that not in RDF Concepts).

"attribute" mentioned in RDF Concept is that in XML / HTML.

spec/index.html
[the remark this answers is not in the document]
afs: `<a>word or phrase</a>` is processed by reSpec to give a `<h href="..."></a>` link to a `<dfn>`.

spec/index.html
afs: Repeats earlier text.

Line 291 explained that `<a>concrete RDF syntaxes</a>` (preferred presumable as it is a link to a `<dfn>`)

As I read it, "serialization format" is general usage, and `<a>concrete RDF syntaxes</a>` is our defined terminolgy.

spec/index.html
afs: line 1011/1107 says its about lexical forms - i.e. quoted strings.

 `a`, `true`, and `false` are keywords of the grammar - no quotes. The original text is fine.

I woudl chnage 1009/1105 and remove the `MAY` -- "Some concrete syntaxes support ..." which is a statement of fact.

domel (approved):

spec/index.html
afs: > Any explanation of the graph-based abstract data model in relation to E/R (conceptual) or EAV (storage?) notions

Maybe in the primer, but not in RDF Concepts. ("Why stop at E/R, EAV?")

Keep RDF Concepts about what RDF is with the terminology RDF uses.

In 1.2 it's quoted `"entity"` and AFAICT because semantics uses the word (and then only twice in explanatory text). The 1.2 clause could even be removed (not this PR!).

spec/index.html
niklasl, line 291: Like this?
```suggestion
      which uses the RDF <a>abstract syntax</a> and <a>concrete RDF syntaxes</a>,
	  the latter also referred to as serialization formats.</p>
```

spec/index.html
afs: If we accept https://github.com/w3c/rdf-concepts/pull/238#discussion_r2364611320 (applies to line 291), we don't need to repeat it by saying "(also called RDF serialization formats)".

PR #240, opened 2025-09-12
Review sec1.5 wv

william-vw:
In response to Enrico's message, I had a look at Section 1.5. I personally found it a bit abstract and difficult to read. This is an attempt to make the section more concrete and reduce its complexity.

I added justifications for the edits as in-line HTML comments.

***
<a href="https://pr-preview.s3.amazonaws.com/william-vw/rdf-concepts/pull/240.html" title="Last updated on Sep 19, 2025, 4:44 PM UTC (1a9d041)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/240/7691d6d...william-vw:1a9d041.html" title="Last updated on Sep 19, 2025, 4:44 PM UTC (1a9d041)">Diff</a>

PR #241, opened 2025-09-16
better examples with triple terms

pchampin:
This PR changes the examples about triple terms, replacing `:Alice :name "Alice"` with `':Alice :familyName "Liddell"`.

The rationale is that the current example might be confusing, looking like a tautology to the untrained eye. This is all the more true in the prose description of those example: '`:Bob` claims that `:Alice` is named "Alice"'...

I'm requesting reviews to check if this idea has traction. I'm keeping the PR as draft until all the other PRs on §1.5 are closed. Then we will need to:

* [ ] reflect the change in any prose description of the figures in the text

### DIFF
--- a/spec/asserted-triple-term.svg
+++ b/spec/asserted-triple-term.svg
@@ -16,11 +16,11 @@
     </g>
     <g transform="translate(180, 2)">
       <path d="M0,0 92,0" class="arrow"/>
-      <text x="26" y="-8">:name</text>
+      <text x="6" y="-8">:familyName</text>
     </g>
     <g transform="translate(363, 0)">
       <ellipse rx="80" ry="36"/>
-      <text class="middle" y="6">"Alice"</text>
+      <text class="middle" y="6">"Liddell"</text>
     </g>
   </g>
 
--- a/spec/triple-term.svg
+++ b/spec/triple-term.svg
@@ -20,11 +20,11 @@
     </g>
     <g transform="translate(180, 2)" class="abstract">
       <path d="M0,0 92,0" class="arrow"/>
-      <text x="26" y="-8">:name</text>
+      <text x="6" y="-8">:familyName</text>
     </g>
     <g transform="translate(363, 0)" class="unlinked">
       <ellipse rx="80" ry="36"/>
-      <text class="middle" y="6">"Alice"</text>
+      <text class="middle" y="6">"Liddell"</text>
     </g>
   </g>
 


william-vw:
Good idea. 

In fact, how about having two different examples:
- an example where a reifier denotes a statement or belief that the proposition holds (as is currently the case)
- an example where a reifier denotes an event or circumstance that makes the proposition true. For example, the Liz marriage example. Or, how Alice Liddell's last name changed to Hargreaves after marriage (according to Wikipedia).

These would illustrate both "example reifier denotations" described in the NOTE box -
> [...] reifiers may denote a variety of things that are related to these propositions (such as a statement or belief that the proposition holds, or an event or circumstance that makes the proposition true)

niklasl:
I agree with @pchampin. IMHO, Concepts (particularly the intro) should be theoretic (not necessarily abstract, but not tied to existing vocabularies (who are continuously evolving/coordinating; FOAF, DBPedia, Schema,. Wikidata, BibFrame, ...)). And I also strongly agree on the balance point (and time). Concrete examples from the wild seems more appropriate in the Primer (and/or an FAQ) where we still have some time for improvements.

niklasl:
@pchampin Just a nitpick to make more room for `:familyName`  in the SVG:s (spec/asserted-triple-term.svg and spec/triple-term.svg) for the terms:
```vim
%s/100/84/
%s/180/164/
%s/92/124/
%s/x="6"/x="20"/
%s/363/379/
``` 
(I couldn't suggest it in the interface; I can also just commit it to this branch if you prefer.)

pchampin:
@niklas thanks, and done in ccbb96a

niklasl (approved):

william-vw:
> Good idea.
> 
> In fact, how about having two different examples:
> 
>     * an example where a reifier denotes a statement or belief that the proposition holds (as is currently the case)
> 
>     * an example where a reifier denotes an event or circumstance that makes the proposition true. For example, the Liz marriage example. Or, how Alice Liddell's last name changed to Hargreaves after marriage (according to Wikipedia).
> 
> 
> These would illustrate both "example reifier denotations" described in the NOTE box -
> 
> > [...] reifiers may denote a variety of things that are related to these propositions (such as a statement or belief that the proposition holds, or an event or circumstance that makes the proposition true)

Perhaps the time has passed for bigger changes (and/or you just want it over with), but giving a second example would cover the gamut of reifier denotation examples mentioned in the document. It can also better fit the case where the triple corresponding to the triple term is asserted. IMO the "accordingTo" example becomes a bit less useful when the triple is also asserted.

For instance,
```
:Alice :familyName "Liddell" .
_:r0 rdf:reifies <<( :Alice :familyName "Liddell" )>> .
_:r0 :from "1852-05-04"^^xsd:date . # (date of birth)
_:r0 :until "1880-09-15"^^xsd:date .
```

niklasl:
I agree this would be good to merge.

Additional examples can also go into the Primer and/or the upcoming reification FAQ. (I think the "reified family names" example is a good one. Had something similar [about Anne Bonny at TPAC 2024](https://niklasl.github.io/rdf-docs/presentations/RDF-reifiers-1/#s19), showing how multiple facts may be "grouped" by a refiier (and reifiers of different kinds).)

hartig (approved):

niklasl:
Given the draft state and checkbox in the description, and #240 now being merged, I believe @pchampin prefers to rebase this on main and then adjust the text accordingly (just `:%s/'s name/'s family name/ | %s/"Alice"/"Liddell"/` AFAICS). I support that and then merge as approved.

niklasl (approved):

spec/index.html
```diff
@@ -350,11 +350,11 @@ <h3>Triple Terms and Reification</h3>
     </figure>
 
     <p>
-      In this example, the [=proposition=] denoted by the [=triple term=] (<code>:Alice</code>'s name being <code>"Alice"</code>) is not claimed to be true.
+      In this example, the [=proposition=] denoted by the [=triple term=] (<code>:Alice</code>'s family name being <code>"Liddell"</code>) is not claimed to be true.
```
hartig, line 353: When I read this, I first thought that the parenthesis is the triple term that this sentence talks about and so I wondered why there are no commas between subject, predicate, and object; why the predicate is not an IRI; etc. Eventually, I realized that this parenthesis is actually meant as a clarification for the proposition that the sentence talks about. To avoid others running into the same confusion, I suggest to change the parenthesis as follows.
```suggestion
      In this example, the [=proposition=] denoted by the [=triple term=] (i.e., the proposition that <code>:Alice</code>'s family name is <code>"Liddell"</code>) is not claimed to be true.
```

PR #245, opened 2025-09-20
section 1.5: alternative second example 

william-vw:
There was a lack of response to https://github.com/w3c/rdf-concepts/pull/241#issuecomment-3311895406 on the PR on improving examples in Section 1.5.

Hence, I repeat my comment here to make sure folks have noticed it.

Another second example could better fit the case where the triple corresponding to the triple term is asserted. IMO the "accordingTo" example becomes a bit less useful when the triple is also asserted.

This PR updates the text to accommodate the new example. If this has traction then I will attempt to make an SVG.

***
<a href="https://pr-preview.s3.amazonaws.com/william-vw/rdf-concepts/pull/245.html" title="Last updated on Sep 20, 2025, 1:19 PM UTC (f642d49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/245/22834d5...william-vw:f642d49.html" title="Last updated on Sep 20, 2025, 1:19 PM UTC (f642d49)">Diff</a>


niklasl:
You're right about the use of "claim" in the spec, and this is perhaps a dangerous use of language. I'd say, from the "outside", a graph is a claim (about the/a world), but the model "in itself" simply holds (provided that it is consistent). My main point is to avoid the implication that modelling claims within the interpretation can somehow affect what holds. We might want to make point explicitly?

I agree that the introduction does not cover all uses. I had hoped its mention of various kinds of reifiers already indicated that, but perhaps some improved wording can make it clearer. Linking to examples in the [Primer](https://www.w3.org/TR/rdf12-primer/#section-triple-terms) can also be valuable there.

william-vw:
> The documents come from the community. 

If there is no rule, then why not add examples for comprehensibility :-)

> Had a change of focus been taken up 2 years ago, then maybe. 

Sorry, I'm not following.

hartig:
> Material should go in [RDF Primer](https://github.com/w3c/rdf-primer) or [RDF New](https://github.com/w3c/rdf-new).

In fact, it may be an option to move (and merge) the whole content of the introduction section into the Primer.

pchampin:
Actually, what we could do is replace " This section describes this common usage" in §1.5 with "This section briefly describes this common usage. For more examples, refer to the RDF Primer."

afs:
> Actually, what we could do is replace " This section describes this common usage" in §1.5 with "This section briefly describes this common usage. For more examples, refer to the RDF Primer."

PR #246 for that suggestion.

And now this discussion can be on a new issue on RDF Primer (AFAICS PR can't be moved).

TallTed:
I'm only up to 4 days ago in this discussion, but I wanted to make sure I didn't forget — we should include an example of disagreement, like where { :Fred :asserts <<( :Peggy :owns :Ferrari )>> } and { :DMV :asserts <<( :Peggy :rents :Ferrari )>> } and...

In other words, *make use* of the ability to talk about statements without asserting them.

TallTed:
@william-vw — re: "help people use rdf-star properly"

`rdf-star` is long since deprecated. Long live RDF 1.2!

PR #255, opened 2025-10-23
RDF Reference IRI

### LINKED ISSUE #169: Relative IRI Reference should bind to the irelative-ref production
From https://github.com/w3c/rdf-turtle/issues/89#issuecomment-2737449139, the definition for [`relative IRI reference`](https://www.w3.org/TR/rdf12-concepts/#dfn-relative-iri) should clarify that this is for IRI references matching the `irelative-ref` production in the [IRI Grammar](https://www.w3.org/TR/rdf12-concepts/#iri-abnf).

### LINKED ISSUE #223: RDF Reference IRI
IRIs ([RFC3986](https://datatracker.ietf.org/doc/html/rfc3986), [RFC3987](https://datatracker.ietf.org/doc/html/rfc3987)) serve many purposes as well as their use in RDF. 

An **RDF Reference IRI** is an [IRI](https://www.w3.org/TR/rdf12-concepts/#iri-abnf) that also meets conditions that reduce variability, and hence increase interoperability, when referring to the same [resource](https://www.w3.org/TR/rdf12-concepts/#dfn-resource).

In the language of RFC 3986, an "RDF Reference IRI" is a "target URI" described in [1.2.3.  Hierarchical Identifiers](https://datatracker.ietf.org/doc/html/rfc3986#section-1.2.3).

The definition is based on the current (July 2025) material in RDF Concepts and goes in [section IRIs](https://www.w3.org/TR/rdf12-concepts/#section-IRIs). The advised conditions include material from [the note in that section](https://www.w3.org/TR/rdf12-concepts/#h-note-5).

Having a defined terminology in one place with an anchor means that all specs for concrete syntaxes can link to a common concept. It can be used in standards managed by other groups.

No data is invalidated by this proposal.

Semantic is unaffected.

This is describing an ideal with "SHOULD" language.  One reason for deciding not to follow a SHOULD is "existing data is already something else".

---

An **RDF Reference IRI**

MUST (no change here):

1. Be an RDF String
1. Conform to the grammar rule `IRI` in the [IRI grammar](https://www.w3.org/TR/rdf12-concepts/#iri-abnf)

SHOULD:

1. be invariant when applying the [resolution algorithm](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2). There should be no change of RDF Reference IRI (e.g. it does not depend on a base IRI) whether the resolution is "non-strict" or "strict" (see [RFC 3986, 5.2.2. Transform References](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2). Being idempotent means it does not matter how many times or where it is resolved, the IRI is unchanged.
1. Follow the recommendations of the URI scheme, where they are known.
1. Scheme names are lower case
1. %-encoding are correct (i.e. 2 hex characters) and uppercase.
1. No '.' or '..' segments [RFC 3986 section 3.3 Paths](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)
   (follows from being stable with respect to resolution).
1. Does not use punycode

Good practice:

1. The RDF String is Unicode Normalization Form C [I18N-Glossary].

We could also include in a note some of the requirements from important schemes (`did:` `http:`, `https:`, `urn`) e.g. http, omit `:80`.

Some material from the note remains as notes.

URLs: [HTML5 URLs](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#urls) are largely compatible with [RFC3987] IRIs, but are based on a processing model important for implementation within web browsers and are not described using a ABNF grammar.

Relative IRI references: Some concrete RDF syntaxes permit relative IRI references as a convenient shorthand that allows authoring of documents independently from their final publishing location. Relative IRI references must be resolved against a base IRI. Therefore, the RDF graph serialized in such syntaxes is well-defined only if a base IRI can be established [RFC3986].

afs:
This closes #223.
This closes #169 by referring to the `irelative-iri` production of the IRI grammar.

Add an informative section about minting IRIs, with defined terminology `RDF Reference IRI` for IRIs suitable for use in published RDF. The basic syntax of IRIs (RFC3986/7) is for other uses of IRIs and is too broad (e.g. `http:abcd`).

The section is "informative" because advice can not be checked or enforced when consuming published RDF. For example, the data consumer can not be assumed to know URI scheme rules.

Changes are in section 3.3.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/255.html" title="Last updated on Nov 3, 2025, 4:42 PM UTC (64634ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/255/c9c542d...64634ac.html" title="Last updated on Nov 3, 2025, 4:42 PM UTC (64634ac)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1038,7 +1038,7 @@ Accept: text/turtle; version=1.2
       </p>
       <p>
         An <dfn data-lt="RDF reference">RDF Reference IRI</dfn>,
-        sometimes just called <a>RDF Reference</a>,
+        sometimes called simply <a>RDF Reference</a>,
         is an <a>IRI</a> that is suitable for use as a global reference.
       </p>
       


spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
 
-      <p><strong>Relative IRI references:</strong>
+    <section id="reference-iris" class="informative">
+      <h4>RDF Reference IRIs</h4>
+      <p>
+        This section is advice to data publishers.
+      </p>
+      <p>
+        <a>IRIs</a> are used to denote <a>resources</a> and should identify the same
+        resource regardless of where the IRI reference is used.
```
hartig:
```suggestion
        resource regardless of where that IRI is used.
```

spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
+        [[RFC3987]] defines a syntax that can express
```
hartig:
```suggestion
        Yet, the syntax for IRIs, as defined by [[RFC3987]], can express
```

spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
+    </div>
 
-      <p><strong>Relative IRI references:</strong>
+    <section id="reference-iris" class="informative">
+      <h4>RDF Reference IRIs</h4>
+      <p>
+        This section is advice to data publishers.
+      </p>
+      <p>
+        <a>IRIs</a> are used to denote <a>resources</a> and should identify the same
```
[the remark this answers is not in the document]
TallTed:
```suggestion
        <a>IRIs</a> are used to denote <a>resources</a>, and each IRI should always identify the same
```

spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
+      
+      <p>
+        <b>Reference Resolution:</b>
+        An <a>RDF Reference IRI</a> is unchanged by
+        <a data-cite="RFC3986#section-5">reference resolution</a>. 
+        The <a data-cite="RFC3986#section-3.3">path component</a> 
+        starts with a `/` character and does not contain `.` or `..` segments.
+      </p>
+      <p>
+        <strong>Relative IRI references:</strong>
```
TallTed: It seems like `Reference Resolution:` should have the same semantic importance and styling as `Relative IRI references:`, so either both should be `<b>` or both should be `<strong>`. I think the latter is *probably* better, so that's my suggestion here.
```suggestion
        <strong>Reference Resolution:</strong>
        An <a>RDF Reference IRI</a> is unchanged by
        <a data-cite="RFC3986#section-5">reference resolution</a>. 
        The <a data-cite="RFC3986#section-3.3">path component</a> 
        starts with a `/` character and does not contain `.` or `..` segments.
      </p>
      <p>
        <strong>Relative IRI references:</strong>
```

spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
+      </p>
 
-      <p><strong>URI Schemes:</strong>
-        Implementations are encouraged to enforce the scheme-specific rules of
+      <p>
+        <strong>URI Schemes:</strong>
+        Implementations are encouraged to follow the scheme-specific rules of
         the common schemes, such as the
-        <a data-cite="RFC7230#section-2.7.1">scheme rules for HTTP/HTTPS</a> and the
+        <a data-cite="RFC7230#section-2.7">scheme rules for HTTP/HTTPS</a> and the
```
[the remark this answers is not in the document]
afs: The link text is supposed in include "scheme rules ...".Section 2.7 has a sections on http and a section on https.

spec/index.html
```diff
@@ -1016,35 +1016,70 @@ <h3>IRIs</h3>
...
+        <a data-cite="rfc7230#section-2.7.1">HTTP URI scheme</a> defines 
+        <a data-cite="rfc7230#http-URI">`http-URI`</a>, which
+        requires the presence of a non-empty host name, and, as a consequence, 
+        the path component will start with `/`. 
+        The [[RFC3987]] syntax permits IRIs such as `http:abcd` and `http:///abcd`,
+        but these are invalid because they do not satisfy the HTTP URI scheme definition.
+      </p>
+      <p>
+        An <dfn data-lt="RDF reference">RDF Reference IRI</dfn>,
+        sometimes just called <a>RDF Reference</a>,
```
TallTed:
```suggestion
        sometimes called simply <a>RDF Reference</a>,
```

Issue #248, opened 2025-10-05
Resolve "Editor's notes"


afs:
There are Editor's Notes (`class="ednote"`) in the following places:

- [x] https://www.w3.org/TR/rdf12-concepts/#sotd
  _Determine how to reference 1.2 test cases._
- [x] https://www.w3.org/TR/rdf12-concepts/#conformance
  _The conformance levels described above are tentative, and still the subject of group discussion. An alternative to conformance levels, "profiles", may be adopted instead, abandoned, or described in another specification._
- [ ] https://www.w3.org/TR/rdf12-concepts/#server-considerations
  Mention of "Unstar" document
- [ ] https://www.w3.org/TR/rdf12-concepts/#acknowledgments-for-rdf-1-2
    _Recognize members of the Task Force? Not an easy to find list of contributors._

pchampin:
> Guessing not, but `partially resolve #248` includes `resolve #248` which GH handles.

GH is too smart for its own good. Sorry about that!

ktk:
Proposed list of contributors:

> James Anderson, Dörthe Arndt, Jerven Bolleman, Erich Bremer, Pierre-Antoine Champin, Souripriya Das, Enrico Franconi, Adrian Gschwend, Olaf Hartig, Gregg Kellogg†, Ora Lassila, Niklas Lindström, Thomas Lörtsch, Peter Patel-Schneider, Dave Raggett, Felix Sasaki, Andy Seaborne, Ruben Taelman, Thomas Pellissier Tanon, Ted Thibodeau Jr, Dominik Tomaszuk, Gregory Williams, William Van Woensel, and Antoine Zimmermann
>
> † Gregg Kellogg passed away in September 2025. His vast contributions to RDF and the broader ecosystem of related standards are gratefully acknowledged.

PR #258, opened 2025-12-04
Reference test cases

afs:
We need to decide whether we need a "RDF 1.2 Test Cases document" similar to NOTE [RDF 1.1 Test Cases document](https://www.w3.org/TR/2014/NOTE-rdf11-testcases-20140225/) or whether some text in RDF Concepts is sufficient.

The other RDF 1.2 documents have a test suite link in their document details.

* Part of #248.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/258.html" title="Last updated on Dec 14, 2025, 5:45 PM UTC (93b917a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/258/39e660b...93b917a.html" title="Last updated on Dec 14, 2025, 5:45 PM UTC (93b917a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -15,6 +15,9 @@
       shortName:            "rdf12-concepts",
       copyrightStart:       "2004",
 
+      implementationReportURI: "https://w3c.github.io/rdf-tests/rdf/rdf12/reports/",
+      testSuiteURI:            "https://w3c.github.io/rdf-tests/rdf/rdf12/",
+
       previousPublishDate:  "2014-02-25",
       previousMaturity:     "REC",
       prevRecShortname:     "rdf11-concepts",


pchampin:
The current practice for RECs is *not* to publish a NOTE for the implementation reports -- most of the time they are hosted on github pages, so the proposal is inline with that.

I would suggest to also add a Respec header pointing to the implementation report:
https://respec.org/docs/#implementationReportURI

afs:
> I would suggest to also add a Respec header pointing to the implementation report:
> https://respec.org/docs/#implementationReportURI

What will be the location of the implementation report?

In RDF 1.1, each of the concrete test suites has it's own "/reports" area.

For now, I've put in 

```
    implementationReportURI: "https://w3c.github.io/rdf-tests/rdf/rdf12/reports/",
```

PR #265, opened 2026-01-15
improve discussion on text direction - fix #262

### LINKED ISSUE #262: Clarification on inline direction
https://www.w3.org/TR/rdf12-concepts/

While `rdf:dirLangString` handles the base direction, the specification should clarify how inline directional changes *within* a plain literal string should be handled, probably as a note, like this:

> While `rdf:dirLangString` specifies the base direction for the entire literal, inline directional changes within the string value (e.g., embedding text of opposite directionality) can be managed using Unicode bidirectional control characters. These control characters isolate parts of the text to ensure correct rendering under the Unicode Bidirectional Algorithm.
> 
> Markup-based approaches (e.g., using HTML `<bdi>` or `<span dir="...">` elements) are not applicable to plain literals like `rdf:dirLangString`, as they are intended for structured formats such as `rdf:HTML`. Instead, rely on Unicode controls for bidi support in plain text strings.
> 
> For more details, see:
> - [Unicode Bidirectional Algorithm basics](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics)
> - [Unicode controls vs. markup for bidi support](https://www.w3.org/International/questions/qa-bidi-controls)
> - [How to use Unicode controls for bidi text](https://www.w3.org/International/questions/qa-bidi-unicode-controls)

pchampin:
I changed the example, because the one that was provided was actually working even without an explicit base direction(the first strongly directional character was providing the right hint).

The new example is more tricky and works only when the base direction is explicitly provided.

To respond to #262 I added a second, even more complex example. Following the recommendations at [1], I presented the use of datatypes rdf:HTML or rdf:XMLLiteral as best practices, and suggested the use of Unicode control character as a fallback.

[1] https://www.w3.org/International/questions/qa-bidi-controls

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/265.html" title="Last updated on Mar 27, 2026, 9:16 AM UTC (7dd16f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/265/bc69779...7dd16f4.html" title="Last updated on Mar 27, 2026, 9:16 AM UTC (7dd16f4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1263,7 +1263,7 @@ Accept: text/turtle; version=1.2
         but this is not always sufficient to correctly render bidirectional text.
 
       <p>
-        Consider the arabic translation of the book title "HTML and CSS: Designing Websites".
+        Consider the Arabic translation of the book title "HTML and CSS: Designing Websites".
         In a left-to-right context (such as an English web page),
         with proper [=bidi isolation=] but <em>without</em> an explicit base direction,
         it would be incorrectly displayed as follows:
@@ -1281,16 +1281,16 @@ Accept: text/turtle; version=1.2
       <p>
         Note that the language and base direction address string external bidirectional issues,
         related to correctly displaying the string in context
-        (e.g. avoiding <a href="https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects">spillover</a>
-        problems). It does not address string internal directional issues,
-        which may occurr in more complex examples, such as the following:
+        (e.g., avoiding <a href="https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects">spillover</a>
+        problems). It does not address directional issues internal to the string,
+        which may occur in more complex examples, such as the following:
       </p>
-      <p class=example dir="ltr">&quot;<bdi dir="rtl" lang="ar">HTML و CSS: تصميم مواقع الويب</bdi>&quot; is the arabic title of the book.</p>
+      <p class=example dir="ltr">&quot;<bdi dir="rtl" lang="ar">HTML و CSS: تصميم مواقع الويب</bdi>&quot; is the Arabic title of the book.</p>
 
       <p>
         A [=directional language-tagged string=]
         representing that example will have the language tag `en` and the base direction `ltr`,
-        but also requires specific Unicode control characters to mark the text between the quotes as `rtl`.
+        but also requires specific Unicode bidirectional formatting characters to isolate and mark the text between the quotes as `rtl`.
       </p>
 
       <div class=note>


aphillips:
I was [actioned](https://github.com/w3c/i18n-actions/issues/204) by I18N with responding to this PR along with related issues in SHACL, etc. I'm going to repeat some of that information here (somewhat edited and thus slightly improved in this iteration):

---

`rdf:dirLangString` exists to address the best practices and requirements raised by I18N's document [StringMeta](https://www.w3.org/TR/string-meta/). Natural language content transmitted in various RDF constructs needs to have language and base string direction metadata in order to ensure proper selection, processing, and presentation. Language and base string direction metadata addresses string **external** bidirectional issues, including spillover problems. It is not intended to (and does not address) string **internal** directional issues, which may require bidi formatting characters (provided by Unicode) or markup.

While `rdf:HTML` (coupled with use of appropriate spanning markup using the `lang` and `dir` attributes) can be used as a solution for value-internal bidi and language changes, it does not replace value-external needs. However, lots of data in RDF contexts is not HTML but simply consists of normal string data. There is no reason to "decorate" this data with HTML markup. We usually presume that the source content, when it contains mixed-direction bidirectional text, has "already dealt" with any potential content-internal presentation issues by using the formatting characters provided by Unicode. The metadata is still necessary for such content (to set the base direction). 

---

This PR introduces the problem of initial text direction, going on to recommend `rdf:HTML` and `rdf:XMLLiteral`. I18N doesn't agree with this recommendation because it imposes a burden on users to decorate/use richer content types. These types are fine when available, but we point out that string data is ubiquitous. HTML or other markup still relies on proper internal handling of the language/direction upstream from the RDF application.

RDF Concepts should not be in the business of solving the content-internal presentation needs, other than to point to best practices and guidelines, such as the links to I18N's documentation (as you do here!). What you should focus on is transmitting or interpolating the value-external ("initial text direction") direction. This can be done by using the markup-based types but is equally well served by using `rdf:dirLangString`. 

I observe that discussion of string values elsewhere confuses the problems of content-internal and content-external needs.

As always, we remain happy to discuss. If I might suggest: a cross-over call might help a lot?

pchampin:
> I think the confusion is that the I18N article [cited](https://www.w3.org/International/questions/qa-bidi-controls) is about formatting **in HTML or XML content**. When you are already in a markup context, using markup is preferred. When in a plain text context, plain text formatting characters are preferred.

Oh my! Sorry for that confusion. It makes so much more sense now :)

I just modified this PR taking your comments into account. Thanks again.

afs (approved):

spec/index.html
```diff
@@ -1281,28 +1260,52 @@ <h4>Initial Text Direction</h4>
...
-      <p>For example, some text with a language tag "`he`" might be displayed
-        in a left-to-right context (such as an English web page) as
-        <bdi dir="ltr">פעילות הבינאום, W3C</bdi>, which is incorrect. When provided
-        to a user agent including base direction information (such as using
-        HTML's `dir` attribute) it can then be correctly presented as:
+      <p>
+        Note that the language and base direction address string external bidirectional issues,
+        related to correctly displaying the string in context
+        (e.g. avoiding <a href="https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects">spillover</a>
+        problems). It does not address string internal directional issues,
```
pchampin:
```suggestion
        problems). It does not address directional issues internal to the string,
```

PR #275, opened 2026-04-23
add note about sets of quads vs. dataset

### LINKED ISSUE #274: Caveat about "datasets as sets of quads"
RDF 1.2 Concepts introduces the notion of quads, stating

> An [RDF dataset](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-dataset) can be considered to be a set of [quads](https://www.w3.org/TR/rdf12-concepts/#dfn-quad)...

This is not true of absolutely all datasets. More precisely, this representation as a set of quads does not allow to represent a dataset with empty named graphs. I know that a number of implementation does not allow that either, precisely because they represent datasets as sets of quads – but since our definition of dataset allows for empty named graphs, I think we should acknowledge this limitation.

This change would be editorial because that section is non-normative anyway.

pchampin:
representing a dataset as a set of quads can be lossy if empty named graphs matter.

fix #274

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/275.html" title="Last updated on Apr 23, 2026, 11:53 AM UTC (63026c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/275/e0d243b...63026c9.html" title="Last updated on Apr 23, 2026, 11:53 AM UTC (63026c9)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1525,6 +1525,11 @@ Accept: text/turtle; version=1.2
       and quads with the same graph name
       supply the triples of the <a>named graph</a> with that name.</p>
 
+    <p>There are cases, however, where the representation of an <a>RDF dataset</a> as a set of <a>quads</a> is lossy:
+      an <a>RDF dataset</a> containing an empty <a>named graph</a> (|n|, ∅)
+      will be undistinguishable from an <a>RDF dataset</a> containing no graph named |n|.
+    </p>
+
     <p class="note">Although a <a>quad</a> without a <a>graph name</a>
       consists of the same three components as a <a>triple</a>,
       it is a distinct concept,
--- a/spec/index.html
+++ b/spec/index.html
@@ -1517,6 +1517,11 @@ Accept: text/turtle; version=1.2
       and quads with the same graph name
       supply the triples of the <a>named graph</a> with that name.</p>
 
+    <p>There are cases, however, where the representation of an <a>RDF dataset</a> as a set of <a>quads</a> is lossy:
+      an <a>RDF dataset</a> containing an empty <a>named graph</a> (|n|, ∅)
+      will be undistinguishable from an <a>RDF dataset</a> containing no graph named |n|.
+    </p>
+
     <p class="note">Although a <a>quad</a> without a <a>graph name</a>
       consists of the same three components as a <a>triple</a>,
       it is a distinct concept,
--- a/spec/index.html
+++ b/spec/index.html
@@ -1503,7 +1503,7 @@ Accept: text/turtle; version=1.2
   </section>
 
   <section id="section-dataset-quad" class="informative">
-    <h3>Dataset as a Set of Quads</h3>
+    <h3>Quads</h3>
 
     <p>A <dfn>quad</dfn> is a <a>triple</a> associated with an optional <a>graph name</a>
       and is used when referring to triples within an <a>RDF dataset</a>.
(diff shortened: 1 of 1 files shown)


hartig (approved):

spec/index.html
```diff
@@ -1525,6 +1525,11 @@ <h3>Dataset as a Set of Quads</h3>
       and quads with the same graph name
       supply the triples of the <a>named graph</a> with that name.</p>
 
+    <p>There are cases, however, where the representation of an <a>RDF dataset</a> as a set of <a>quads</a> is lossy:
+      an <a>RDF dataset</a> containing an empty <a>named graph</a> (|n|, ∅)
+      will be undistinguishable from an <a>RDF dataset</a> containing no graph named |n|.
+    </p>
```
afs:
```suggestion
    <p>However, the representation of an <a>RDF dataset</a> seen as a set of <a>quads</a> 
      does not allow empty <a>named graphs</a> to be shown.
    </p>
```

spec/index.html
```diff
@@ -1525,6 +1525,11 @@ <h3>Dataset as a Set of Quads</h3>
...
+
     <p class="note">Although a <a>quad</a> without a <a>graph name</a>
```
afs, line 1520: This is now moved away from the text it is most related to (about default graphs).

afs:
Datasets are now well-established whereas at RDF 1.1 they were new outside SPARQL.

I do wonder whether this section is needed anymore.  Adding caveats/exceptions (more text) only goes so far for what is supposed to be helpful information. What matters for the user is what their store or remote service actually does.

I think the new text is ingestion centric:

> the representation of an RDF dataset as a set of quads is lossy.

From a users POV, the more usual case is that a triple store is presenting its contents as a dataset. There is no loss - the presentation is what what the storage has, whether JSON-LD, TriG or N-Quads

If we're looking at this section:

> where quads with no [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) supply the [triples](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-triple) of the [default graph](https://www.w3.org/TR/rdf12-concepts/#dfn-default-graph),

which then gets explained by the note (!!)

A possibility is remove the section and put in N-Quads:

"Note: N-Quads can not represent an empty named graph."

which is significant even if the store has empty named graphs.

spec/index.html
pchampin: Sorry, I don't understand this comment

spec/index.html
pchampin: This is indeed simpler, but the term "shown" seems a bit handwavy.

pchampin:
> Datasets are now well-established whereas at RDF 1.1 they were new outside SPARQL.

agreed

> I do wonder whether this section is needed anymore.

Which section exactly do you mean? Note that Section "4.3 Dataset as a Set of Quads" did not exist in RDF 1.1. We added it in RDF 1.2 because, while the notion of "quad" is pervasive in implementations, it did not appear anywere in the spec (not even, I believe, in the N-Quads spec, besides the name of the format!).

> Adding caveats/exceptions (more text) only goes so far for what is supposed to be helpful information. What matters for the user is what their store or remote service actually does.
> 
> I think the new text is ingestion centric:
> 
> > the representation of an RDF dataset as a set of quads is lossy.

This sentence is not meant to be ingestion centric. Whether you are using a quad-based format or a quad-based implementation, you will find yourself unable to accurately represent some datasets.

> From a users POV, the more usual case is that a triple store is presenting its contents as a dataset. There is no loss - the presentation is what what the storage has, whether JSON-LD, TriG or N-Quads

If your store manages empty named graphs, and exposes its data as n-quads, then there is a loss.

Conversely, if you have the description of a dataset such as in [this test from the SPARQL test suite](https://w3c.github.io/rdf-tests/sparql/sparql11/aggregates/index.html#agg-empty-group-count-graph)
and your implementation does not manage empty named graphs, there is a loss.

(it so happens that I raise this issue after realizing that my implementation did not pass this test precisely because it did not support empty graph names).
> 
> If we're looking at this section:
> 
> > where quads with no [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) supply the [triples](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-triple) of the [default graph](https://www.w3.org/TR/rdf12-concepts/#dfn-default-graph),
> 
> which then gets explained by the note (!!)

The note has a different focus, it does not aim to explain that part...

> A possibility is remove the section and put in N-Quads:
> 
> "Note: N-Quads can not represent an empty named graph."
> 
> which is significant even if the store has empty named graphs.

I would be -0.9 on that, for the reason exposed at the top of my response.

hartig:
I agree with @pchampin

afs (approved):

ktk:
@pchampin please merge

PR #280, opened 2026-05-29
refactor considerations on versions

### LINKED ISSUE #189: Consolidate advice about versions of RDF syntax and HTTP access

pchampin:
this intends to fix https://github.com/w3c/rdf-star-wg/issues/189

* it simplifies §1.10 to make it more intro-like;
* it moves all the discussion about protocol (HTTP conneg) in an appendix;
* it mostly moves content around, but I did enrich some, stealing things from https://github.com/w3c/rdf-turtle/pull/135

NB: I did not really move things between RDF Concepts and Turtle; I found that overall, what was in each document fit there.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/pull/280.html" title="Last updated on Aug 6, 2026, 5:07 PM UTC (5b54432)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-concepts/280/74f17a6...5b54432.html" title="Last updated on Aug 6, 2026, 5:07 PM UTC (5b54432)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2131,7 +2131,8 @@ Content-Type: text/turtle; version=1.2
 <section id="sec-version-conneg" class="appendix informative">
   <h2>RDF versions and HTTP content-negotiation</h2>
 
-    <p>Concrete syntaxes for RDF 1.2, such as [[[RDF12-TURTLE]]] [[RDF12-TURTLE]],
+    <p>In addition to supporting inline version announcement,
+    concrete syntaxes for RDF 1.2, such as [[[RDF12-TURTLE]]] [[RDF12-TURTLE]],
     accept a `version` parameter in their media type.
     This enables servers (or clients sending data)
     to advertise RDF versions in the HTTP `Content-Type` header:
--- a/spec/index.html
+++ b/spec/index.html
@@ -706,8 +706,8 @@ Content-Type: text/turtle; version=1.2
 
     <p>When the version is indicated both in a media type parameter and in syntax,
       they are expected to be the same.
-      If they differ,
-      parsers use the version from the media type parameter and might emit a warning about the mismatch.</p>
+      Otherwise,
+      parsers might signal an error or a warning about the mismatch.</p>
   </section>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -789,8 +789,11 @@ Content-Type: text/turtle; version=1.2
       "be conservative in what you send, be liberal in what you accept".
       In this context, it means the following:
       <ul>
-        <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
-        <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
+        <li>implementations that produce data should announce the minimal version of RDF this data complies with,
+          in any way that is compatible with that version; and that</li>
+        <li>implementations that consume data should be ready to accept the data
+          even if no version is announced
+          or the announced version is not consistent with the RDF features used in this data.</li>
       </ul>
     </div>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -787,7 +787,7 @@ Content-Type: text/turtle; version=1.2
       <a href="https://en.wikipedia.org/wiki/Robustness_principle">robustness principle</a>
       (also known as Postel's Law):
       "be conservative in what you send, be liberal in what you accept".
-      In this context, it means that
+      In this context, it means the following:
       <ul>
         <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
         <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
@@ -2153,7 +2153,7 @@ Accept: text/turtle; version=1.2
       of the requested media type but potentially having a `version` other than what was
       requested.</p>
 
-    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like:</p>
+    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like this:</p>
 
     <pre class="box http">GET /document HTTP/1.1
 Host: example.com
--- a/spec/index.html
+++ b/spec/index.html
@@ -789,8 +789,8 @@ Content-Type: text/turtle; version=1.2
       "be conservative in what you send, be liberal in what you accept".
       In this context, it means that
       <ul>
-        <li>implementations producing data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
-        <li>implementation consuming data should be ready to accept it even when no version is announced, or the announced version is not consistent with the RDF features used in this data.</li>
+        <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
+        <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
       </ul>
     </div>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -694,7 +694,7 @@ For example, the term <code>:Alice</code> from a [=triple term=] and <code>:Alic
 
     <p>To announce the version in the HTTP responses using the <code>Content-Type</code> header,
       the server is expected to use the <code>version</code> parameter,
-      as illustrated in the following example response.</p>
+      as illustrated in the following example response:</p>
 
     <pre class="box http">HTTP/1.1 200 OK
 Content-Type: text/turtle; version=1.2
@@ -789,9 +789,9 @@ Content-Type: text/turtle; version=1.2
       "be conservative in what you send, be liberal in what you accept".
       In this context, it means the following:
       <ul>
-        <li>implementations that produce data should announce the minimal version of RDF this data complies with,
-          in any way that is compatible with that version; and that</li>
-        <li>implementations that consume data should be ready to accept the data
+        <li>Implementations that produce data should announce the minimal version of RDF this data complies with,
+          in any way that is compatible with that version.</li>
+        <li>Implementations that consume data should be ready to accept the data
           even if no version is announced
           or the announced version is not consistent with the RDF features used in this data.</li>
       </ul>
@@ -2131,7 +2131,7 @@ Content-Type: text/turtle; version=1.2
 <section id="sec-version-conneg" class="appendix informative">
   <h2>RDF versions and HTTP content-negotiation</h2>
 
-    <p>In addition to supporting inline version announcement,
+    <p>In addition to supporting inline version announcements,
     concrete syntaxes for RDF 1.2, such as [[[RDF12-TURTLE]]] [[RDF12-TURTLE]],
     accept a `version` parameter in their media type.
     This enables servers (or clients sending data)
@@ -2142,8 +2142,9 @@ Content-Type: text/turtle; version=1.2
 Content-Type: text/turtle; version=1.2
     </pre>
 
-    <p>This also enables clients to request a specific version
-      using <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]],
+    <p>This also enables clients to use
+      <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
+      to request a specific version,
       by including it in the <code>Accept</code> request header,
       as illustrated in the following example request:</p>
 
@@ -2157,14 +2158,18 @@ Accept: text/turtle; version=1.2
       of the requested media type but potentially having a `version` other than what was
       requested.</p>
 
-    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like this:</p>
+    <p>Note that a client does not have to use the `version` parameter in
+      the `Accept` header. A request can simply look like this:</p>
+
 
     <pre class="box http">GET /document HTTP/1.1
 Host: example.com
 Accept: text/turtle
     </pre>
 
-    <p>This carries no implication about the parsing capabilities of the client. However, a conservative approach would be to assume that this client only supports Turtle 1.1.
+    <p>This carries no implication about the parsing capabilities of the client.
+      However, a conservative approach would be to assume that this client only
+      supports Turtle 1.1.
     </p>
 
     <section id="server-considerations" class="informative">
@@ -2192,8 +2197,10 @@ Accept: text/turtle
 
       <p>The strategies above are not an exhaustive list.
         Neither are they exclusive.
-        For example, an implementation may chose to rewrite <a>triple terms</a> and truncate <a>base directions</a>,
-        or to rewrite <a>directional language-tagged strings</a> but return `406 "Not Acceptable"` in the presence of <a>triple terms</a>.
+        For example, an implementation may choose to rewrite <a>triple terms</a>
+        and truncate <a>base directions</a>,
+        or to rewrite <a>directional language-tagged strings</a>
+        but return `406 "Not Acceptable"` in the presence of <a>triple terms</a>.
       </p>
 
     </section>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -793,60 +776,28 @@ <h3>Version Labels</h3>
...
-          in the <a data-cite="JSON-LD11#the-i18n-namespace">`i18n` namespace</a>, as defined in [[JSON-LD11]]
-          (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
-          </ul></li>
-        <li>Truncate the data to remove the incompatible features:<ul>
-          <li>Drop any <a>triple</a> that contains a <a>triple term</a> (downgrading from version "1.2" to "1.2-basic").</li>
-          <li>Turn <a>directional language-tagged strings</a> into <a>language-tagged string</a> by dropping their base direction (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
-          </ul></li>
-        <li>Return `406 "Not Acceptable"`.</li>
-        <li>Ignore the requested version and return a native representation.</li>
+        <li>implementations producing data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
```
hartig:
```suggestion
        <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
```

spec/index.html
```diff
@@ -793,60 +776,28 @@ <h3>Version Labels</h3>
...
+        <li>implementation consuming data should be ready to accept it even when no version is announced, or the announced version is not consistent with the RDF features used in this data.</li>
```
hartig:
```suggestion
        <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
```

hartig (approved):
Just two minor edit suggestions.

spec/index.html
```diff
@@ -2174,6 +2125,87 @@ <h2>Replacing Blank Nodes with IRIs</h2>
     IRI that results in a well-known <a>URI</a> after IRI-to-URI mapping [[!RFC3987]].</p>
 </section>
 
+<section id="sec-version-conneg" class="appendix informative">
+  <h2>RDF versions and HTTP content-negotiation</h2>
+
```
afs: It would be worth mentioning in this section that the concrete syntaxes also have version information in-line.

spec/index.html
```diff
@@ -2174,6 +2125,87 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+        <li>Rewrite the data in an information-preserving way, for example:<ul>
+          <li>Eliminate <a>triple terms</a> by using
+          <i>basic encoding</i>, as defined in [[[RDF-INTEROP]]] (downgrading from version "1.2" to "1.2-basic").
+          </li>
+          <li>Replace <a>directional language-tagged strings</a>
+          with typed <a>literals</a> having a <a>datatype IRI</a>
+          in the <a data-cite="JSON-LD11#the-i18n-namespace">`i18n` namespace</a>, as defined in [[JSON-LD11]]
+          (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
+          </ul></li>
+        <li>Truncate the data to remove the incompatible features:<ul>
```
afs: Minor: maybe better to list (i.e. group) by feature triple term, text direction, not technique, "rewrite", "truncate". Otherwise it feels like a server chooses one technique .

spec/index.html
```diff
@@ -702,29 +702,12 @@ <h3>RDF Version Announcement</h3>
...
-    <p class="note">As HTTP <a data-cite="webarch#frag-coneg">content negotiation</a> is advisory,
-      clients receiving a document should be prepared to properly handle a document
-      of the requested media type but potentially having a `version` other than what was
-      requested.
-      Clients may consider down-grading the content to an appropriate version themselves
-      as discussed in <a href="#server-considerations" class="sectionRef"></a>.</p>
+    <p>When the version is indicated both in a media type parameter and in syntax,
+      they are expected to be the same.
+      If they differ,
+      parsers use the version from the media type parameter and might emit a warning about the mismatch.</p>
```
Tpt: Please ignore if I am too late in the discussion. I am a bit surprised by this line: I would tend to think that "closer" should override "wider" and so, the version set in the syntax is likely to be more accurate than the one set in the media type (a lot of media type in the wild are automatically guessed, badly configured...). This sounds less error prone to me.

pchampin: That's also something that I always found surprising, but my understanding is that this is the generally agreed upon heuristics, at least that's how the browsers work for charset encoding of HTML files.

For the rationale, I'm not sure whether I made that up or read it somewhere, but I think it was that end-users (putting HTML files in their public_html) are more likely to make mistakes that professional sysadmins...

afs: Tricky, it has to work in both directions , `GET` and `PUT`:

A version In `Content-Type` is not permanent; `PUT` only leaves the data file `VERSION` after the request.

`POST` is just messy!

Can we say that it's an error if there is a mismatch? Or for that the parameter should be a version greater than the contents? "should" - non-normative section.

niklasl:
Would it make sense to move the detailed `Content-Type` example from "1.10 RDF Version Announcement" to the new Appendix " C. RDF versions and HTTP content-negotiation" (and add a reference from 1.10 to C just as from "2.1 Version Labels")?

(I also think @pchampin's idea in https://github.com/w3c/rdf-star-wg/issues/189#issuecomment-4577329532 is interesting: to put Appendix C  in https://github.com/w3c/rdf-interop/ instead. But I think it's best to land this PR first and let that idea marinate for a while.)

spec/index.html
```diff
@@ -793,60 +776,28 @@ <h3>Version Labels</h3>
...
-          (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
-          </ul></li>
-        <li>Truncate the data to remove the incompatible features:<ul>
-          <li>Drop any <a>triple</a> that contains a <a>triple term</a> (downgrading from version "1.2" to "1.2-basic").</li>
-          <li>Turn <a>directional language-tagged strings</a> into <a>language-tagged string</a> by dropping their base direction (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
-          </ul></li>
-        <li>Return `406 "Not Acceptable"`.</li>
-        <li>Ignore the requested version and return a native representation.</li>
+        <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
+        <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
```
TallTed:
```suggestion
        <li>Implementations that produce data should announce the minimal version of
          RDF this data complies with, in any way that is compatible with that
          version.</li>
        <li>Implementations that consume data should be ready to accept the data even
          if no version is announced or the announced version is not consistent with
          the RDF features used in this data.</li>
```

spec/index.html
```diff
@@ -793,60 +776,28 @@ <h3>Version Labels</h3>
...
-      <h4>Server Considerations</h4>
-      <p>When serializing a graph or dataset which uses features incompatible
-        with a requested version, servers can consider different alternatives:</p>
-
+    <div class="note">
+      Implementations are encouraged to apply the
+      <a href="https://en.wikipedia.org/wiki/Robustness_principle">robustness principle</a>
+      (also known as Postel's Law):
+      "be conservative in what you send, be liberal in what you accept".
+      In this context, it means that
```
TallTed:
```suggestion
      In this context, it means the following:
```

spec/index.html
```diff
@@ -2174,6 +2125,87 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+Host: example.com
+Accept: text/turtle; version=1.2
+    </pre>
+
+    <p>As HTTP content negotiation is advisory,
+      clients receiving a document should be prepared to properly handle a document
+      of the requested media type but potentially having a `version` other than what was
+      requested.</p>
+
+    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like:</p>
```
TallTed:
```suggestion
    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like this:</p>
```

spec/index.html
pchampin: @TallTed I agree that those lines are longer than needed, but I prefer to use semantic linebreaks (https://sembr.org/). I'll address this.

spec/index.html
pchampin: I removed the explicit priority to media-type, and added the possibility to signal an error or a warning (the latter implying that the parser continues, making it best guess)

spec/index.html
pchampin: done

spec/index.html
csarven: I think one important perspective is to consider what's intended to be authoritative here. Generally speaking, the sender of an HTTP message is the one that's giving what's intended. Sometimes this is differentiated from the author-level intention.

Perhaps terminology along those lines can be expressed. See also https://www.w3.org/2001/tag/doc/mime-respect#embedded which considers `Content-Type` to be more authoritative than embedded metadata.

spec/index.html
pchampin: It is not that easy, because some options (e.g. "return 406 Not Acceptable") are not "per feature".

I will add a sentence after the list emphasizing that these options are neither exhaustive nor exclusive.

afs (approved):

spec/index.html
```diff
@@ -787,10 +787,13 @@ <h3>Version Labels</h3>
...
-      In this context, it means that
+      In this context, it means the following:
       <ul>
-        <li>implementations that produce data should announce the minimal version of RDF this data complies with, in any way that is compatible with that version; and that</li>
-        <li>implementations that consume data should be ready to accept the data even if no version is announced or the announced version is not consistent with the RDF features used in this data.</li>
+        <li>implementations that produce data should announce the minimal version of RDF this data complies with,
+          in any way that is compatible with that version; and that</li>
+        <li>implementations that consume data should be ready to accept the data
+          even if no version is announced
+          or the announced version is not consistent with the RDF features used in this data.</li>
```
TallTed:
```suggestion
        <li>Implementations that produce data should announce the minimal version of RDF this data complies with,
          in any way that is compatible with that version.</li>
        <li>Implementations that consume data should be ready to accept the data
          even if no version is announced
          or the announced version is not consistent with the RDF features used in this data.</li>
```

spec/index.html
niklasl:
```suggestion
      as illustrated in the following example response:</p>
```

niklasl (approved):
A good improvement! (But see editorial suggestions before merging.)

Perhaps 1.10 should explicitly reference C, but since it references 2.1 which references C, it's probably fine.

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
     IRI that results in a well-known <a>URI</a> after IRI-to-URI mapping [[!RFC3987]].</p>
 </section>
 
+<section id="sec-version-conneg" class="appendix informative">
+  <h2>RDF versions and HTTP content-negotiation</h2>
+
+    <p>In addition to supporting inline version announcement,
```
TallTed:
```suggestion
    <p>In addition to supporting inline version announcements,
```

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+    to advertise RDF versions in the HTTP `Content-Type` header:
+    </p>
+
+    <pre class="box http">HTTP/1.1 200 OK
+Content-Type: text/turtle; version=1.2
+    </pre>
+
+    <p>This also enables clients to request a specific version
+      using <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]],
+      by including it in the <code>Accept</code> request header,
```
TallTed:
```suggestion
    <p>This also enables clients to use
      <a data-cite="webarch#frag-coneg">content negotiation</a> [[WEBARCH]]
      to request a specific version,
      by including it in the <code>Accept</code> request header,
```

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+Host: example.com
+Accept: text/turtle; version=1.2
+    </pre>
+
+    <p>As HTTP content negotiation is advisory,
+      clients receiving a document should be prepared to properly handle a document
+      of the requested media type but potentially having a `version` other than what was
+      requested.</p>
+
+    <p>Note that a client does not have to use the `version` parameter in the `Accept` header. A request can simply look like this:</p>
```
TallTed:
```suggestion
    <p>Note that a client does not have to use the `version` parameter in
      the `Accept` header. A request can simply look like this:</p>

```

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+
+    <pre class="box http">GET /document HTTP/1.1
+Host: example.com
+Accept: text/turtle
+    </pre>
+
+    <p>This carries no implication about the parsing capabilities of the client. However, a conservative approach would be to assume that this client only supports Turtle 1.1.
```
TallTed:
```suggestion
    <p>This carries no implication about the parsing capabilities of the client.
      However, a conservative approach would be to assume that this client only
      supports Turtle 1.1.
```

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+          <li>Turn <a>directional language-tagged strings</a> into <a>language-tagged string</a> by dropping their <a>base direction</a> (downgrading from version "1.2" or "1.2-basic" to "1.1").</li>
+          </ul></li>
+        <li>Return `406 "Not Acceptable"`.</li>
+        <li>Ignore the requested version and return a native representation.</li>
+      </ul>
+
+      <p>The strategies above are not an exhaustive list.
+        Neither are they exclusive.
+        For example, an implementation may chose to rewrite <a>triple terms</a> and truncate <a>base directions</a>,
+        or to rewrite <a>directional language-tagged strings</a> but return `406 "Not Acceptable"` in the presence of <a>triple terms</a>.
```
TallTed:
```suggestion
        For example, an implementation may choose to rewrite <a>triple terms</a>
        and truncate <a>base directions</a>,
        or to rewrite <a>directional language-tagged strings</a>
        but return `406 "Not Acceptable"` in the presence of <a>triple terms</a>.
```

spec/index.html
```diff
@@ -2174,6 +2128,94 @@ <h2>Replacing Blank Nodes with IRIs</h2>
...
+      <h2>Client Considerations</h2>
+      <p>If a document has no stated `version` (either by HTTP header or internal),
+        clients should assume the maximum version that they support.</p>
+      <p>When parsing a document with conflicting versions,
+        or when parsing a document using unsupported versions, a parser
+        may do any of the following:</p>
+      <ul>
+        <li>Ignore the version directive.</li>
+        <li>Raise an error and abort.</li>
+        <li>Issue a warning and ignore the triples using unrecognized features.</li>
```
TallTed:
```suggestion
        <li>Issue a warning and ignore the triples that use unrecognized features.</li>
```

niklasl:
@pchampin The CRLF-bug appears to have struck again in [f3a91da](https://github.com/w3c/rdf-concepts/pull/280/commits/f3a91daee1420d4d8a0b623269eef8d2189551d2) (I guess the simplest fix is to amend that commit locally by `s/\r\n/\n/` and `push --force-with-lease`.)

